PR #6, opened 2017-03-16
[MRESOLVER-14] Statistics should be calculated using System.nanoTime …

ChristianSchulte:
…instead of System.currentTimeMillis.


jira-importer:
Resolve #1448

jira-importer:
Resolve #773

jira-importer:
Resolve #1448

PR #14, opened 2018-01-03
MRESOLVER-39: Sleep for short duration in spin lock.

stevestorey:
Revert the change to the sleep duration made in
ad50215d27feede0ad0e5eb83ae96c6d6fdcc639 as this causes any thread which
needs to sleep to do so for 900 seconds (15 minutes) by default
(DEFAULT_REQUEST_TIMEOUT / 2) which seems way to long.

### DIFF
--- a/maven-resolver-connector-basic/src/main/java/org/eclipse/aether/connector/basic/PartialFile.java
+++ b/maven-resolver-connector-basic/src/main/java/org/eclipse/aether/connector/basic/PartialFile.java
@@ -110,7 +110,7 @@ final class PartialFile
 
                     try
                     {
-                        Thread.sleep( Math.max( requestTimeout / 2, 100 ) );
+                        Thread.sleep( 100 );
                     }
                     catch ( InterruptedException e )
                     {


stevestorey:
I believe this is the fix for MRESOLVER-39. For those arriving here wondering how to test the patch:
1. Clone the repo, cd into it
2. `mvn clean package -am -pl maven-resolver-connector-basic -Drat.skip=true` (I had to add `-Drat.skip=true` to make it work locally both with and without my commit here - not sure if it's my setup or something else)
3. `cp maven-resolver-connector-basic/target/maven-resolver-connector-basic-1.1.1-SNAPSHOT.jar .../apache-maven-3.5.2/lib/ext/`

maven-resolver-connector-basic/src/main/java/org/eclipse/aether/connector/basic/PartialFile.java
```diff
@@ -110,7 +110,15 @@ else if ( requestTimeout > 0 && currentTime - lastTime > Math.max( requestTimeou
 
                     try
                     {
-                        Thread.sleep( Math.max( requestTimeout / 2, 100 ) );
+                        // This is intended to be a spin lock, so we should sleep for short durations
+                        // each time (and not large chunks of the requestTimeout, otherwise we will be waiting
+                        // minutes between checks for the lock, slowing down the build massively)
+                        long sleepDuration = 100l;
```
rfscholte: I don't understand why this fixes the issue. It used to be a sleep of 100ms or less, now it is simply 100ms (+logging...). So the sleep was often shorter, right? 
However, based on [ad50215](https://github.com/apache/maven-resolver/commit/ad50215d27feede0ad0e5eb83ae96c6d6fdcc639#diff-200a9e2ad4fc09fd4d917e113d7babc5) that used to be the original code.
I think this is masking another problem I don't understand yet. If you remove the logging I'm willing to merge this.

stevestorey: No, it was 100ms or *more* (Math.max), and `requestTimeout / 2` defaulted to 900000 - 15 minutes. I'm now wondering if a better fix is in fact to just change to Math.min instead, as perhaps that's what the original intent was?

rfscholte: Ah, you're right. I think the commit contains too much changes, so going back to simply 100ms is good enough for me.

stevestorey:
I've changed the commit to just simply revert the sleep back to 100ms, and removed the logging as requested.

slachiewicz:
looks like this pull can be closed as fix was merged in 599cc2b

jira-importer:
Resolve #1386

PR #23, opened 2018-11-13
[MRESOLVER-63] FileTransformer#transformData(File) should also throw TransformException

mcimbora:
Up-for-grabs #DevoxxBE

### DIFF
--- a/maven-resolver-api/src/main/java/org/eclipse/aether/transform/FileTransformer.java
+++ b/maven-resolver-api/src/main/java/org/eclipse/aether/transform/FileTransformer.java
@@ -46,6 +46,9 @@ public interface FileTransformer
      * 
      * @param file the file with the original data
      * @return the transformed data
+     * @throws IOException If an I/O error occurred
+     * @throws TransformException If the file could not be transformed
      */
-    InputStream transformData( File file ) throws IOException;
+    InputStream transformData( File file )
+        throws IOException, TransformException;
 }
--- a/maven-resolver-api/src/main/java/org/eclipse/aether/transform/TransformException.java
+++ b/maven-resolver-api/src/main/java/org/eclipse/aether/transform/TransformException.java
@@ -1,0 +1,68 @@
+package org.eclipse.aether.transform;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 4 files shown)


maven-resolver-api/src/main/java/org/eclipse/aether/transform/FileTransformationException.java
```diff
@@ -0,0 +1,79 @@
...
+
+import java.io.File;
+
+import org.eclipse.aether.RepositoryException;
+
+/**
+ * Thrown when file transformation failed.
+ */
+public class FileTransformationException
+        extends RepositoryException
```
rfscholte: This is not really a repository exception. Consider the following usecase: we want to transform the pom.xml during upload with xslt. When trying to generate the XMLParser you must catch the SaxException. This one should be wrapped and thrown by this method.

maven-resolver-api/src/main/java/org/eclipse/aether/transform/FileTransformationException.java
```diff
@@ -0,0 +1,79 @@
...
+{
+
+    private final transient File file;
+
+    /**
+     * Creates a new exception with the specified file.
+     *
+     * @param file The file which transformation failed, may be {@code null}.
+     */
+    public FileTransformationException( File file )
```
rfscholte: There's not always a file involved yet, no need to use this.

maven-resolver-api/src/main/java/org/eclipse/aether/transform/FileTransformer.java
```diff
@@ -47,5 +46,5 @@
      * @param file the file with the original data
      * @return the transformed data
      */
-    InputStream transformData( File file ) throws IOException;
+    InputStream transformData( File file ) throws FileTransformationException;
```
rfscholte: Both IOException and TransformException must be thrown. Better not make it too explicit, we might do other transformations as well in the future, not only filebased.

maven-resolver-api/src/main/java/org/eclipse/aether/transform/FileTransformationException.java
mcimbora: Removed.

maven-resolver-api/src/main/java/org/eclipse/aether/transform/FileTransformer.java
mcimbora: Included TransformException in the throws clause.

maven-resolver-api/src/main/java/org/eclipse/aether/transform/FileTransformationException.java
mcimbora: Good point, thanks for the explaination and prompt review. The exception now inherits directly from Exception.

rfscholte:
If you want you can add yourself as a contributor in the pomfile as well.

mcimbora:
@rfscholte PR updated, added contributor section to pom.xml.

jira-importer:
Resolve #844

PR #27, opened 2019-03-02
document IllegalArgumentException thrown by CTR

damienburke:


damienburke:
woot! thanks guys. onwards..

PR #33, opened 2019-04-23
[MRESOLVER-87] Refresh examples to use maven-resolver artifacts

slachiewicz:


jira-importer:
Resolve #849

jira-importer:
Resolve #849

PR #41, opened 2020-02-14
MNG-6803: Fix race condition in DefaultArtifactResolver via "Retry"

rreich:
With "Retry" I mean that resolve() looks into the local repository again after performing the downloads.


rreich:
Doesn't fix the error, just makes it less probable.

jira-importer:
Resolve #890

jira-importer:
Resolve #890

PR #42, opened 2020-02-14
MNG-6803: Fix race condition in DefaultArtifactResolver via synchronization

rreich:


rreich:
Continued in #60

jira-importer:
Resolve #890

PR #69, opened 2020-09-12
[MRESOLVER-109] - AndDependencySelector should override toString

michaelboyles:
https://issues.apache.org/jira/browse/MRESOLVER-109

Intentionally omitted tests. These are just for designed debugging and tests would only make it harder to change them. I did test the methods, just not with unit tests

### DIFF
--- a/maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/selector/AndDependencySelector.java
+++ b/maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/selector/AndDependencySelector.java
@@ -209,7 +209,7 @@ public final class AndDependencySelector
     {
         StringBuilder builder = new StringBuilder().append( this.getClass().getSimpleName() ).append( '(' );
         Iterator<? extends DependencySelector> iterator = this.selectors.iterator();
-        while (iterator.hasNext())
+        while ( iterator.hasNext() )
         {
             final DependencySelector selector = iterator.next();
             builder.append( selector.toString() );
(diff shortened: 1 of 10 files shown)


elharo (changes requested):
Needs tests

michaelboyles:
Well, if you insist. Added

elharo (approved):

elharo:
waiting for jenkins: https://ci-builds.apache.org/job/Maven/job/maven-box/job/maven-resolver/job/69/

michael-o (changes requested):
I don't like the usage of antipatterns like `List#get(int)` I would prefer an iterator for testing the last member. If this class is immutable, could this value be precalculated? Does this make sense?

michaelboyles:
Iterator is a better solution. I've changed it, thanks.

Java 7 means no StringJoiner, String.join or streams. Other resolver modules include lang3, which has StringUtils with a join method, but this module doesn't. So using a list was already my 5th choice.

I don't see the value in pre-calculating it. You shouldn't add complexity to a class (i.e. extra fields) for something that's only intended for debugging. As a performance optimization it's also ineffective because you'd be forcing most users to eagerly compute a value they don't use, actually making performance slightly (though insignificantly) worse.

michael-o (changes requested):
This may or may not be a matter of taste, but the the output should not look like a constructor call with `()`. I prefer `SimpleName[...]`.

michaelboyles:
I deferred to copying [Lombok's style](https://projectlombok.org/features/ToString), which is the closest thing I could think of to a standard. Most people are happy with the output of that.

I checked quite a few of the other implementations in the project, there didn't seem to be any particular precedent to follow.

michael-o (approved):

jira-importer:
Resolve #877

jira-importer:
Resolve #877

PR #84, opened 2020-12-27
[MRESOLVER-154] Move SyncContextFactory to SPI module

cstamas:
Move out this iface from impl to spi module.

https://issues.apache.org/jira/browse/MRESOLVER-154


michael-o (approved):
Looks good, but there should be a JIRA issue which shows that the interface has been moved.

jira-importer:
Resolve #926

PR #92, opened 2021-02-02
[MRESOLVER-160] Deprecate service locator

cstamas:
Deprecate service locator implementation as well, as integrators
usually extend this class (see Maven for example), and they don't
deal with deprecated interfaces directly, so the fact they are deprecated
may be missed at first sight.

By deprecating this implementation as well, the integrators
will get a warning in their build and/or IDEs, so intent will be
more obvious.


michael-o (approved):

jira-importer:
Resolve #924

jira-importer:
Resolve #924

PR #110, opened 2021-05-28
[MRESOLVER-184] Add destroy hook

cstamas:
And use this hook in Hz and Redisson implemenetations. Drop redundant
code from HZ as well.


rfscholte:
Is this a candidate for autoclosable?

cstamas:
> Is this a candidate for autoclosable?

@rfscholte um, could you elaborate your question a bit? NamedLockSupport IS AutoCloseable, so what you mean by "this"?

rfscholte:
I was wondering if `destroy` was a synonym for `close`. Only for the latter you can use autoclosable. 
But is seems like that is already covered, so never mind.

cstamas:
Superseded by https://github.com/apache/maven-resolver/pull/111

jira-importer:
Resolve #1416

jira-importer:
Resolve #1416

PR #111, opened 2021-05-29
[MRESOLVER-184] Cleaner destroy

cstamas:
This PR cleans up implementations, aligns method calls, and implementation/backing lock device creation and initialization.

High level changes:
* decouples NamedLockSupport and NamedLockFactorySupport closeLock, really is just name needed
* all locking "implementations" (semaphores, locks) are created in same way and inited on same spot (not scattered across internal inner classes etc)
* keeps things simple: those implementation that need to "destroy" maintain their own map of names to implementation objects.


jira-importer:
Resolve #1416

jira-importer:
Resolve #1416

PR #125, opened 2021-09-10
[MRESOLVER-204] Add a SessionData#computeIfAbsent method

gnodet:


maven-resolver-api/src/main/java/org/eclipse/aether/DefaultSessionData.java
```diff
@@ -81,4 +82,8 @@ public Object get( Object key )
         return data.get( key );
     }
 
+    public Object computeIfAbsent( Object key, Function<Object, Object> mappingFunction )
```
cstamas, line 85: Why not UnaryOperator<Object>?

gnodet: The `computeIfAbsent` method on the `Map` class is a function that receives the key and computes the value.

PR #140, opened 2021-12-14
[MRESOLVER-231] Extend smart checksum support

cstamas:
Refactor and make it pluggable. Also, make it work with Maven Central.

This change broadens "smart checksum" support in Resolver, and
gets them from HTTP response header, if present. So far only
Nexus2 was supported, but this now adds more implementations,
so for example Central is supported now as well, hence, to
download a JAR from Central will now take only 1 HTTP request
and not 2, that effectively **halves count of HTTP requests done
today against central** as well.

This change improves transport-http only, Wagon transport is
a big burden of layers of layers of abstraction of abstraction above
and am unsure how to do this there.

High level changes:
* introduce `ChecksumExtractor` component and provide OOTB 3 implementations for it (the nexus2 was inlined, this is like "factor out nexus2" + add 2 more)
* modify HttpTransporter to use the components
* support nx2, nx3, artifactory, central, google central mirror as all those send inlined checksum

---

https://issues.apache.org/jira/browse/MRESOLVER-231

### DIFF
--- a/maven-resolver-transport-http/src/main/java/org/eclipse/aether/transport/http/HttpTransporter.java
+++ b/maven-resolver-transport-http/src/main/java/org/eclipse/aether/transport/http/HttpTransporter.java
@@ -83,6 +83,8 @@ import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import static java.util.Objects.requireNonNull;
+
 /**
  * A transporter for HTTP/HTTPS.
  */
@@ -94,6 +96,8 @@ final class HttpTransporter
         Pattern.compile( "\\s*bytes\\s+([0-9]+)\\s*-\\s*([0-9]+)\\s*/.*" );
 
     private static final Logger LOGGER = LoggerFactory.getLogger( HttpTransporter.class );
+
+    private final Map<String, ChecksumExtractor> checksumExtractors;
 
     private final AuthenticationContext repoAuthContext;
 
@@ -111,7 +115,9 @@ final class HttpTransporter
 
     private final LocalState state;
 
-    HttpTransporter( RemoteRepository repository, RepositorySystemSession session )
+    HttpTransporter( Map<String, ChecksumExtractor> checksumExtractors,
+                     RemoteRepository repository,
+                     RepositorySystemSession session )
         throws NoTransporterException
     {
         if ( !"http".equalsIgnoreCase( repository.getProtocol() )
@@ -119,6 +125,7 @@ final class HttpTransporter
         {
             throw new NoTransporterException( repository );
         }
+        this.checksumExtractors = requireNonNull( checksumExtractors, "checksum extractors must not be null" );
         try
         {
             this.baseUri = new URI( repository.getUrl() ).parseServerAuthority();
@@ -264,22 +271,58 @@ final class HttpTransporter
     protected void implGet( GetTask task )
         throws Exception
     {
+        boolean resume = true;
+        boolean applyChecksumExtractors = true;
+
         EntityGetter getter = new EntityGetter( task );
         HttpGet request = commonHeaders( new HttpGet( resolve( task ) ) );
-        resume( request, task );
-        try
+        while ( true )
         {
-            execute( request, getter );
-        }
-        catch ( HttpResponseException e )
-        {
-            if ( e.getStatusCode() == HttpStatus.SC_PRECONDITION_FAILED && request.containsHeader( HttpHeaders.RANGE ) )
+            try
             {
-                request = commonHeaders( new HttpGet( request.getURI() ) );
+                if ( resume )
+                {
+                    resume( request, task );
+                }
+                if ( applyChecksumExtractors )
+                {
+                    for ( ChecksumExtractor checksumExtractor : checksumExtractors.values() )
+                    {
+                        checksumExtractor.prepareRequest( request );
+                    }
+                }
                 execute( request, getter );
-                return;
+                break;
             }
-            throw e;
+            catch ( HttpResponseException e )
+            {
+                if ( resume && e.getStatusCode() == HttpStatus.SC_PRECONDITION_FAILED
+                        && request.containsHeader( HttpHeaders.RANGE ) )
+                {
+                    request = commonHeaders( new HttpGet( resolve( task ) ) );
+                    resume = false;
+                    continue;
+                }
+                if ( applyChecksumExtractors )
+                {
+                    boolean retryWithoutExtractors = false;
+                    for ( ChecksumExtractor checksumExtractor : checksumExtractors.values() )
+                    {
+                        if ( checksumExtractor.retryWithoutExtractor( e ) )
+                        {
+                            retryWithoutExtractors = true;
+                            break;
+                        }
+                    }
+                    if ( retryWithoutExtractors )
+                    {
+                        request = commonHeaders( new HttpGet( resolve( task ) ) );
+                        applyChecksumExtractors = false;
+                        continue;
+                    }
+                }
+                throw e;
+            }
         }
     }
 
@@ -535,7 +578,8 @@ final class HttpTransporter
             }
 
             long offset = 0L, length = entity.getContentLength();
-            String range = getHeader( response, HttpHeaders.CONTENT_RANGE );
+            Header rangeHeader = response.getFirstHeader( HttpHeaders.CONTENT_RANGE );
+            String range = rangeHeader != null ? rangeHeader.getValue() : null;
             if ( range != null )
             {
                 Matcher m = CONTENT_RANGE_PATTERN.matcher( range );
@@ -559,24 +603,16 @@ final class HttpTransporter
 
         private void extractChecksums( HttpResponse response )
         {
-            // Nexus-style, ETag: "{SHA1{d40d68ba1f88d8e9b0040f175a6ff41928abd5e7}}"
-            String etag = getHeader( response, HttpHeaders.ETAG );
-            if ( etag != null )
+            for ( Map.Entry<String, ChecksumExtractor> extractorEntry : checksumExtractors.entrySet() )
             {
-                int start = etag.indexOf( "SHA1{" ), end = etag.indexOf( "}", start + 5 );
-                if ( start >= 0 && end > start )
+                Map<String, String> checksums = extractorEntry.getValue().extractChecksums( response );
+                if ( checksums != null )
                 {
-                    task.setChecksum( "SHA-1", etag.substring( start + 5, end ) );
+                    checksums.forEach( task::setChecksum );
+                    return;
                 }
             }
         }
-
-        private String getHeader( HttpResponse response, String name )
-        {
-            Header header = response.getFirstHeader( name );
-            return ( header != null ) ? header.getValue() : null;
-        }
-
     }
 
     private class PutTaskEntity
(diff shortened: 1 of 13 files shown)


maven-resolver-transport-http/src/main/java/org/eclipse/aether/transport/http/DigestChecksumExtractor.java
```diff
@@ -0,0 +1,81 @@
...
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.client.methods.HttpUriRequest;
+
+import javax.inject.Named;
+import javax.inject.Singleton;
+import java.util.Collections;
+import java.util.Map;
+
+/**
+ * A component extracting Digest header value from response.
```
kwin, line 33: A link to the draft spec would be useful here: https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-digest-headers-07. Maybe there is also somewhere a reference that Maven Central supports this?

kwin:
This is a really nice and smart feature. Would be good to adjust the documentation at https://github.com/apache/maven-resolver/blob/master/src/site/markdown/configuration.md for `aether.connector.smartChecksums` as well, though.

maven-resolver-transport-http/src/main/java/org/eclipse/aether/transport/http/DigestChecksumExtractor.java
cstamas: Agreed, but we discussed this on ASF Slack: I really added Digest support just as a "showcase" :smile: so I may even remove this, as even myself am unsure what service supports it. Central does not support it, central uses `x-checksum-XXX` headers instead (or is that Varnish? Unsure)

cstamas:
> This is a really nice and smart feature. Would be good to adjust the documentation at https://github.com/apache/maven-resolver/blob/master/src/site/markdown/configuration.md for `aether.connector.smartChecksums` as well, though.

Yup, I agree here. Doco is wrong, but this (Nexus2 exclusive) feature was present since eons in resolver (former Sonatype Aether), and this PR merely refactors it and makes it extensible. All former resolver releases supported this (but exclusively for Nexus2). and this PR makes it more universal, extensible.

michael-o (changes requested):

src/site/markdown/smart-checksum-strategies.md
```diff
@@ -39,12 +39,15 @@ way. Naturally, this means only SHA-1 is available in artifact response header.
 Emitted by: Sonatype Nexus2 only.
 
 
-## Non-standard headers
+## Non-standard `X-` headers
 
-Some MRMs and Maven Central emit headers `x-checksum-sha1` and `x-checksum-md5`. Resolver
-will detect these and use their value.
+Maven Central emit headers `x-checksum-sha1` and `x-checksum-md5` along with artifact response.
```
michael-o, line 44: Maven Central emits

src/site/markdown/smart-checksum-strategies.md
```diff
@@ -39,12 +39,15 @@ way. Naturally, this means only SHA-1 is available in artifact response header.
 Emitted by: Sonatype Nexus2 only.
 
 
-## Non-standard headers
+## Non-standard `X-` headers
 
-Some MRMs and Maven Central emit headers `x-checksum-sha1` and `x-checksum-md5`. Resolver
-will detect these and use their value.
+Maven Central emit headers `x-checksum-sha1` and `x-checksum-md5` along with artifact response. 
+Google GCS on the other hand uses `x-goog-meta-checksum-sha1` and `x-goog-meta-checksum-md5` heaers.
```
michael-o, line 45: headers

src/site/markdown/smart-checksum-strategies.md
```diff
@@ -39,12 +39,15 @@ way. Naturally, this means only SHA-1 is available in artifact response header.
...
+Resolver will detect these and use their value.
 
-Emitted by: Maven Central and Artifactory, maybe by some others as well.
+Emitted by: Maven Central and some MRMs and . Kinda , maybe by some others as well.
```
michael-o, line 48: This sentence is weird

src/site/markdown/smart-checksum-strategies.md
```diff
@@ -0,0 +1,56 @@
...
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+By default, Resolver will fetch the payload checksum from remote repository. These
+checksums are used to enforce transport validity (that download is not corrupted).
+
+This implies, that to get one artifact, resolver needs to issue two HTTP reuest:
```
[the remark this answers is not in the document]
cstamas: Did not want to get into detail, I just wanted to explain "resolver to get one -- of those you enumerated -- needs to make two HTTP request", Unsure is detailing it such is helpful... or just call it "payload"?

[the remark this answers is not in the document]
michael-o: Since we have two elements in Resolver: Artifacts and Metadata, use them. Remember that `SyncContext` uses them too.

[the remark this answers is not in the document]
cstamas: Fixed

maven-resolver-transport-http/src/main/java/org/eclipse/aether/transport/http/HttpTransporterFactory.java
```diff
@@ -37,17 +42,39 @@
 public final class HttpTransporterFactory
     implements TransporterFactory
 {
+    private static final Map<String, ChecksumExtractor> EXTRACTORS;
+
+    static
+    {
+        HashMap<String, ChecksumExtractor> map = new HashMap<>();
+        map.put( NexusChecksumExtractor.NAME, new NexusChecksumExtractor() );
```
michael-o, line 50: Does this apply for Nexus 2 an 3?

maven-resolver-transport-http/src/main/java/org/eclipse/aether/transport/http/XChecksumChecksumExtractor.java
```diff
@@ -0,0 +1,86 @@
...
+        }
+        // Central style: x-goog-meta-checksum-sha1: 9ad0d8e3482767c122e85f83567b8ce6
+        value = extractChecksum( response, "x-goog-meta-checksum-md5" );
+        if ( value != null )
+        {
+            return Collections.singletonMap( "MD5", value );
+        }
+
+        return null;
+    }
```
michael-o, line 79: I am confused, can't multiple headers appear on the response? Why return on the first found?

michael-o (changes requested):
This entire smart checksum notion needs to be named to included, since this is how it is named in the code and nothing is smart here.
The config property will be deprecated and replaced with a new one.

maven-resolver-transport-http/src/main/java/org/eclipse/aether/transport/http/HttpTransporterFactory.java
cstamas: Nx2 only, fixing

maven-resolver-transport-http/src/main/java/org/eclipse/aether/transport/http/XChecksumChecksumExtractor.java
cstamas: Yes, they can, but is irrelevant: first present (SHA1 or MD5) IF present, will be MATCHED and will match/fail. So, it is always "first matched" so even if both sha1 and md5 present, and sha1 not matches, md5 will not be tried for at all, so to say, is "not worth" to have it. Still, you are correct, if response contains both, let we extract both, and finally, this may be useful in some "exotic" user setup (for example if user explicitly sets checksums as "MD5, SHA-1" -- so for some reason reverts the orders from those defaulted).

maven-resolver-transport-http/src/main/java/org/eclipse/aether/transport/http/HttpTransporter.java
```diff
@@ -264,22 +271,47 @@ protected void implPeek( PeekTask task )
     protected void implGet( GetTask task )
         throws Exception
     {
+        boolean resume = true;
+        boolean smartChecksum = true;
```
michael-o, line 275: This should be renamed.

src/site/markdown/configuration.md
```diff
@@ -39,7 +39,7 @@ Option | Type | Description | Default Value | Supports Repo ID Suffix
...
 `aether.connector.resumeThreshold` | long | The size in bytes which a partial download needs to have at least to be resumed. Requires `aether.connector.resumeDownloads` to be `true` to be effective. | `64 * 1024` | yes
 `aether.connector.requestTimeout` | long | Request timeout in milliseconds. | `1800000` | yes
-`aether.connector.smartChecksums` | boolean | Flag indicating that instead of comparing the explicit checksum from the remote repo with the calculated one it will try to extract the reference checksum from the actual artifact requests's response header named `ETag` in format `{SHA1{<checksum>}}`. This only works for HTTP(S) requests and certain transport extensions. In addition it only supports SHA-1. | `true` | no
+`aether.connector.smartChecksums` | boolean | Flag indicating that instead of comparing the explicit checksum fetched from the remote repo with the calculated one, it should try to extract the reference checksum from the actual artifact requests's response headers (several (strategies supported)[smart-checksum-strategies.html]). This only works for transport-http transport. | `true` | no
```
michael-o, line 42: This link is now broken.

src/site/markdown/included-checksum-strategies.md
```diff
@@ -0,0 +1,61 @@
...
+## Non-standard `X-` headers
+
+Maven Central emits headers `x-checksum-sha1` and `x-checksum-md5` along with artifact response. 
+Google GCS on the other hand uses `x-goog-meta-checksum-sha1` and `x-goog-meta-checksum-md5` 
+headers. Resolver will detect these and use their value.
+
+Emitted by: Maven Central, GCS, some CDNs and probably more.
+
+
+## Digest Headers (draft)
```
michael-o, line 54: I wonder whether this should be backed out until it is a real RFC then should be renamed to `RfcXXXXChecksumExtractor`. I still need to read the draft, it is massive.

michael-o (changes requested):

src/site/markdown/included-checksum-strategies.md
cstamas: Let's drop it, having covered Maven Central , google Mirror and all major MRMs is still massive.

src/site/markdown/configuration.md
```diff
@@ -39,7 +39,7 @@ Option | Type | Description | Default Value | Supports Repo ID Suffix
...
 `aether.connector.resumeThreshold` | long | The size in bytes which a partial download needs to have at least to be resumed. Requires `aether.connector.resumeDownloads` to be `true` to be effective. | `64 * 1024` | yes
 `aether.connector.requestTimeout` | long | Request timeout in milliseconds. | `1800000` | yes
-`aether.connector.smartChecksums` | boolean | Flag indicating that instead of comparing the explicit checksum fetched from the remote repo with the calculated one, it should try to extract the reference checksum from the actual artifact requests's response headers (several (strategies supported)[smart-checksum-strategies.html]). This only works for transport-http transport. | `true` | no
+`aether.connector.smartChecksums` | boolean | Flag indicating that instead of comparing the explicit checksum fetched from the remote repo with the calculated one, it should try to extract the reference checksum from the actual artifact requests's response headers (several (strategies supported)[included-checksum-strategies.html]). This only works for transport-http transport. | `true` | no
```
michael-o, line 42: Note that those are named "external" and not explicit anymore.

michael-o (approved):

jira-importer:
Resolve #1405

PR #146, opened 2022-01-20
[1.6.x][MRESOLVER-235] Fix dependency management issue

gnodet:
This PR provides a fix for MRESOLVER-235 / MNG-7003.
However, the fix breaks two ITs: mng4720 and mng4396.
For mng4720, so core of the IT is not broken, but as specifically commented in the IT see https://github.com/apache/maven-integration-testing/blob/master/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4720DependencyManagementExclusionMergeTest.java#L70, one of the dependency is not included anymore.
The PR contains unit tests that replicate the ITs for mng4720 and for MNG-7003 (available at https://github.com/gnodet/maven-integration-testing/tree/MNG-7003).  I'm not sure there's any way to fix both at the same time.

### DIFF
--- a/maven-resolver-impl/src/test/java/org/eclipse/aether/internal/impl/collect/DefaultDependencyCollectorTest.java
+++ b/maven-resolver-impl/src/test/java/org/eclipse/aether/internal/impl/collect/DefaultDependencyCollectorTest.java
@@ -36,6 +36,8 @@ import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import java.util.TreeSet;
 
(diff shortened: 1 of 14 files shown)


cstamas:
Resolver has 3 DependencyManager implementations: "classic" (so far all Maven 3+ versions used it), "transitive" (almost same as "classic" sans the 2 depth thing) and "default" (unsure about the intent here).

The issue PR would modify the "classic" depMgr, while it would then become (almost) same as "transitive" is. Hence, am rejecting the PR, and the issue altogether, as to me this seems like just a need for another depMgr exposed in Maven?

See:
* https://issues.apache.org/jira/browse/MRESOLVER-235
* https://github.com/apache/maven/pull/1000

PR #159, opened 2022-03-15
[MRESOLVER-236] Add demo doing signature resolution w/ FAIL policy

cstamas:
And use FAIL checksum policy. This should prove that resolving classic
GPG (or any other) signature is now possible.

---

https://issues.apache.org/jira/browse/MRESOLVER-236


jira-importer:
Resolve #1404

PR #160, opened 2022-03-17
[MRESOLVER-248] DependencyCollector: let DF vs BF implementations coexists

cstamas:
Revive pre 69aeda6c7976e5ff39ed7d0f888e4895a50a7f75 commit DF DependencyCollector, and keep the BF one. as well. 

High level changes:
* for now, for simplicity sake, _copied_ the classes (changes or deleted) to bf and df subpackage.
* DefaultDependencyCollector is now just an indirector, that delegates based on session config
* the `aether.collector.impl` introduced config property (names: "df", "bf") can be used to select collector implementation
* default is as pre BF commit, "df"

---

https://issues.apache.org/jira/browse/MRESOLVER-248


cstamas:
@caiwei-ebay ping here

kwin:
I think this deserves a JIRA ticket with some reasoning why this should be reintroduced. Also the new config property should be documented at https://github.com/apache/maven-resolver/blob/master/src/site/markdown/configuration.md.

cstamas:
> MRESOLVER-247: Introduce skipper to avoid unnecessary dependency reso… #158

Sure, this PR is merely to simplify testing/comparison phase, for explanation of "why" see here (and above for perf test results): https://github.com/apache/maven-resolver/pull/158#issuecomment-1070591546

In short: My (incomplete) perf testing showed slight but measurable degradation of performance, so the coexistence is here only to simplify testing. Once we decide what should happen (keep this, keep that, or keep both) a complete overhaul, doco and other update will happen. Still, right now am leaning to keep both, at least until we see related changes done (BFS + skipper + parallel POM download)...

cstamas:
Superseded by https://github.com/apache/maven-resolver/pull/161

PR #163, opened 2022-04-10
MRESOLVER-133 unnessary poms downloadings for dependency using versio…

caiwei-ebay:
…n range

### DIFF
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/bf/BfDependencyCollector.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/bf/BfDependencyCollector.java
@@ -452,6 +452,7 @@ public class BfDependencyCollector
                     {
                         doRecurse( args, context.withDependency( d ), descriptorResult, child );
                     }
+                    break;
                 }
             }
             else
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/df/DfDependencyCollector.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/df/DfDependencyCollector.java
@@ -368,6 +368,7 @@ public class DfDependencyCollector
             return;
         }
 
+        Collections.reverse( versions );
         for ( Version version : versions )
         {
             Artifact originalArtifact = dependency.getArtifact().setVersion( version.toString() );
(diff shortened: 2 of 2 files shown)


caiwei-ebay:
@michael-o @cstamas @ibabiankou @cannucklehead

This is the draft PR for MRESOLVER-133.

The idea is below as referenced from Ivan's comments in the ticket: 
> Below loop should iterate over reversed list of versions, it should only try to get artifact descriptor and break as soon as got one successfully.
https://github.com/apache/maven-resolver/blob/master/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/bf/BfDependencyCollector.java#L394

The same changes applied to original DF solution.

cstamas:
A hard NO for now, as this needs more investigating.

First of all, if you add break here, you are changing the semantics of resolver _collection_ (would return "highest version" not "all applicable"). Second, your goal to get "highest version" is already doable w/ resolver and is present in UTs, look at these tests:
https://github.com/apache/maven-resolver/blob/915186d643f37cff9a217a482c62d9b47767c95d/maven-resolver-impl/src/test/java/org/eclipse/aether/internal/impl/collect/df/DfDependencyCollectorTest.java#L563
https://github.com/apache/maven-resolver/blob/915186d643f37cff9a217a482c62d9b47767c95d/maven-resolver-impl/src/test/java/org/eclipse/aether/internal/impl/collect/bf/BfDependencyCollectorTest.java#L570

So both UT end up in loop (you'd break) with SINGLE version, due to set `HighestVersionFilter`. Meaning, Resolver already have the means to achieve this if asked for. OTOH, Maven "as resolver client" does NOT use version filters, but it asks for "all" and then "mediates" based on it's own rules. This is sub-otpimal of course, but this requires more careful research to not break resolver and/or Maven, at least 3.x line.

See these issues for some context please:
https://issues.apache.org/jira/browse/MRESOLVER-164
https://issues.apache.org/jira/browse/MNG-7049

And an experiment done here:
https://github.com/cstamas/maven/commit/99709fd8a715101e5c64ab99c651a468b37f0d15

In short: resolver CAN do this out of the box, but Maven does not USE this feature. OTOH, there are cases in Maven itself, where Maven DOES THIS, see `org.apache.maven.repository.internal.DefaultModelResolver#resolveModel(org.apache.maven.model.Dependency)` for example, so we need first to figure out why was it done like this (maybe just lack of time to properly finish integration) but it may be something else as well (like "closest wins" strategy).

jira-importer:
Resolve #889

PR #164, opened 2022-04-11
[MRESOLVER-249] Update Hazelcast to 5.1.1

cstamas:
Update Hazelcast from 5.0.2 to latest 5.1.1. This is really
not a game changer, as Hazelcast 5.x is API compatible, so
end users may choose which version to use for real at runtime.

---

https://issues.apache.org/jira/browse/MRESOLVER-249


michael-o (changes requested):
This one is missing: https://maven.apache.org/resolver/maven-resolver-named-locks-hazelcast/index.html

michael-o (approved):
Please add the HZ versoin to the commit title and the JIRA issue as well.

PR #166, opened 2022-04-12
[MRESOLVER-250] Wrap DataPool descriptors into weak refs

cstamas:
The DataPool.descriptors in case of "good" descriptor references it's own key, hence prevents collection of it. Fixing by wrapping the value (descriptor) with `WeakReference`, as proposed user on related JIRA.

---

https://issues.apache.org/jira/browse/MRESOLVER-250

### DIFF
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/DataPool.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/DataPool.java
@@ -19,6 +19,7 @@ package org.eclipse.aether.internal.impl.collect;
  * under the License.
  */
 
+import java.lang.ref.WeakReference;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
@@ -66,7 +67,7 @@ public final class DataPool
 
     private ObjectPool<Dependency> dependencies;
 
-    private Map<Object, Descriptor> descriptors;
+    private Map<Object, WeakReference<Descriptor>> descriptors;
 
     private final Map<Object, Constraint> constraints = new HashMap<>();
 
@@ -81,7 +82,7 @@ public final class DataPool
         {
             artifacts = (ObjectPool<Artifact>) cache.get( session, ARTIFACT_POOL );
             dependencies = (ObjectPool<Dependency>) cache.get( session, DEPENDENCY_POOL );
-            descriptors = (Map<Object, Descriptor>) cache.get( session, DESCRIPTORS );
+            descriptors = (Map<Object, WeakReference<Descriptor>>) cache.get( session, DESCRIPTORS );
         }
 
         if ( artifacts == null )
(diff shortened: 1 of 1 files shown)


michael-o (approved):

psiroky:
I _believe_ this change is the cause behind the "slowness" I am seeing when comparing Maven 3.9.0 and 3.8.7. See https://lists.apache.org/thread/r9p236z8kvqqk7ykvkgmc5wgps6n1hkf for some numbers comparing couple of different builds on 3.9.0 and 3.8.7.

From what I was able to observe (using async-profiler and JFR/JMC) the number of created instances of `org.eclipse.aether.artifact.DefaultArtifact` is about 4 times higher with 3.9.0 (and the logic in the constructor is somewhat "complex", merging maps, which takes time and creates quite a lot "garbage"). For small projects this is not really noticeable (e.g. if you create 1k or 4k instances that is not something to noticeable slow down the build). However, for big multi-module projects (like e.g. Quarkus which I am using for testing), the difference is ~2mil objects vs ~8mil objects and that adds up.

I have tried to revert this commit (on current maven-resolver master, since it is basically the same as version 1.9.4 which is used in 3.9.0) and then build the 3.9.x branch with that version (1.9.5-SNAPSHOT) -- and the result is that the build times are basically equal to those of 3.8.7.

At this point I am not saying this change is bad or so, maybe we need to do adjustments elsewhere. Also, even ~2mil created `org.eclipse.aether.artifact.DefaultArtifact` seems like a lot, so maybe there is a space for some more aggressive caching (I am just speculating, since I have not studied the code in depth yet).

Sorry for hijacking this old PR, but I wanted to let you know asap. I am planning to create a proper JIRA, but before that I need to gather a bit more data and double-check my findings.

cstamas:
This explanation sounds quite plausible, basically the lack of weak ref (added in this PR) preserved DefaultArtifact instances, that resulted in less instance to be created, less GC pressure.... OTOH, by looking at original JIRA, am unsure that "I observed a high heap consumption" is a real show stopper (I mean, if you do resolve huge dep trees, maybe you can expect somewhat bigger memory usage as well?). Ideally we could get some feedback from reporter...

cstamas:
@psiroky given you have setup already, did you try BF collector as well? 
(run mvn with `-Daether.dependencyCollector.impl=bf`)

cstamas:
Good, thanks! Now we learnt two things: not only this above @psiroky found out, but finally we have a "guinea pig" project (quarkus) that is big enough, to present notable diff between the DF (original) and BF (new) collector. 
Two points with one kick :smile:

psiroky:
I created https://issues.apache.org/jira/browse/MNG-7677 to track this. I think it will be better to continue the discussion there. I did create it under MNG, since I am not 100% sure this maven-resolver PR is the cause and also maybe, there are other things related directly in Maven (and I did notice it in the Maven itself anyway).

jira-importer:
Resolve #1390

PR #174, opened 2022-05-06
Add Test cases for the repair of BF algorithm of dependent package wh…

weibo1995:
Add Test cases for the repair of BF algorithm of dependent package when children are empty（ https://github.com/apache/maven-resolver/pull/172 ).

After adding, it covers two cases.

The first is that  the dependent package of no children  is on the left. The dependencies are as follows:
image
The expected final dependency tree after conflict resolving is as follows:
image

The second is that  the dependent package of no children  is on the right. The dependencies are as follows:
image

The expected final dependency tree after conflict resolving is as follows:
image

thanks.

Signed-off-by: hongbo.weihb <hongbo.weihb@alibaba-inc.com>


cstamas:
Superseded by https://github.com/apache/maven-resolver/pull/175

PR #183, opened 2022-06-22
Example of using japicmp to ensure binary API compatibility

cstamas:
This now fails when building, as 1.8.x breaks vs 1.6.3
(and is ok, this is merely a "demo" that we should integrate)


michael-o (approved):

cstamas:
https://github.com/siom79/japicmp/pull/326

cstamas:
https://github.com/siom79/japicmp/pull/328

cstamas:
Jenkins still uses Maven 3.8.5 and it fails due https://github.com/siom79/japicmp/issues/327

cstamas:
This is on hold until https://github.com/siom79/japicmp/pull/328 gets merged and we got new release with that fix

cstamas:
Updated to latest JApiCmp that has all the patches.

PR #187, opened 2022-06-24
[MRESOLVER-265] Checksum of checksum

cstamas:
There is a discrepancy when a checksum is attached (so not added
by resolver itself), and resolver will checksum it if the algorithm
is not configured (ie. default config used but SHA512 added).

This makes it work as expected, as checksum recognition works
now against ALL SUPPORTED and not configured ONLY checksums).

---

https://issues.apache.org/jira/browse/MRESOLVER-265

### DIFF
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/Maven2RepositoryLayoutFactory.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/Maven2RepositoryLayoutFactory.java
@@ -142,6 +142,7 @@ public final class Maven2RepositoryLayoutFactory
         }
 
         return new Maven2RepositoryLayout(
+                new ArrayList<>( checksumAlgorithmFactorySelector.getChecksumAlgorithmFactories() ),
                 checksumsAlgorithms,
                 omitChecksumsForExtensions
         );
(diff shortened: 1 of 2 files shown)


kwin:
I am pretty sure some older Maven repository implementations (e.g. Nexus2) always require MD5/SHA1 (even for sha512 files). Not sure about the implications of this change with those repo managers.

cstamas:
> I am pretty sure some older Maven repository implementations (e.g. Nexus2) always require MD5/SHA1 (even for sha512 files). Not sure about the implications of this change with those repo managers.

I am pretty sure this is not true for deploy (to some hosted repository), is it possible you talk about staging maybe? As in that case, Nexus2 (on repo.a.o but also OSS.s.o) enforce some things, and yes, I'd expect some strangeness here.... but given it's closed source, they should fix it IMHO :smile:

kwin:
These are the staging rule conflicts I am aware of: https://issues.apache.org/jira/browse/INFRA-14923

cstamas:
Seems https://issues.sonatype.org/browse/NEXUS-21802 is fixed, and ASF NX2 is 2.14.20 and oss.s.o is 2.15 so both should be fine.

jira-importer:
Resolve #758

PR #188, opened 2022-06-24
[MRESOLVER-266] Simplify named lock adapter creation and align config source

cstamas:
Simplified adapter creation and now as everything else in resolver, session
config properties are used as configuration source, not Java system
properties.

---

https://issues.apache.org/jira/browse/MRESOLVER-266


jira-importer:
Resolve #756

PR #202, opened 2022-10-12
add test to ensure that a version constraint having same upper and lower

kwin:
bound works

compare with MNG-7106


cstamas:
Honestly, I was surprised by the existence of archaic maven-artifact `org.apache.maven.artifact.versioning` package, that is not even compatible (functionality wise) with resolver Version (and it's schemes). Personally, I'd just deprecate all this package as is.

kwin:
@cstamas Are you fine with committing this additional test case?

PR #209, opened 2022-10-27
[MRESOLVER-278] On repository system shutdown

cstamas:
Drops problematic onSessionClose, keeps only onSystemClose support. Issue will be adjusted accordingly.

Key changes:
* undone all related to onSessionClose
* kept onRepositorySystemClose + shutdown
* all 3 users on onSessionClose (locking + 2 "recording" features) will act on shutdown
* moved out of session.data use as it is completely unreliable, now singleton components hold caches (hence the shutdown persistence)
* DefaultSyncContext: fixed as explained, does not use system properties anymore, but is made pretty much same as before https://github.com/apache/maven-resolver/pull/196
* renamed all new components to `camelCase`

----

https://issues.apache.org/jira/browse/MRESOLVER-278


jira-importer:
Resolve #954

jira-importer:
Resolve #954

PR #214, opened 2022-11-02
[EXPERIMENT] Do what Java does: retry and hope the best

cstamas:
Java in bugfix for [1] does very same thing: on
AccessDeniedEx it just retries. Here, the existing Retry utility is extended, and now "retry for given time" new method added "retry for given attempts" as well. The count of attempts and sleep times between attempts (probably unimportant) are exposed as system properties.

[1] https://bugs.openjdk.org/browse/JDK-8252883

Patch experiment for https://github.com/apache/maven-mvnd/issues/728


cstamas:
This fix is reported as working as well.

cstamas:
Superseded by https://github.com/apache/maven-resolver/pull/216

PR #218, opened 2022-11-11
[MRESOLVER-290] Expand use of atomic file ops

cstamas:
Drop old and legacy code doing smart(er) things, use common shared primitives instead.

Also, make IO handling stricter, do not be "forgiving" by swallowing IOEx-es as they will inevitably lead to failures, it does not make resolver more robust, but actually makes it harder to figure out the problems, and could lead to incomplete downloads or uploads. Simply put, if reading or writing a file on disk fails with IOEx, do not assume "is fine" in any case. This probably shows some existing issue like permissions or whatever.

---

https://issues.apache.org/jira/browse/MRESOLVER-290

### DIFF
--- a/maven-resolver-connector-basic/src/main/java/org/eclipse/aether/connector/basic/BasicRepositoryConnector.java
+++ b/maven-resolver-connector-basic/src/main/java/org/eclipse/aether/connector/basic/BasicRepositoryConnector.java
@@ -25,6 +25,7 @@ import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.UncheckedIOException;
 import java.net.URI;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -455,9 +456,7 @@ final class BasicRepositoryConnector
         protected void runTask()
                 throws Exception
         {
-            fileProcessor.mkdirs( file.getParentFile() );
-
-            try ( FileUtils.TempFile tempFile = FileUtils.newTempFile( file.toPath() ) )
+            try ( FileUtils.CollocatedTempFile tempFile = FileUtils.newTempFile( file.toPath() ) )
             {
                 final File tmp = tempFile.getPath().toFile();
                 listener.setChecksumCalculator( checksumValidator.newChecksumCalculator( tmp ) );
@@ -489,7 +488,7 @@ final class BasicRepositoryConnector
                         }
                     }
                 }
-                fileProcessor.move( tmp, file );
+                tempFile.move();
                 if ( persistedChecksums )
                 {
                     checksumValidator.commit();
@@ -602,6 +601,7 @@ final class BasicRepositoryConnector
             catch ( IOException e )
             {
                 LOGGER.warn( "Failed to upload checksums for {}", file, e );
+                throw new UncheckedIOException( e );
             }
         }
 
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultFileProcessor.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultFileProcessor.java
@@ -19,20 +19,23 @@ package org.eclipse.aether.internal.impl;
  * under the License.
  */
 
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileOutputStream;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
-import java.nio.ByteBuffer;
-import java.nio.charset.StandardCharsets;
-
 import javax.inject.Named;
 import javax.inject.Singleton;
 
+import java.io.BufferedInputStream;
+import java.io.BufferedOutputStream;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.io.UncheckedIOException;
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+
 import org.eclipse.aether.spi.io.FileProcessor;
 import org.eclipse.aether.util.ChecksumUtils;
+import org.eclipse.aether.util.FileUtils;
 
 /**
  * A utility class helping with file-based operations.
@@ -40,7 +43,7 @@ import org.eclipse.aether.util.ChecksumUtils;
 @Singleton
 @Named
 public class DefaultFileProcessor
-    implements FileProcessor
+        implements FileProcessor
 {
 
     /**
@@ -50,7 +53,7 @@ public class DefaultFileProcessor
      *
      * @param directory The directory to create, may be {@code null}.
      * @return {@code true} if and only if the directory was created, along with all necessary parent directories;
-     *         {@code false} otherwise
+     * {@code false} otherwise
      */
     public boolean mkdirs( File directory )
     {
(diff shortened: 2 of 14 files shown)


maven-resolver-util/src/main/java/org/eclipse/aether/util/FileUtils.java
```diff
@@ -44,9 +44,23 @@ private FileUtils()
...
     public interface TempFile extends Closeable
     {
+        /**
+         * Returns the path of the created temp file.
+         */
         Path getPath();
     }
 
+    /**
+     * A collocated temporary file, that resides next to a "target" file, and is removed when closed.
```
michael-o, line 54: colocated and not collocated?!

michael-o (changes requested):

maven-resolver-connector-basic/src/main/java/org/eclipse/aether/connector/basic/BasicRepositoryConnector.java
```diff
@@ -489,7 +488,7 @@ protected void runTask()
                         }
                     }
                 }
-                fileProcessor.move( tmp, file );
```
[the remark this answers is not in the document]
cstamas: Yes, is unused (connector was only one using it), hence I deprecated it (as it is in SPI that is public layer)

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultFileProcessor.java
```diff
@@ -75,144 +78,49 @@ public boolean mkdirs( File directory )
...
-                {
-                    out.close();
-                }
-            }
-            catch ( final IOException e )
-            {
-                // Suppressed due to an exception already thrown in the try block.
-            }
-        }
+        FileUtils.writeFile( target.toPath(), p -> Files.write( p, data.getBytes( StandardCharsets.UTF_8 ) ) );
```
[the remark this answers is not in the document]
cstamas: I would actually go other way around: make almost everything use FileProcessor (and extend it with some of things coming from FileUtils), as FileProcessor is a component, and hence, is replaceable. So, one could if needed, wrap the original component and "listen" for file changes happening. This is impossible with static helper....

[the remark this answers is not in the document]
cstamas: Not now but in near future https://issues.apache.org/jira/browse/MRESOLVER-295

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultFileProcessor.java
```diff
@@ -241,8 +149,10 @@ private long copy( OutputStream os, InputStream is, ProgressListener listener )
         return total;
     }
 
+    @Deprecated
```
slawekjaranowski, line 152: We should add description - why is deprecated, what replacement ... and so.
later nobody will remember a reason

cstamas: I undid deprecation

cstamas: Or missed this one?

cstamas: fixed

cstamas:
There's much more to do, but am stopping here due https://issues.apache.org/jira/browse/MRESOLVER-295.
My plan is following:
* component should use component (even if trivial as Slawek noted) to perform IO. Reason: components can be swapped out
* split FileProcessor into PathProcessor (as name says, for files) and ChecksumProcessor (to read write checksums ONLY, to not mix concerns). This opens up the possibility to implement different ChecksumWriters (BSD/GNU/whatever), as currently we read BSD/GNU/resolver-native but write only resolver-native format.
* but, due module layout (SPI not depending on Utils, and both Impl and Connector depending on SPI but Connector not on Impl) this needs a bit more work: see how many implementations of FileProcessor we have across, exactly for this reason. So, I want to do something with it.
* direct use of static helper reduces testability, there is a nice test where file copy fails, that cannot be done if directly static helper used.

slawekjaranowski (approved):

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultFileProcessor.java
```diff
@@ -75,144 +78,49 @@ public boolean mkdirs( File directory )
...
-                {
-                    out.close();
-                }
-            }
-            catch ( final IOException e )
-            {
-                // Suppressed due to an exception already thrown in the try block.
-            }
-        }
+        FileUtils.writeFile( target.toPath(), p -> Files.copy( source, p ) );
```
michael-o, line 97: Do we actually need oneline wrappers for `Files` at all?

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultTrackingFileManager.java
```diff
@@ -41,66 +42,55 @@
...
+            try ( InputStream stream = Files.newInputStream( filePath ) )
             {
-                return null;
+                Properties props = new Properties();
+                props.load( stream );
+                return props;
+            }
+            catch ( IOException e )
+            {
+                LOGGER.warn( "Failed to read tracking file '{}'", file, e );
```
michael-o, line 63: This pattern should be then consistent throughout, means file param should be always in single quotes.

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultTrackingFileManager.java
```diff
@@ -41,66 +42,55 @@
...
-        finally
-        {
-            close( stream, file );
         }
-
         return null;
     }
 
     @Override
     public Properties update( File file, Map<String, String> updates )
```
michael-o, line 71: Since this is run under `SyncContext` it is guaranteed to be locked, so atomic moves aren't required?!

michael-o (changes requested):

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultTrackingFileManager.java
cstamas: Atomic files are not to circumvent locking at all, they are to guarantee that all (like downloads) are "all or nothing" even in case of errors, resolver will not leave partially written file under any circumstances (as partially written tmp file is deleted if fails).

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultFileProcessor.java
cstamas: File|Utils leverages TempFile, there is no equivalent for it in Files...

michael-o: Right...

michael-o (approved):

jira-importer:
Resolve #965

PR #223, opened 2022-11-18
[MRESOLVER-297] Chained LRM

cstamas:
Not for production, but useful in testing environments.

---

https://issues.apache.org/jira/browse/MRESOLVER-297


jira-importer:
Resolve #976

PR #225, opened 2022-11-29
[EXPERIMENT] Jetty transport

cstamas:
Adds JettyClient backed transport.

The base code is copy-paste from transport-http (native), and tests are disabled for now (same for SSL code).

To build this Java 17 is must.


olamy:
see comment here https://github.com/eclipse/jetty.project/issues/8974#issuecomment-1331529730
 not sure how you find it mandatory to use 17?
I tried this PR by commenting `<maven.compiler.target>17</maven.compiler.target>` and the build passed.
what error do you get exactly?

PR #228, opened 2022-12-09
[MRESOLVER-305] Handle blocked state at connector level

cstamas:
As blocked is really about block remote access. Locally cached things should not be affected, but current solution did not cover all execution paths.

---

https://issues.apache.org/jira/browse/MRESOLVER-305

### DIFF
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultRepositoryConnectorProvider.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultRepositoryConnectorProvider.java
@@ -109,6 +109,20 @@ public class DefaultRepositoryConnectorProvider
         throws NoRepositoryConnectorException
     {
         requireNonNull( repository, "remote repository cannot be null" );
+
+        if ( repository.isBlocked() )
+        {
+            if ( repository.getMirroredRepositories().isEmpty() )
+            {
+                throw new NoRepositoryConnectorException( repository, "Blocked repository: " + repository );
+            }
+            else
+            {
+                throw new NoRepositoryConnectorException( repository, "Blocked mirror for repositories: "
+                        + repository.getMirroredRepositories() );
+            }
+        }
+
         RemoteRepositoryFilter filter = remoteRepositoryFilterManager.getRemoteRepositoryFilter( session );
 
         PrioritizedComponents<RepositoryConnectorFactory> factories = new PrioritizedComponents<>( session );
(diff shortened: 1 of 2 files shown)


maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultRepositoryConnectorProvider.java
```diff
@@ -109,6 +109,20 @@ public RepositoryConnector newRepositoryConnector( RepositorySystemSession sessi
         throws NoRepositoryConnectorException
     {
         requireNonNull( repository, "remote repository cannot be null" );
+
+        if ( repository.isBlocked() )
+        {
+            if ( repository.getMirroredRepositories().isEmpty() )
+            {
+                throw new NoRepositoryConnectorException( repository, "Blocked repository: " + repository );
```
kwin, line 117: we should probably have some tests, because to me it is not clear where this exception is caught and logged. Does it prevent other remote repos from being tried or not?

cstamas: if you look from where this block was _removed_, it is in same try-catch where this method is invoked (and throws this exception). This basically prevents any "remote access" to blocked repositories.

kwin: But if I have 1 blocked remote repo and one not-blocked one I expect the latter to be used instead (if possible). Resolving should only fail if the requested artifact/metadata could not be delivered from any non-blocked remote repo.

cstamas: Yes, as connector provider provides connector _for one given remote repo_ (is param on method), so it will happily provide connector instance for other non-blocked remote repositories....

kwin: Ok, sorry, now I see the related catch block in https://github.com/apache/maven-resolver/blob/752756f123cde40249f3d87490d3c2daa957127e/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultArtifactResolver.java#L584 and indeed that should work just as expected.

cstamas: Yes, in short: resolver will refuse to provide any connector for blocked RemoteRepository instances. Everywhere.

jira-importer:
Resolve #973

PR #229, opened 2022-12-10
Drop FileTransformer API

cstamas:
Just to be clear where it all is: api, spi, impl, connector-basic.


cstamas:
Superseded by https://github.com/apache/maven-resolver/pull/337

PR #230, opened 2022-12-11
MRESOLVER-307 - Support listing of workspace artifacts

laeubi:
Add support for listing artifacts a `WorkspaceReader` know about. 
See https://issues.apache.org/jira/browse/MRESOLVER-307

Currently the build fails because of an incompatibility:
> There is at least one incompatibility:
> org.eclipse.aether.repository.WorkspaceReader.listArtifacts():METHOD_NEW_DEFAULT

As `WorkspaceReader` is a very specialized class and there is a default implementation I won't expect much problems but I'm not sure how to handle this here. Maybe it would be better to even not default implement it to make consumers aware of the new method? Any guidance would be appreciated.

### DIFF
--- a/maven-resolver-api/src/main/java/org/eclipse/aether/repository/WorkspaceReader.java
+++ b/maven-resolver-api/src/main/java/org/eclipse/aether/repository/WorkspaceReader.java
@@ -21,13 +21,18 @@ package org.eclipse.aether.repository;
 
 import java.io.File;
 import java.util.List;
+import java.util.stream.Stream;
 
 import org.eclipse.aether.artifact.Artifact;
 
 /**
  * Manages a repository backed by the IDE workspace, a build session or a similar ad-hoc collection of artifacts.
+ * <em>Note:</em> Implementors are strongly advised to inherit from {@link AbstractWorkspaceReader} instead of directly
+ * implementing this interface.
  * 
  * @see org.eclipse.aether.RepositorySystemSession#getWorkspaceReader()
+ * @noimplement This interface is not intended to be implemented by clients.
+ * @noextend This interface is not intended to be extended by clients.
  */
 public interface WorkspaceReader
 {
@@ -54,5 +59,12 @@ public interface WorkspaceReader
      * @return The available versions of the artifact, must not be {@code null}.
      */
     List<String> findVersions( Artifact artifact );
+
+    /**
+     * List all available artifacts this workspace repository manages.
+     * 
+     * @return a stream of artifacts in no particular order
+     */
+    Stream<Artifact> listArtifacts();
 
 }
--- a/maven-resolver-api/src/main/java/org/eclipse/aether/repository/AbstractWorkspaceReader.java
+++ b/maven-resolver-api/src/main/java/org/eclipse/aether/repository/AbstractWorkspaceReader.java
@@ -1,0 +1,78 @@
+package org.eclipse.aether.repository;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * 
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import java.io.File;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Stream;
+
+import org.eclipse.aether.artifact.Artifact;
+
+/**
+ * A skeleton implementation for custom workspace readers. The default implementation represents an empty repository
+ * that never finds anything.
+ */
+public abstract class AbstractWorkspaceReader
+    implements WorkspaceReader
+{
+
+    protected final WorkspaceRepository repository;
+
+    protected AbstractWorkspaceReader()
+    {
+        this( "empty" );
+    }
+
+    protected AbstractWorkspaceReader( String type )
+    {
+        this( type, null );
+    }
+
+
+    protected AbstractWorkspaceReader( String type, Object key )
+    {
+        repository = new WorkspaceRepository( type, key );
+    }
+
+    @Override
+    public WorkspaceRepository getRepository()
+    {
+        return repository;
+    }
+
+    @Override
+    public File findArtifact( Artifact artifact )
+    {
+        return null;
+    }
+
+    @Override
+    public List<String> findVersions( Artifact artifact )
+    {
+        return Collections.emptyList();
+    }
+
+    @Override
+    public Stream<Artifact> listArtifacts()
+    {
+        return Stream.empty();
+    }
+}
--- a/maven-resolver-impl/src/test/java/org/eclipse/aether/internal/impl/DefaultArtifactResolverTest.java
+++ b/maven-resolver-impl/src/test/java/org/eclipse/aether/internal/impl/DefaultArtifactResolverTest.java
@@ -18,8 +18,13 @@ package org.eclipse.aether.internal.impl;
  * specific language governing permissions and limitations
  * under the License.
  */
-
-import static org.junit.Assert.*;
+import static org.junit.Assert.assertEquals;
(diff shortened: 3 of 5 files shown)


laeubi:
@cstamas @olamy maybe you can take a look here?

cstamas:
This is for sure incompatible change (hence the build failure), so I'd be careful here: Maven implements this class, but also Tycho, m2e etc... so a LOT of breakage here...

Ideally, I'd make this in 2.0 of resolver, and would introduce "safety measures" like for the rest: iface directly should not be implemented, but provide a "support class"?

laeubi:
> implements this class, but also Tycho, m2e

Tycho/m2e is actually the reason I'd like to have such method :-)

> so a LOT of breakage here...

That's the question, do we really "break" anything? As it is a default implemented method, everything will compile as before and no one will really notice even though it is not 100% binary compatible.

> Ideally, I'd make this in 2.0 of resolver, 

Is there actually any plan for 2.0 release anywhere soon or is it more "sometimes in the future"? Should I target another branch? If 2.x is the way to go I'll update this to not use a default method then, beside this do you think this can be done that way or are there any other concerns?

> and would introduce "safety measures" like for the rest: iface directly should not be implemented, but provide a "support class"?

What do you mean by "support class"? And how is it more safe than in interface? I think interfaces are more appropriate for this and are as safe as an (abstract) class?

laeubi:
> allows us to introduce methods that would be otherwise breaking, and provide some default logic.

I'm not really understand the difference to a default implemented interface method?!?

> FORBIDS direct implementation

At laest maven extends AND implements this interface, I think this will still be possible but "discouraged"?

cstamas:
It is a bit japicm as well, see here https://github.com/siom79/japicmp/issues/201

But presence of "indirection" in form of support class allows us:
* to move iface from source to binary only compat change (then build would not fail)
* implement any newly added methods in support class, to retain source compatibility
I bet there are cases like in Maven as well, but somewhere we have to start :smile:

laeubi:
> implement any newly added methods in support class, to retain source compatibility

I still don'T get why this is considdered more "source compatible" than a default interface method because both seem only source compatible?!?

But anyways so I would

1. add `noimplement` / `noextend`
2. add an `AbstractWorkspaceReaderClass` (?)
3. Wait for 2.0 ...

gnodet:
> It is a bit japicm as well, see here [siom79/japicmp#201](https://github.com/siom79/japicmp/issues/201)
> 
> But presence of "indirection" in form of support class allows us:
> 
> * to move iface from source to binary only compat check (then build would not fail)
> * implement any newly added methods in support class, to retain source compatibility
>   I bet there are cases like in Maven as well, but somewhere we have to start 😄

I don't buy the argument that using a class solves the incompatibility problem.  If you add this class to the API, the exact same problem happens with the class.  You still have a possible source incompatibility.  I do think classes with default methods are useless, but for specific reasons (like you usually only want to implement a single method among a lot of methods, such as a listener interface).  For a more common interface, where you usually want to implement all methods, there's no benefit imho.

cstamas:
After reconsidering this change, I am -1 on it (now or 2.x), reasons:
* I see no point of adding new methods to resolver interface that resolver itself is NOT going to use
* obviously, these methods are needed to some "integrator" of resolver, but as can be seen on example of Maven (as one of integrator), one can introduces specific reader (in maven it is MavenWorkspaceReader) and decorate it by adding specific methods to it
* also,by adding methods here  we would force any other integrator to implement them even if they would not need it (as resolver itself is not using these either).

maven-resolver-api/src/main/java/org/eclipse/aether/repository/WorkspaceReader.java
```diff
@@ -55,4 +60,11 @@
      */
     List<String> findVersions( Artifact artifact );
 
+    /**
+     * List all available artifacts this workspace repository manages.
+     * 
+     * @return a stream of artifacts in no particular order
+     */
+    Stream<Artifact> listArtifacts();
```
gnodet, line 68: Given the implementation in https://github.com/apache/maven/pull/913/files#diff-f9e3f0f97d42ff3256cbce238fa8de9cc37a14cfb961d408fc0f4d02f5dd1577R128-R133, I think it should be clearly documented that the returned stream contains the primary artifacts only, i.e. one artifact per maven module.  Attached artifacts are not returned.

laeubi: This is an implementation detail a repository might choose what it claims to be "contained" and I would assume that each of the returned artifacts also return a File if I call `findArtifact` / `findVersion` / `findModel` so this seem consistent for me.

laeubi: Attached Artifacts are also returned now.

maven-resolver-api/src/main/java/org/eclipse/aether/repository/WorkspaceReader.java
```diff
@@ -21,13 +21,18 @@
...
 import org.eclipse.aether.artifact.Artifact;
 
 /**
  * Manages a repository backed by the IDE workspace, a build session or a similar ad-hoc collection of artifacts.
+ * <em>Note:</em> Implementors are strongly advised to inherit from {@link AbstractWorkspaceReader} instead of directly
+ * implementing this interface.
  * 
  * @see org.eclipse.aether.RepositorySystemSession#getWorkspaceReader()
+ * @noimplement This interface is not intended to be implemented by clients.
+ * @noextend This interface is not intended to be extended by clients.
```
gnodet, line 35: I think that's exactly the opposite. The resolver does not provide any concrete implementation, so one has to be provided to the resolver.

maven-resolver-api/src/main/java/org/eclipse/aether/repository/WorkspaceReader.java
gnodet: > Eclipse-plugin never calls that API, 
>resolver does 

Where ? That's exactly my point.  I can't see any usage of this new method.  If really needed, it should be provided by maven, not the resolver it seems.

>and possibly maven-plugins call it (that run in the IDE but are not developed by the IDE maintainers).
> 
> The rationale is to allow for a `maven-plugin` to discover all projects (e.g. in a reactor build or IDE), currently one can only call `MavenSession.getProjects()` / `MavenSession.getAllProjects()` but these are usually populated for different purpose (e.g. `-am` / `-amd`) and require MavenProjects to be setup (so no way to find out "artifacts" only).

Yes, and the projects listed has changed in maven 4.x, and I'm working on another change.  So if this method is to be used for anything else than a hint, it should be more clear about the artifacts returned or how the return stream can be used imho.

> 
> e.g. in Eclipse PDE there is a "TargetPlatform", that contains all available artifacts a project can resolve and projects then can use these to resolve their dependencies, also this is something where it could be used and Tycho currently exposes this with a special component (that obviously only plugins can use that "know" Tycho/PDE).
> 
> So this is more "maven" generic, just think about my build fails with
> 
> > can't find my.atifact in any remote-repository
> 
> I can write a plugin that scans all artifacts using fuzzy-search and gives a message:
> 
> > Did you probabbly mean `my.artifact` instead of `my.atifact`

Sounds good, but if the implementation simply iterates on all projects, not sure why adding it here since this has nothing to do with the resolver really.

maven-resolver-api/src/main/java/org/eclipse/aether/repository/WorkspaceReader.java
laeubi: This was not my idea but @cstamas asked for this :-)

cstamas: Um, where did I do that? If I did, it was a mistake...

laeubi: See https://github.com/apache/maven-resolver/pull/230#issuecomment-1345642240
> Ideally, I'd make this in 2.0 of resolver, and would introduce "safety measures" like for the rest: iface directly should not be implemented, but provide a "support class"?

And here https://github.com/apache/maven-resolver/pull/230#issuecomment-1345989051
You reference a class that uses 
```
 @noimplement This interface is not intended to be implemented by clients.
 @noextend This interface is not intended to be extended by clients.
```

so maybe I misunderstood?

gnodet: I _think_ `client` is to be understood in the context of the resolver, i.e. any piece of software using the resolver.  The main client is maven, and it's the one implementing this interface.  This interface is pure SPI and _should be provided_ by the client.

laeubi:
As @cstamas has explained, this might not be useful for resolver if it never has a need to call the API, so lets close this, probabbly some time later but it won't be hard to recover or re-implement it.

jira-importer:
Resolve #979

PR #232, opened 2022-12-27
[MRESOLVER-302] Introduce onSessionClose hooks

cstamas:
That integrators must integrate, to provide onSessionClosed hooks functionality to any component requiring it.

---

https://issues.apache.org/jira/browse/MRESOLVER-302

### DIFF
--- a/maven-resolver-api/src/main/java/org/eclipse/aether/RepositorySystem.java
+++ b/maven-resolver-api/src/main/java/org/eclipse/aether/RepositorySystem.java
@@ -302,4 +302,39 @@ public interface RepositorySystem
      * @since 1.9.0
      */
     void shutdown();
+
+    /**
+     * Registers the session for lifecycle tracking: it marks that the passed in session instance is about to start.
+     * After this call it is possible to register "on close" handlers using
+     * {@link #addOnSessionEndedHandle(RepositorySystemSession, Runnable)} method that will execute once
+     * {@link #sessionEnded(RepositorySystemSession)} method was invoked.
+     * <p>
+     * <en>Same session instance can be started only once.</em>
+     *
+     * @param session the session that is about to start, never {@code null}.
+     * @since TBD
+     */
+    void sessionStarted( RepositorySystemSession session );
+
+    /**
+     * Registers a handler to execute when this session ends. This method throws, if the passed in session instance
+     * was not passed to method {@link #sessionStarted(RepositorySystemSession)} beforehand.
+     *
+     * @param session the session for which the handler needs  to be registered, never {@code null}.
+     * @param handler the handler, never {@code null}.
+     * @since TBD
+     */
+    void addOnSessionEndedHandle( RepositorySystemSession session, Runnable handler );
+
+    /**
+     * Signals to repository system that passed in session ended, it will not be used anymore. Repository system
+     * will invoke the registered handlers for this session, if any. This method throws if the passed in session
+     * instance was not passed to method {@link #sessionStarted(RepositorySystemSession)} beforehand.
+     * <p>
+     * <en>Same session instance can be ended only once.</em>
+     *
+     * @param session the session that just ended, never {@code null}.
+     * @since TBD
+     */
+    void sessionEnded( RepositorySystemSession session );
 }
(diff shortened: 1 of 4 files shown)


maven-resolver-api/src/main/java/org/eclipse/aether/RepositorySystem.java
```diff
@@ -302,4 +302,39 @@ List<RemoteRepository> newResolutionRepositories( RepositorySystemSession sessio
...
      */
     void shutdown();
+
+    /**
+     * Registers the session for lifecycle tracking: it marks that the passed in session instance is about to start.
+     * After this call it is possible to register "on close" handlers using
+     * {@link #addOnSessionEndedHandle(RepositorySystemSession, Runnable)} method that will execute once
+     * {@link #sessionEnded(RepositorySystemSession)} method was invoked.
+     * <p>
+     * <en>Same session instance can be started only once.</em>
```
kwin, line 312: wrong start tag `en`. Affects multiple sources in this PR.

maven-resolver-api/src/main/java/org/eclipse/aether/RepositorySystem.java
```diff
@@ -302,4 +302,39 @@ List<RemoteRepository> newResolutionRepositories( RepositorySystemSession sessio
...
+     * Signals to repository system that passed in session ended, it will not be used anymore. Repository system
+     * will invoke the registered handlers for this session, if any. This method throws if the passed in session
+     * instance was not passed to method {@link #sessionStarted(RepositorySystemSession)} beforehand.
+     * <p>
+     * <en>Same session instance can be ended only once.</em>
+     *
+     * @param session the session that just ended, never {@code null}.
+     * @since TBD
+     */
+    void sessionEnded( RepositorySystemSession session );
```
kwin, line 339: From a consumer perspective a `close()` method on RSS would be a lot easier and in addition allows to use
- try with resources
- semantic code analysis tools to emit a WARN when someone forgot to call close.

maven-resolver-api/src/main/java/org/eclipse/aether/RepositorySystem.java
```diff
@@ -302,4 +302,39 @@ List<RemoteRepository> newResolutionRepositories( RepositorySystemSession sessio
...
+
+    /**
+     * Registers the session for lifecycle tracking: it marks that the passed in session instance is about to start.
+     * After this call it is possible to register "on close" handlers using
+     * {@link #addOnSessionEndedHandle(RepositorySystemSession, Runnable)} method that will execute once
+     * {@link #sessionEnded(RepositorySystemSession)} method was invoked.
+     * <p>
+     * <en>Same session instance can be started only once.</em>
+     *
+     * @param session the session that is about to start, never {@code null}.
```
kwin, line 314: Why can’t this be transparently handled whenever RSS is created?

maven-resolver-api/src/main/java/org/eclipse/aether/RepositorySystem.java
cstamas: This is the least painful way with 1.x. Problem is how session instances are created (by client code, using def ctor of DefaultRepositorySystemSession, no factory for it) and _derived_. Just look thru Maven and plugin sources as example. They are not "boxed" nor are hierarchical, you can get session A and have it "derived" (using copy ctor) into session B and then continue to use it (typical when local repo is overridden in plugins), etc. And then i did not even mention "forwarding" session...

So yes, ideally session would be:
* created by some factory (repo system, ideally?)
* closeable

But doing this would be very disruptive, so baby steps... leave these changes for 2.0

kwin: Understood, but the same uncertainty is now applicable to `sessionEnded(...)`. When should consumers call it? What about sessions created with https://maven.apache.org/resolver/apidocs/org/eclipse/aether/DefaultRepositorySystemSession.html#%3Cinit%3E(org.eclipse.aether.RepositorySystemSession)? What about forwarding sessions. The Javadocs need to be extended in this regard.

kwin: Given all that I still think that a `close()` on `RepositorySystemSession` is preferable.
The different implementations need to implement that method differently:
- `DefaultRepositorySystemSession.close()` should really do the cleanup in case it has been created from scratch
- `DefaultRepositorySystemSession.close()` should be a noop in case it has been created via copy constructor
- `AbstractForwardingRepositorySystemSession.close()` should be a noop

Doing this should be fully backwards compatible and still expose an API which is much easier to digest.

cstamas:
Superseded by https://github.com/apache/maven-resolver/pull/357

jira-importer:
Resolve #974

PR #233, opened 2023-01-04
[MRESOLVER-309] Do not kill the build when unkown remote layout

cstamas:
In case remote repository defines unsupported layout, filter should just let it pass (as resolver will gracefully fail with it anyway), while throwing RuntimeException caused abruptly stopped build.

---

https://issues.apache.org/jira/browse/MRESOLVER-309


gnodet (approved):
This PR fixes the problem for me.

jira-importer:
Resolve #994

PR #241, opened 2023-02-08
[MRESOLVER-320] Use strong references by default

cstamas:
This increases memory usage (500MB vs 1.1G on tested project), but this is on par with Maven 3.8.7 behaviour and ceases the slowdown as well (speed goes back on par as well).

---

https://issues.apache.org/jira/browse/MRESOLVER-320


maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/DataPool.java
```diff
@@ -80,14 +76,14 @@ public DataPool( RepositorySystemSession session )
 
         if ( cache != null )
         {
-            artifacts = (ObjectPool<Artifact>) cache.get( session, ARTIFACT_POOL );
-            dependencies = (ObjectPool<Dependency>) cache.get( session, DEPENDENCY_POOL );
-            descriptors = (Map<Object, WeakReference<Descriptor>>) cache.get( session, DESCRIPTORS );
+            artifacts = (ConcurrentHashMap<Artifact, Artifact>) cache.get( session, ARTIFACT_POOL );
+            dependencies = (ConcurrentHashMap<Dependency, Dependency>) cache.get( session, DEPENDENCY_POOL );
+            descriptors = (ConcurrentHashMap<Object, Descriptor>) cache.get( session, DESCRIPTORS );
```
gnodet, line 81: Could we add a `computeIfAbsent` method on the `RepositoryCache` to avoid this boilerplate ?
```
Object computeIfAbsent(@Nonnull Object key, @Nonnull Supplier<Object> supplier);
```
so that we can the do
```
artifacts = cache.computeIfAbsent( session, ARTIFACT_POOL, () -> new ConcurrentHashMap<>() );
```

gnodet (changes requested):

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/DataPool.java
cstamas: Well, we could of course, but if you look at class
https://github.com/apache/maven-resolver/blob/master/maven-resolver-api/src/main/java/org/eclipse/aether/RepositoryCache.java

It is not marked as "noextend", hence it would be a breaking change => resolver major version. In fact, I think m2e does implement this class, this was one of the main reasons to introduce it AFAIR. This class is very differernt than SessionData, that is NOT meant to be implemented outside of resolver....

gnodet: And with a default method in the interface ?  `Map` as tons of them...
```
    default Object computeIfAbsent( RepositorySystemSession session, Object key, Supplier<Object> supplier )
    {
        Object value = get( session, key );
        if ( value == null )
        {
            value = supplier.get();
            put( session, key, value );
        }
        return value;
    }
```
This keeps compatibility.  The default implementation should of course leverage the `cache.computeIfAbsent()` method of the underlying `ConcurrentHashMap`...

gnodet:
Also, wouldn't it be preferable to intern the artifacts and dependencies earlier, i.e. at creation time, rather than later ?  Don't most created artifacts / dependencies end up in the end result ?

And subsequent question, have you tried without pooling at all ? To see the difference in perfs / memory consumption ?

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/DataPool.java
cstamas: maybe, but according to japicmp that above is incompatible change.... and we promise this about API https://maven.apache.org/resolver/api-compatibility.html

gnodet: As of Chapter 13. of the [JLS](https://docs.oracle.com/javase/specs/jls/se8/html/jls-13.html#jls-13.5.3):

> Adding a default method, or changing a method from abstract to default, does not break compatibility with pre-existing binaries, but may cause an IncompatibleClassChangeError if a pre-existing binary attempts to invoke the method.

Such usage is unexpected to say the least, so japicmp goes on the safe side, but that should not be a problem really.  See https://github.com/siom79/japicmp/issues/201

gnodet (approved):

cstamas:
Reworked a bit: I simply hated the fact that this PR would potentially make Resolver unable to calculate (some probably bigger graph) of dependencies in case of confined memory, even at cost of 10% speed reduction.

Hence, reworked the PR a bit: added a config, that by default makes resolver behave as with original patch (aggressive heap use, loose the 10% slow down), but, in case the behavior can be returned the 1.8.x resolver way (using weak pools).

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/DataPool.java
cstamas: True, but even if this method present (on cache), cache itself is nullable, so we cannot just drop all this

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/DataPool.java
```diff
@@ -53,6 +52,7 @@
  */
 public final class DataPool
 {
+    private static final String CONFIG_PROP_COLLECTOR_POOL_WEAK = "aether.dependencyCollector.pool.weak";
```
gnodet, line 55: Maybe use a non boolean property `aether.dependencyCollector.pool` so that we can experiment with a `weak`, `hard`, `none` ?

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/DataPool.java
```diff
@@ -415,4 +440,46 @@ public int hashCode()
...
+            return map.computeIfAbsent( key, k -> value );
+        }
+    }
+
+    private static class WeakInternPool<K, V> implements InternPool<K, V>
+    {
+        private final WeakHashMap<K, WeakReference<V>> map = new WeakHashMap<>( 256 );
+
+        @Override
+        public V get( K key )
```
gnodet, line 473: I think it's missing a `synchronized` flag to support concurrent read/write access, as I don't think the `WeakHashMap` does support that.

gnodet (changes requested):

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/DataPool.java
gnodet: That would also allow things like `lru(1024)` ... or whatever.

cstamas: done

cstamas:
All good ideas! now we support "hard" (default), "weak" (the 1.8 way) and "none" (for experiment sake) pools

gnodet (approved):

jira-importer:
Resolve #999

PR #247, opened 2023-02-15
Use shared GitHub actions v3

slawekjaranowski:


slawekjaranowski:
> What adds v3? (did not follow GH Actions changes lately)

- we should have one place so simply update
- v3 was created due to breaking changes in parameters names
- and some of cosmetics changes

PR #249, opened 2023-02-16
[MRESOLVER-314] IAEx thrown when sorting versions with hex values

cstamas:
Long outstanding problem: GenericVersion since long time had issue with sorting some peculiar version strings (well, parsing them and the producing comparable Version instances to be precise). One kind of these peculiar strings are hex strings, that we even encourage should be used on this page https://maven.apache.org/maven-ci-friendly.html Problems happen, when Maven decides to sort these versions, like it happened to user using `versions-maven-plugin:set-property` that involves version sorting during version resolution https://issues.apache.org/jira/browse/MRESOLVER-314 (see stack trace). Again, this is nothing new, seems this was like it since ages, but again, we started promoting "CI friendly" not so long ago, and my guess is people started using those, and this issue will just spread more and more.

Origin of the problem is that GenericVersion instances, when parsed out of hex-like strings, produce instances that violate Comparable contract (comparison is transitive), hence Java throws `Caused by: java.lang.IllegalArgumentException: Comparison method violates its general contract!` just out of the blue. Moreover, this problem affects both set of users: those who produce hex-like versioned artifacts, but also those who consume them.

Timeline of events:
* User reported https://issues.apache.org/jira/browse/MRESOLVER-314 against resolver 1.9.4 (albeit this affects all versions IMHO)
* @cstamas  confirmed this is all true that stands in issue (at his amazement)
* @gnodet  created somewhat "synthetic" but reliably working reproducer UT (using UUIDs as versions)
* banging head into wall
* swapped the nasty break with a continue

Problem was compare padding: when a number and non-number was about to be compared, the comparePadding was invoked, that simply stopped when a number was followed by a non-number. Proper fix is to make iteration all way to the end, otherwise here we end up with compare result 0 that means "equals", which in fact is not.

---

https://issues.apache.org/jira/browse/MRESOLVER-314

### DIFF
--- a/maven-resolver-util/src/test/java/org/eclipse/aether/util/version/GenericVersionTest.java
+++ b/maven-resolver-util/src/test/java/org/eclipse/aether/util/version/GenericVersionTest.java
@@ -20,10 +20,12 @@ package org.eclipse.aether.util.version;
  */
 
 import java.util.Locale;
+import java.util.stream.Stream;
 
-import org.eclipse.aether.util.version.GenericVersion;
 import org.eclipse.aether.version.Version;
 import org.junit.Test;
+
+import static java.util.stream.Collectors.toList;
 
 /**
  */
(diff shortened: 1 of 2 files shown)


cstamas:
Undone first attempt :disappointed: now "only" the new UT fails

cstamas:
Locally built:
* this PR (as resolver 1.9.5-SNAPSHOT)
* maven from https://github.com/apache/maven/pull/989 (as maven 3.9.1-SNAPSHOT)
* ran IT suite => pass OK

gnodet (approved):

maven-resolver-util/src/test/java/org/eclipse/aether/util/version/GenericVersionTest.java
```diff
@@ -349,6 +354,39 @@ public void testMaximumSegment()
 
     /**
      * UT for <a href="https://issues.apache.org/jira/browse/MRESOLVER-314">MRESOLVER-314</a>.
+     *
+     * Generates random UUID string based versions and tries to sort them. While this test is not as reliable
+     * as {@link #testCompareUuidVersionStringStream()}, it covers broader range and in case it fails it records
+     * the failed array, so we can investigate more.
+     */
+    @Test
+    public void testCompareUuidRandom() {
```
gnodet, line 363:
```
@RepeatedTest(3)
```

cstamas: Is that junit4?

gnodet: No :-(

PR #250, opened 2023-02-19
[MRESOLVER-311] Upgrade Parent to 39

slachiewicz:
No code change, only formatting applied. Two key plugins install/deploy upped to latest 3.1.0. Followup commit with git-blame-ignore-revs incoming.

---

https://issues.apache.org/jira/browse/MRESOLVER-311


pom.xml
```diff
@@ -288,16 +288,6 @@
   <build>
     <pluginManagement>
       <plugins>
-        <plugin> <!-- remove once parent POM updated -->
-          <groupId>org.apache.maven.plugins</groupId>
-          <artifactId>maven-deploy-plugin</artifactId>
-          <version>3.0.0</version>
-        </plugin>
```
slawekjaranowski, line 295: parent 39 has a m-deploy-p in version 2.8.2

cstamas (approved):

PR #252, opened 2023-02-22
Be more graceful when updating the index file

laeubi:


laeubi:
@cstamas @michael-o WDYT? Can you probabbly enable a run of the workflows?

PR #260, opened 2023-02-27
[MRESOLVER-329] More robust IO in default tracking file manager

cstamas:
In case of concurrent r/w of tracking file, the if Files.isReablable may return true, but the file might be gone once it arrives to Files.newInputStream. In these cases, just ignore the missing file related exception.

---

https://issues.apache.org/jira/browse/MRESOLVER-329


michael-o (changes requested):
Not for 1.9.6 before we understand the cause, us or Java/OS/FS.

I don't like to shove dust under the carpet.

jira-importer:
Resolve #1011

jira-importer:
Resolve #1011

PR #262, opened 2023-03-01
[MRESOLVER-337] Real cause when artifact not found with repository filtering

slawekjaranowski:
When we use remote repository filtering real cause for artifact not found may be hidden.

https://issues.apache.org/jira/browse/MRESOLVER-337

### DIFF
--- a/maven-resolver-api/src/main/java/org/eclipse/aether/resolution/ArtifactResolutionException.java
+++ b/maven-resolver-api/src/main/java/org/eclipse/aether/resolution/ArtifactResolutionException.java
@@ -23,6 +23,7 @@ import java.util.List;
 
 import org.eclipse.aether.RepositoryException;
 import org.eclipse.aether.transfer.ArtifactNotFoundException;
+import org.eclipse.aether.transfer.RepositoryFilteredException;
 import org.eclipse.aether.transfer.RepositoryOfflineException;
 
 /**
@@ -122,7 +123,7 @@ public class ArtifactResolutionException extends RepositoryException {
                 Throwable notFound = null, offline = null;
                 for (Throwable t : result.getExceptions()) {
                     if (t instanceof ArtifactNotFoundException) {
-                        if (notFound == null) {
+                        if (notFound == null || notFound instanceof RepositoryFilteredException) {
                             notFound = t;
                         }
                         if (offline == null && t.getCause() instanceof RepositoryOfflineException) {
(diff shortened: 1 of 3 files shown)


cstamas:
Can you paste a new log with this PR and mark what was changed, please?

maven-resolver-api/src/main/java/org/eclipse/aether/resolution/ArtifactResolutionException.java
```diff
@@ -58,7 +59,7 @@ public ArtifactResolutionException(List<ArtifactResult> results, String message)
      *
      * @param results The resolution results at the point the exception occurred, may be {@code null}.
      * @param message The detail message, may be {@code null}.
-     * @param cause The exception that caused this one, may be {@code null}.
+     * @param cause   The exception that caused this one, may be {@code null}.
```
cstamas, line 62: ?

maven-resolver-api/src/main/java/org/eclipse/aether/transfer/RepositoryFilteredException.java
```diff
@@ -0,0 +1,38 @@
...
+
+import org.eclipse.aether.artifact.Artifact;
+import org.eclipse.aether.repository.RemoteRepository;
+
+/**
+ * Thrown when an artifact was filtered from a particular repository.
+ */
+public class RepositoryFilteredException extends ArtifactNotFoundException {
+
+    /*** Creates a new exception with the specified artifact, repository and detail message.
```
cstamas, line 29: ?

cstamas:
In general I like this PR and idea to distinguish "not found remotely" vs "filtered out" (not even tried remotely)... so I am +1 for this PR but also I'd like to generally improve overall error reporting of resolver.... but that is NOT for 1.9.6 but more like 1.10.0

maven-resolver-api/src/main/java/org/eclipse/aether/transfer/RepositoryFilteredException.java
```diff
@@ -0,0 +1,39 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.eclipse.aether.transfer;
+
+import org.eclipse.aether.artifact.Artifact;
+import org.eclipse.aether.repository.RemoteRepository;
+
+/**
+ * Thrown when an artifact was filtered from a particular repository.
```
cstamas, line 25: @since 1.9.6

cstamas:
This is fine for 1.9.6, let's merge this (add since to javadoc and go ahead)

cstamas (approved):
make new class since 1.9.6 and merge away!

cstamas (approved):

PR #263, opened 2023-03-06
[MRESOLVER-339] Preemptive auth fix

cstamas:
Code is broken when default HTTP or HTTPS (80 or 443) ports are used, as SharedAuthCache "normalized" host will not match those of pre-set in case of preemptive auth. The UT case (non-default ports) works and is valid, as UT cannot bind ports 80 or 443.

Fix: expose AuthCache and do things "by the book".

---

https://issues.apache.org/jira/browse/MRESOLVER-339


slawekjaranowski (approved):
Now is ok, all request have auth

jira-importer:
Resolve #1010

PR #264, opened 2023-03-07
[MRESOLVER-340] Make WebDAV "dance" disabled by default

cstamas:
Vast majority of our users use MRMs and not some "repo server made out of sticks" (like WebDAV server was the fashion in 2000s), but we still "tax" majority of our users with "WebDAV dance".

Make it disabled by default, while those who really need it, still can it enable, even per server.

---

https://issues.apache.org/jira/browse/MRESOLVER-340


jira-importer:
Resolve #1016

jira-importer:
Resolve #1016

PR #265, opened 2023-03-07
[MRESOLVER-341] Preemptive PUT auth

cstamas:
As resolver faces many broken servers that
does expect broken behaviour of Wagon, that blindly does preemptive auth for PUT. Those users who
are not able to fix their servers, can use
this "mild fallback" to make transport-http
behave as stupidly as well.

By default this is disabled.

---

https://issues.apache.org/jira/browse/MRESOLVER-341


jira-importer:
Resolve #1018

PR #268, opened 2023-03-09
remove unused deps

elharo:


cstamas:
These are needed deps, but not coded directly or linked against, but indirectly (guice provides the DI, logging is needed as otherwise you have SLF4J warning, etc).

elharo (commented):
I've restored the Guice and SLF4J deps. Let's see if that works. Looks like the Guice issue is that Plexus isn't correctly declaring its dependencies. SLF4J has been annoyingly broken about this for years.

PR #271, opened 2023-03-22
[MRESOLVER-345] Conflict resolution in verbose mode is sensitive to version ordering

cstamas:
Our two collector implementations produce slightly different dirty trees when in transitive dependencies ranges are in use. The new BF produces (w/ and w/o skipper) produces this dirty tree:
```
menforcer426:aid:ext:1 compile
+- menforcer426:bid:ext:1 compile
|  +- menforcer426:cid:ext:3 compile
|  +- menforcer426:cid:ext:2 compile
|  \- menforcer426:cid:ext:1 compile
+- menforcer426:cid:ext:3 compile
+- menforcer426:cid:ext:2 compile
\- menforcer426:cid:ext:1 compile
```
The "old" DF produces this one:
```
menforcer426:aid:ext:1 compile
+- menforcer426:bid:ext:1 compile
|  +- menforcer426:cid:ext:1 compile
|  +- menforcer426:cid:ext:2 compile
|  \- menforcer426:cid:ext:3 compile
+- menforcer426:cid:ext:1 compile
+- menforcer426:cid:ext:2 compile
\- menforcer426:cid:ext:3 compile
```

Spot the difference: the two dirty trees are "semantically" (or content-wise?) equal/same, except for the artifact ordering, where there was a version range (collector in this case discovers available versions that "fits" range and created nodes for them, one for each version that lies within version constraint). DF collector relies and provides "metadata based" ordering (as metadata contained the versions), while BF explicitly sorts them in descending order (for internal optimization purposes). Point is, both dirty trees are ok.

But, Conflict resolver in verbose mode for two inputs above produces different outputs. For DF with "divergence found",  while for BF "no divergence found" (correctly).

Overall changes in this PR:
* most are test and test related resources
* cosmetic changes like javadoc typos, adding override etc
* added reusable `DependencyGraphDumper`
* key changes are in `ConflictResolver` covered by `ConflictResolverTest` UT changes
* overall fix is to make `ConflictResolver` insensitive to input dirty tree version ordering, make sure output is same for "semantically" (or content-wise?) same inputs.

How tested this:
* built/installed this PR
* built maven-3.9.x with this resolver
* ran maven IT suite -- OK
* ran https://github.com/apache/maven-enforcer/pull/259 w/ built maven (so fixed resolver is used). This PR contains only one reproducer IT that fails with any released Maven version -- OK w/ maven built as above.

Just realized how enforcer ITs are good source of inspiration for resolver use cases, so many if not all of new tests are actually inspired by enforcer ITs.

---

https://issues.apache.org/jira/browse/MRESOLVER-345


cstamas:
Reason is simple: new BF does this:
https://github.com/apache/maven-resolver/blob/master/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/bf/BfDependencyCollector.java#L409-L410

In short: new BF collects versions in reverse order (reverse of the found in metadata, DESC), while DF collects versions by the order in metadata (in both case we assume versions are in ascending order, example https://repo.maven.apache.org/maven2/org/apache/maven/maven-core/maven-metadata.xml)

cstamas:
Created a "demo" snippet, and it produces this output:
```
org.apache.maven.resolver.demo.mresolver345:a:jar:1.0
+- org.apache.maven.resolver.demo.mresolver345:b:jar:1.0 [compile]
|  \- org.apache.maven.resolver.demo.mresolver345:c:jar:3.0 [compile]
\- org.apache.maven.resolver.demo.mresolver345:c:jar:3.0 [compile]
```

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/visitor/DependencyGraphDumper.java
```diff
@@ -31,25 +31,27 @@
...
+
 /**
- * A dependency visitor that dumps the graph to the console.
+ * A dependency visitor that dumps the graph to any {@link Consumer<String>}. Meant for diagnostic and testing, as
+ * it may output the graph to standard output, error or even some logging interface.
+ *
+ * @since 1.9.8
  */
-public class ConsoleDependencyGraphDumper implements DependencyVisitor {
+public class DependencyGraphDumper implements DependencyVisitor {
```
slawekjaranowski, line 42: 👍  - we have similar implementation in many places

slawekjaranowski (approved):

jira-importer:
Resolve #1021

PR #273, opened 2023-03-27
[MRESOLVER-348] SslConfig new httpSecurityMode change is not detected

cstamas:
The new value was left out from equals/hashCode. This is a **no issue at runtime** (as session does not change), but I spotted this during adding UTs and is not correct nevertheless.

---

https://issues.apache.org/jira/browse/MRESOLVER-348


slawekjaranowski (approved):

jira-importer:
Resolve #1022

jira-importer:
Resolve #1022

PR #277, opened 2023-04-05
[MRESOLVER-350] Inline StringUtils

gnodet:
Fwiwm `StringUtils` was deprecated since quite some time, but I'd agree that a micro release is not the best place to remove a public class.  Maybe wait for next minor...

### DIFF
--- a/maven-resolver-util/src/main/java/org/eclipse/aether/util/StringUtils.java
+++ b/maven-resolver-util/src/main/java/org/eclipse/aether/util/StringUtils.java
@@ -21,6 +21,7 @@ package org.eclipse.aether.util;
 /**
  * A utility class to ease string processing.
  */
+@Deprecated
 public final class StringUtils {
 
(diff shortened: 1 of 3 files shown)


maven-resolver-util/src/main/java/org/eclipse/aether/util/StringUtils.java
```diff
@@ -32,7 +33,9 @@ private StringUtils() {
      *
      * @param string The string to check, may be {@code null}.
      * @return {@code true} if the string is {@code null} or of zero length, {@code false} otherwise.
+     * @deprecated
```
cstamas, line 36: Maybe "Deprecated since 1.5.0 without any alternative provided. Use Java 8 APIs instead."?

jira-importer:
Resolve #1025

jira-importer:
Resolve #1025

PR #280, opened 2023-04-21
[MRESOLVER-352] - Duplicate METADATA_DOWNLOADING event is being sent

c00ler:
Ensure that METADATA_DOWNLOADING event is sent only once in case the lock "upgrade" happens


cstamas (approved):

cstamas:
Note: same thing for (similar) DefaultArtifactResolver is NOT happening, as there actual "work" is factored out into performDownloads method happening AFTER the locks upgrade happened (so event is sent only once, event pairs are ok there)

michael-o (approved):
Nice catch!

PR #284, opened 2023-04-28
[MRESOLVER-358] Upgrade some dependencies

cstamas:
Upgrades
* guava to 31.1-jre (was 30.1-jre)
* mockito-core to 4.11.0 (was 4.8.1)
* plexus-utils to 3.5.1 (was 3.5.0)
* maven deps (used in demos only) to 3.8.7 (was 3.8.4)

---

https://issues.apache.org/jira/browse/MRESOLVER-358


maven-resolver-demos/pom.xml
```diff
@@ -38,7 +38,7 @@
   </modules>
 
   <properties>
-    <mavenVersion>3.8.4</mavenVersion>
+    <mavenVersion>3.8.8</mavenVersion>
```
slawekjaranowski, line 41: There is IT in resolver-demo-maven-plugin which will fail when Maven < 3.8.8 will be used ...
But we enforce minimal version as 3.8.4

slawekjaranowski (changes requested):

maven-resolver-demos/pom.xml
cstamas: Unsure I get it: ITs ar built by CI and also by me locally. Which IT will fail, as there is none failing for me?

cstamas: Locally I have no IT failure either, so what is the problem exactly?

slawekjaranowski (approved):

jira-importer:
Resolve #1034

jira-importer:
Resolve #1034

PR #288, opened 2023-05-23
[MRESOLVER-361] Unreliable TCP and retries

cstamas:
Make httpClient retry for PUT.

---

https://issues.apache.org/jira/browse/MRESOLVER-361


Eh2406:
Should this retry handler respect `aether.connector.http.retryHandler.count`?

cstamas:
> Should this retry handler respect `aether.connector.http.retryHandler.count`?

This pr changes ONLY the fact that now client follows rfc2616 and treats listed methods (put among them) as idempotent.

cstamas:
Latest commit exposes all aspects of retryHandler to users via config (defaults to "standard" handler as before in this PR)

Eh2406:
Under what circumstances would a user want the standard retry handler?

cstamas:
No idea, but presence of these options allows to make resolver work "as before". Just an escape hatch, nothing more.

PR #289, opened 2023-05-26
[MRESOLVER-356] Remove Guava

cstamas:
Guava is not used anywhere in codebase, but as transitive dependency of Guice was "controled", but this makes no sense.

---

https://issues.apache.org/jira/browse/MRESOLVER-356


jira-importer:
Resolve #1031

jira-importer:
Resolve #1031

PR #291, opened 2023-05-26
[MRESOLVER-363] Maven metadata not updated in some cases

cstamas:
In case of metadata, when XML file exists, but the resolver-status.properties file does not, or the "lastUpdated" key is not found, the XML is never updated.

NOTE: UT fails, we need more discussion about this, see JIRA issue.

---

https://issues.apache.org/jira/browse/MRESOLVER-363


cstamas:
This is completely wrong, and resolver DOES behaves correctly, as update policy is "never"

PR #292, opened 2023-05-26
Make code self explanatory

cstamas:
This PR does NOT make any semantic change, it literally just makes existing code self-explanatory.

Just to avoid that in 2 years another of us looks at this and become puzzled what in case of "last modified" means 0L, and why a method returns 1L, etc.

This class is fully covered with self explanatory UTs as well.

No JIRA created either.


gnodet (approved):
Looks good, though the value `1` is still a magic value because it's never read and it works only because it's always less than any clock timestamp...

michael-o (approved):

cstamas:
Again, I did not change anything. But spent so much time figure out the logic behind all this and "why 1???", so intent is really just to add constants that are "decorated" with javadoc explaining their meaning.

PR #295, opened 2023-06-01
[MRESOLVER-364] Return file locking in tracking file manager

cstamas:
This (logically) reverts the fcb6be59c5a5855573886b09c70dab80074a1d27

---

https://issues.apache.org/jira/browse/MRESOLVER-364


michael-o (changes requested):
This class' Javadoc should contain an description why locking is wrong here and why it is still present.

cstamas:
Added Javadoc with pointers for future generations

gnodet (approved):

jira-importer:
Resolve #1030

PR #297, opened 2023-06-03
[MRESOLVER-369] Introduce update check policy scope

cstamas:
This change introduces new configuration: update check policy scope, that limits where (artifact, metadata, both) the update policy is applied. If update policy is not applied, presence/absence decides instead.

To achieve "old" behaviour (ie. for use in Maven3), the configuration properties should always be set to "all" in session factory.

---

https://issues.apache.org/jira/browse/MRESOLVER-369


maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultUpdateCheckManager.java
```diff
@@ -185,7 +224,8 @@ public void checkArtifact(RepositorySystemSession session, UpdateCheck<Artifact,
             if (error != null) {
                 check.setException(newException(error, artifact, repository));
             }
-        } else if (isUpdatedRequired(session, lastUpdated, check.getPolicy())) {
+        } else if (updatePolicyScope.isApplyToArtifact()
+                && isUpdatedRequired(session, lastUpdated, check.getPolicy())) {
             check.setRequired(true);
         } else if (fileExists) {
             LOGGER.debug("Skipped remote request for {}, locally cached artifact up-to-date", check.getItem());
```
cstamas, line 231: Here the message may need adjustment: "locally cache artifact is up-to-date" (if policyScope applies) ot "local cache artifact is present" (if policyScope does not applies and file is present)

gnodet (approved):

cstamas:
PR is on-hold: testing shows that now when Maven "remembers" that Artifact cannot be downloaded, there is no way to get out of it (-U does not help, as policy is not applied to artifact by default anymore, before it simply went here as -U was "always", and implicitly cleared the error flags as well, so it "fixed" things). 

IMO, this stems from the fact that policy ALWAYS was simply misused on Maven side (-U), and instead, "force update" on maven side may mean something else, like:
* set update policy to ALWAYS
* set cache not found/transfer to false
* ... (what else?)

cstamas:
MRESOLVER-377 is proper fix

PR #299, opened 2023-06-12
[MRESOLVER-370] Lock factory diagnostic on failure

cstamas:
The change "pulls" all `NamedLock` interface methods "level up" from implementations to `NamedLockSupport`, to be able to augment instances at `NamedLockSupport` level.

If diagnostic enabled:
* factory makes all named lock instances to gather "diag stats" (in memory!)
* on adapter failure, factory will dump out diagnostic state,
* each active lock (non closed) will dump it's state out (per thread with "steps")
* output goes to lock factory INFO logger, potentially swarming console

By default, when "diagnostic" not turned on, no change really.

---

https://issues.apache.org/jira/browse/MRESOLVER-370

### DIFF
--- a/maven-resolver-named-locks/src/main/java/org/eclipse/aether/named/NamedLockFactory.java
+++ b/maven-resolver-named-locks/src/main/java/org/eclipse/aether/named/NamedLockFactory.java
@@ -42,13 +42,21 @@ public interface NamedLockFactory {
      * Utility method to provide more (factory specific) description when a locking operation failed. Assumption is
      * that provided list has at least one element (one failure) or more (in case of retries), must not be {@code null}
      * to get meaningful exceptions. The returned exception will be new instance of {@link IllegalStateException} with
-     * passed in list added as suppressed exceptions. Still, the fact this method has be invoked, means there is a
+     * passed in list added as suppressed exceptions. Still, the fact this method has been invoked, means there is a
      * "abort failure" ahead, so factory may either decorate (add info about state) or even log some diagnostic
      * about the state of the locks.
+     * <p>
+     * The default implementation merely does what happened before this change: adds no extra information.
      *
      * @since TBD
      * @return A new instance of {@link IllegalStateException} (decorated) and may have other side effects as well
-     * (dumping state), never {@code null}.
+     * (dumping state), never returns {@code null}.
      */
-    IllegalStateException failure(List<IllegalStateException> attempts);
+    default IllegalStateException failure(List<IllegalStateException> attempts) {
+        IllegalStateException ex = new IllegalStateException("Could not acquire lock(s)");
+        if (attempts != null) {
+            attempts.forEach(ex::addSuppressed);
+        }
+        return ex;
+    }
 }
--- a/maven-resolver-named-locks/src/main/java/org/eclipse/aether/named/NamedLockFactory.java
+++ b/maven-resolver-named-locks/src/main/java/org/eclipse/aether/named/NamedLockFactory.java
@@ -18,8 +18,6 @@
  */
 package org.eclipse.aether.named;
 
-import java.util.List;
-
 /**
  * A factory of {@link NamedLock}s.
  */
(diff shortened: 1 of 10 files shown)


maven-resolver-named-locks/src/main/java/org/eclipse/aether/named/NamedLockFactory.java
```diff
@@ -35,4 +37,18 @@ public interface NamedLockFactory {
...
+     * to get meaningful exceptions. The returned exception will be new instance of {@link IllegalStateException} with
+     * passed in list added as suppressed exceptions. Still, the fact this method has be invoked, means there is a
+     * "abort failure" ahead, so factory may either decorate (add info about state) or even log some diagnostic
+     * about the state of the locks.
+     *
+     * @since TBD
+     * @return A new instance of {@link IllegalStateException} (decorated) and may have other side effects as well
+     * (dumping state), never {@code null}.
+     */
+    IllegalStateException failure(List<IllegalStateException> attempts);
```
gnodet, line 53: If `NamedLockFactory` is a public SPI, maybe a `default` method would be more appropriate here ?

maven-resolver-named-locks/src/main/java/org/eclipse/aether/named/support/NamedLockSupport.java
```diff
@@ -32,18 +38,76 @@ public abstract class NamedLockSupport implements NamedLock {
...
     @Override
     public String name() {
         return name;
     }
 
+    @Override
+    public boolean lockShared(long time, TimeUnit unit) throws InterruptedException {
+        if (state != null) {
+            state.computeIfAbsent(Thread.currentThread(), k -> new ArrayDeque<>())
+                    .push("S");
```
gnodet, line 58: if we don't expect more than a few locks per thread at the same time, I would use `shared` and `exclusive`, it will not require the reader to go back to the source to understand what S/X means.

cstamas: Agreed, although named locks API does not limit (boxing depth), in reality syncContext does "shallow" boxing.

bmarwell (approved):

maven-resolver-named-locks/src/main/java/org/eclipse/aether/named/support/NamedLockFactorySupport.java
```diff
@@ -34,6 +35,8 @@
 public abstract class NamedLockFactorySupport implements NamedLockFactory {
     protected final Logger logger = LoggerFactory.getLogger(getClass());
 
+    final boolean diagnostic = logger.isDebugEnabled();
```
gnodet, line 38: That's a bad idea.  Loggers level can be updated at runtime, and the call to `isDebugEnabled()` should be quite enough to not warrant caching the value.

maven-resolver-named-locks/src/main/java/org/eclipse/aether/named/support/NamedLockSupport.java
```diff
@@ -43,7 +43,7 @@ public abstract class NamedLockSupport implements NamedLock {
     public NamedLockSupport(final String name, final NamedLockFactorySupport factory) {
         this.name = name;
         this.factory = factory;
-        this.state = factory.diagnostic ? new ConcurrentHashMap<>() : null;
+        this.state = factory.isDiagnosticEnabled() ? new ConcurrentHashMap<>() : null;
```
gnodet, line 46: If the logger level changes at runtime, that may lead to an NPE, so I'd just create the map in all cases.

gnodet (changes requested):

maven-resolver-named-locks/src/main/java/org/eclipse/aether/named/support/NamedLockSupport.java
gnodet: You're right, it won't lead to a NPE, I misread the code.  However, it's caching the value of `isDiagnosticEnabled()`, so if we assume that its value can change at runtime, we should create the map, then check again with `factory.isDiagnosticEnabled()` in the code instead of `state != null`.

cstamas: Am actually quite uneasy with this "logger level changes at runtime".... As named lock instances come and go, but this is possible as well:
```
DI container 
  |-------------------------------------------------------------|
NamedLockFactory (as is singleton)
  |-------------------------------------------------------------|
Session1
  |-------------------------------|
Session2
                | ---------------------|
Session3
                                     | ---------------------|
NamedLock1
  |---------------------------------------------------|
NamedLock2
                                     |-------------|
```

As instances are cached (and ref counted), potentially they can span multiple sessions even, without any connection to session. IF we would always obey logger level, we could end up with "incomplete" diag data as well, as logger debug may be disabled-enabled-disabled-enabled, but if lock instance spans across all 4 changes, diag map would contain only part of data....

cstamas: Due this above (and absence of session, plus the life-span of factory and lock instance across many sessions), I'd change "diag" switch just to be a Java system property... @gnodet WDYT?

maven-resolver-named-locks/src/main/java/org/eclipse/aether/named/support/NamedLockSupport.java
```diff
@@ -32,18 +39,74 @@ public abstract class NamedLockSupport implements NamedLock {
 
     private final NamedLockFactorySupport factory;
 
+    private final ConcurrentHashMap<Thread, Deque<String>> diagnosticState; // non-null only if diag enabled
```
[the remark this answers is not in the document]
cstamas: How would we list then ALL threads having locks?

gnodet (approved):

jira-importer:
Resolve #1047

PR #300, opened 2023-06-16
[MRESOLVER-371] Fix getMutex method noise

cstamas:
When file does not exists. Also, make the method more robust.

---

https://issues.apache.org/jira/browse/MRESOLVER-371


gnodet (approved):

jira-importer:
Resolve #1045

jira-importer:
Resolve #1045

PR #301, opened 2023-06-19
[MRESOLVER-220] Refuse proactively unsupported operation

cstamas:
Instead to silently "send result" (that is false) to caller, as this makes totally strange outcome: "failed to lock within 30sec" but caller did not wait anything. Moreover, the bug where upgrade is attempted slips in unnoticed.

---

https://issues.apache.org/jira/browse/MRESOLVER-220

### DIFF
--- a/maven-resolver-impl/src/test/java/org/eclipse/aether/internal/impl/synccontext/NamedLockFactoryAdapterTestSupport.java
+++ b/maven-resolver-impl/src/test/java/org/eclipse/aether/internal/impl/synccontext/NamedLockFactoryAdapterTestSupport.java
@@ -31,6 +31,7 @@ import org.eclipse.aether.SyncContext;
 import org.eclipse.aether.artifact.DefaultArtifact;
 import org.eclipse.aether.internal.impl.synccontext.named.*;
 import org.eclipse.aether.named.NamedLockFactory;
+import org.eclipse.aether.named.support.LockUpgradeNotSupportedException;
 import org.eclipse.aether.repository.LocalRepository;
 import org.eclipse.aether.spi.synccontext.SyncContextFactory;
 import org.junit.AfterClass;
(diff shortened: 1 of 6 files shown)


maven-resolver-impl/src/test/java/org/eclipse/aether/internal/impl/synccontext/NamedLockFactoryAdapterTestSupport.java
```diff
@@ -266,7 +267,7 @@ public void run() {
                         chained.run();
                     }
                     loser.await();
-                } catch (IllegalStateException e) {
+                } catch (IllegalStateException | LockUpgradeNotSupportedException e) {
                     e.printStackTrace(); // for ref purposes
```
gnodet, line 271: I know this is unrelated, but we could use a logger instead of stderr or just remove this statement or make it conditional ?

maven-resolver-named-locks-hazelcast/src/test/java/org/eclipse/aether/named/hazelcast/NamedLockFactoryAdapterTestSupport.java
```diff
@@ -230,7 +231,7 @@ public void run() {
                         chained.run();
                     }
                     loser.await();
-                } catch (IllegalStateException e) {
+                } catch (IllegalStateException | LockUpgradeNotSupportedException e) {
                     loser.countDown();
```
gnodet, line 235: The previous statement does print something, that one should be coherent imho.

gnodet (approved):

maven-resolver-impl/src/test/java/org/eclipse/aether/internal/impl/synccontext/NamedLockFactoryAdapterTestSupport.java
cstamas: This is test support class, not used in prod

maven-resolver-named-locks-hazelcast/src/test/java/org/eclipse/aether/named/hazelcast/NamedLockFactoryAdapterTestSupport.java
cstamas: This is test support class, not used in prod (a bit mutated but they originate from same code)

jira-importer:
Resolve #946

PR #302, opened 2023-06-19
[MRESOLVER-373] Partially undo MRESOLVER-346

cstamas:
In change 4c5e9ea98f8815c6df8bf26baa9032c8d9cd5f2d we introduced code that in some case tries illegal "lock upgrade", that due lack of MRESOLVER-220 went unnoticed.

This change merely undoes the changes for DefaultArtifactResolver and DefaultMetadataResolver (will as before, use exclusive SyncContext immediately, not be "optimistic"), while change for DefaultDeploy is still correct.

---

https://issues.apache.org/jira/browse/MRESOLVER-373


gnodet (approved):

jira-importer:
Resolve #1046

jira-importer:
Resolve #1046

PR #303, opened 2023-06-20
[MRESOLVER-373] Make GavNameResolver to distinguish names better

cstamas:
This class currently gives away same names for artifact and metadata locks, that causes MRESOLVER-373 where artifact and metadata resolver together attempt illegal "lock upgrade", as both operate on same named lock.

This is wrong, as all other name mappers distinguish among them, also in case of snapshots, there are cases when shared lock is enough for artifact but metadata MAY need refresh, hence exclusive.

Important note: changing "naming" implies, that Maven carrying resolver with this change will be UNABLE to properly "share" local repository with older Mavens (so if this gets into Maven 3.9.3, it will properly share local repository other Maven 3.9.3+ instances, but not with 3.9.2, 3.9.1 or 3.9.0!)

---

https://issues.apache.org/jira/browse/MRESOLVER-373


gnodet (approved):

jira-importer:
Resolve #1046

jira-importer:
Resolve #1046

PR #305, opened 2023-06-24
refactor: refactor bad smell ProtectedMemberInFinalClass

MartinWitt:
# Repairing Code Style Issues

## ProtectedMemberInFinalClass
Since final classes cannot be inherited, marking a member as protected may be confusing. It is better to declare such members as private or package-visible instead.

# Repairing Code Style Issues
* ProtectedMemberInFinalClass (2)


cstamas:
Maybe. But opening up their visibility?

MartinWitt:
You mean converting them to `public`? The `protected` modifier currently does nothing, as these classes are final.

PR #307, opened 2023-06-29
[MRESOLVER-375] Provided and trusted checksum fixes

cstamas:
Several key aspects are broken in provided and trusted checksum feature of resolver:
* trusted checksum post processor: the checksum algorithm is not prefixed (as documented)
* the trusted to provided adapter is defunct without MRM being configured
* the existing ProvidedChecksumsSource interface is broken, needs to be replaced
* some unused loggers and misplaced logging (ie. log may be misleading, move it to actual file reading)

---

https://issues.apache.org/jira/browse/MRESOLVER-375


jira-importer:
Resolve #1052

PR #310, opened 2023-06-29
[MRESOLVER-378] Update parent POM to 40

cstamas:
---

https://issues.apache.org/jira/browse/MRESOLVER-378


cstamas:
There are plugin entries to be removed, and probably japicmp update

slawekjaranowski:
We can remove version for install/deploy plugin

jira-importer:
Resolve #1059

PR #311, opened 2023-06-30
[MRESOLVER-376] SOError fix for BF collector

cstamas:
Solve the mixup of Artifact used as key, as in case of relocation A1 -> A2, when A1 descriptor is read, the result will have A2 artifact set.

So, in code, we must go for "originally asked" A1 artifact, as in case of relocation, the descriptor will hold A1 relocation target, that is A2, causing endless loop.

---

https://issues.apache.org/jira/browse/MRESOLVER-376


cstamas:
Btw: here is a simple reproducer https://gist.github.com/cstamas/cce5cfd94f6fa24a40c9239ca7040230

jira-importer:
Resolve #1057

PR #313, opened 2023-06-30
[MRESOLVER-380] Failed locking attempts are not removed

cstamas:
And logged thread diag does not distinguish: "waits for X" or "has X".

---

https://issues.apache.org/jira/browse/MRESOLVER-380


jira-importer:
Resolve #1056

jira-importer:
Resolve #1056

PR #314, opened 2023-06-30
[MRESOLVER-381] Make "happy path" shared locking again

cstamas:
Partially undo MRESOLVER-346: 1acec9baf6fcb8eb010d64e8009778a13a88d38e
Also, make existing "static" name mapper obey the rule (distinguish between artifacts and metadata) and make not about this on NameMapper interface as we..

---

https://issues.apache.org/jira/browse/MRESOLVER-381


jira-importer:
Resolve #1062

jira-importer:
Resolve #1062

PR #317, opened 2023-07-19
[MRESOLVER-385] Reduce aether.connector.http.connectionMaxTtl to 300

kwin:
This is the same default value as in Wagon.
Also clarify that this time to life is the not related to the idle time.


cstamas (approved):

jira-importer:
Resolve #1063

jira-importer:
Resolve #1063

PR #322, opened 2023-07-27
[MRESOLVER-390] Customize graph visiting strategy

cstamas:
Resolver (and thus Maven) had "wired in" strategy to flatten dependency graph into list, and it was always "preOrder", while there existed alternate solutions like "postOrder". This PR introduces "levelOrder" as asked by users, and also makes it configurable (via session config) which strategy should be used, and exposes helper classes and methods.

This PR introduces new visitors that are able to perform the 3 visiting strategies and uses them, while the old ones are left in place (for binary compatibility) but are NOT used in Resolver and their use is discouraged.

Now session can be configured to use following strategies to flatten the graph:
* "preOrder" -- as before
* "postOrder" -- was already present, but unused (unless some code used it directly)
* "levelOrder" -- as asked by users, artifact list is ordered by their depth (distance from root)

By default, this PR introduces NO behavior changes, but opens configuration that does allow to alter these.

---

https://issues.apache.org/jira/browse/MRESOLVER-390


jira-importer:
Resolve #1074

PR #332, opened 2023-09-21
[MRESOLVER-406] Update Jetty to 9.4.52.v20230823

cstamas:
The older versions trigger nasty CVE reports. No real effect, as this is really just a test dependency.

---

https://issues.apache.org/jira/browse/MRESOLVER-406


michael-o (approved):

jira-importer:
Resolve #1086

PR #337, opened 2023-10-14
[MRESOLVER-338] Drop FileTransform API

cstamas:
To streamline logic all way down to transport and remove OOM prone code.

Note: japicmp config was added, as we did a mistake, we deprecated types, but not related methods, so without these japicm would fail the build due breakages.

---

https://issues.apache.org/jira/browse/MRESOLVER-338


jira-importer:
Resolve #1009

PR #340, opened 2023-10-14
[MRESOLVER-157] Drop Service Locator

cstamas:
Changes:
* Drop deprecated ServiceLocator.
* Switch to Maven4, required (using 4.0.0-alpha-7), as it dropped SL use.
* Drop Guice modules (were deprecated in 1.9.x) and demos using it.

---

https://issues.apache.org/jira/browse/MRESOLVER-157
https://issues.apache.org/jira/browse/MRESOLVER-410
https://issues.apache.org/jira/browse/MRESOLVER-411


jira-importer:
Resolve #1089

PR #341, opened 2023-10-15
[MRESOLVER-413] BF collector should not use maven.artifact.threads

cstamas:
The property is for artifact downloading/uploading, and was introduced way before BF that itself contains a thread pool. Using value 1 for maven.artifact.threads caused build failure.

Update doco accordingly.

---

https://issues.apache.org/jira/browse/MRESOLVER-413


jira-importer:
Resolve #1090

jira-importer:
Resolve #1090

PR #342, opened 2023-10-15
[MRESOLVER-414] Fully split update policies internally

cstamas:
And externally provide means to not use this feature (to use them as one), basically to fall back to 1.x behavior.

The code was half-baked, so now resolver internally fully splits and maintaines two policies, one for artifacts and one for metadata. Still, on surface ther are getUpdatePolicy() and setUpdatePolicy() that makes resolver behave as before (as in 1.x).

---

https://issues.apache.org/jira/browse/MRESOLVER-414


jira-importer:
Resolve #1041

jira-importer:
Resolve #1041

PR #346, opened 2023-10-20
[MRESOLVER-409] More deprecation removal

cstamas:
Now the last bit remaining is ChecksumUtils that is to be handled in different issue. This PR merely removes all the compiler warnings about deprecation but ChecksumUtils.

---

https://issues.apache.org/jira/browse/MRESOLVER-409


jira-importer:
Resolve #1083

PR #347, opened 2023-10-20
[MRESOLVER-419] Preserve component names

cstamas:
Do not inject/convert, inject plain Map instead.

---

https://issues.apache.org/jira/browse/MRESOLVER-419


jira-importer:
Resolve #1096

jira-importer:
Resolve #1096

PR #348, opened 2023-10-20
[MRESOLVER-420] Cache prioritized components

cstamas:
Cache them, as creating of those is not cheap, but detect (dynamic injected) map changes. Allow users to disable this feature, in case of fire.

---

https://issues.apache.org/jira/browse/MRESOLVER-420


jira-importer:
Resolve #1095

PR #351, opened 2023-10-27
[MRESOLVER-392] Remove change detection logic on install

cstamas:
Still, make possible to restore the "skip if missing" as I find no reasonong for it, but my guess is that it may be about some legacy code somewhere, most probably adding Metadata without backing file (as in Maven2 metadata was added by hand).

---

https://issues.apache.org/jira/browse/MRESOLVER-392


jira-importer:
Resolve #1065

PR #352, opened 2023-10-27
[MRESOLVER-295] ChecksumUtils should be gone (deprecated)

cstamas:
Shuffle it's pieces along to their final place, and deprecate the whole class, but leave it in place, as Resolver 1.x did not deprecate it. To be removed in 2.1 of Resolver.

---

https://issues.apache.org/jira/browse/MRESOLVER-295


jira-importer:
Resolve #970

jira-importer:
Resolve #970

PR #355, opened 2023-11-07
[MRESOLVER-424] Make graph dumper extensible

cstamas:
Make graph dumper extensible, as it is meant to be reusable, but not everyone may want same info on output. Also, most of the methods were private...

---

https://issues.apache.org/jira/browse/MRESOLVER-424


jira-importer:
Resolve #1093

jira-importer:
Resolve #1093

PR #356, opened 2023-11-07
[MRESOLVER-426] Make some HTTP configuration shared

cstamas:
Many of HTTP related options are not Apache HTTP transport specific. Also, make possible some newer java.net.http.HttpClient options to be applied.

---

https://issues.apache.org/jira/browse/MRESOLVER-426


cstamas:
This PR needs more: instead of doing hoops and loops (all to circumvent enforcer on dependencies and more) just raise the build Java requirement to 21, as HttpClient::close method is added. We can keep this "fine grained" Java support (on 19 bind parameter, on 21 client close), but IMHO is overkill...

jira-importer:
Resolve #1098

PR #357, opened 2023-11-08
[MRESOLVER-302] Introduce onSessionClose

cstamas:
And make it used in new HTTP/2 transports.

Note about implementation:
* code base integrating Resolver 2.x (like Maven) SHOULD switch to new session handling
* code base using resolver w/o managing it (like Maven Mojos are) does NOT have to move to new session handling (basically resolver for them work unchanged).

In short, the Resolver 1.x session handling is still present, but is deprecated (DefaultRepositorySystemSession default ctor) and will emit warnings if application integrating Resolver 2.x still uses Resolver 1.x session handling and is about to use new Resolver 2.x features (like HTTP/2 transport) that require onSessionClose. Doing that will produce leaks, unless they are one-time CLI apps.

---

https://issues.apache.org/jira/browse/MRESOLVER-302


maven-resolver-api/src/main/java/org/eclipse/aether/RepositorySystem.java
```diff
@@ -304,8 +314,26 @@ List<RemoteRepository> newResolutionRepositories(
...
+     * Repository system may perform some resource cleanup, if applicable. Not using this method may cause leaks or
+     * unclean shutdown of some subsystem.
+     * <p>
+     * When shutdown happens, all the registered on-close handlers will be invoked (even if some throws), and at end
+     * of operation a {@link MultiRuntimeException} may be thrown, signaling that some handler(s) failed. This exception
+     * may be ignored, is at the discretion of caller.
+     *
+     * @since TBD
+     */
+    void close();
```
gnodet, line 339: Do we want:
```
default void close() {
    shutdown();
}
```
This would help with compatibility, but I don't think anyone is _implementing_ this interface, right ? Maybe wrapped ?

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultRepositorySystemLifecycle.java
```diff
@@ -69,6 +74,59 @@ public void addOnSystemEndedHandler(Runnable handler) {
...
+    public void sessionEnded(CloseableRepositorySystemSession session) {
+        requireNonNull(session, "session cannot be null");
+        requireNotShutdown();
+        String sessionId = session.sessionId();
+        ArrayList<Runnable> handlers = new ArrayList<>();
+        onSessionEndedHandlers.compute(sessionId, (k, v) -> {
+            if (v == null) {
+                throw new IllegalStateException("session instance not registered");
+            }
+            handlers.addAll(v);
```
gnodet, line 100: Why not `handlers = v;` ? and initialize `handlers = Collections.emptyList()` ?

maven-resolver-api/src/main/java/org/eclipse/aether/RepositorySystem.java
kwin, line 338: What about a default implementation just calling shutdown()?

maven-resolver-api/src/main/java/org/eclipse/aether/RepositorySystem.java
```diff
@@ -304,8 +314,26 @@ List<RemoteRepository> newResolutionRepositories(
...
+     * <p>
+     * Repository system may perform some resource cleanup, if applicable. Not using this method may cause leaks or
+     * unclean shutdown of some subsystem.
+     * <p>
+     * When shutdown happens, all the registered on-close handlers will be invoked (even if some throws), and at end
+     * of operation a {@link MultiRuntimeException} may be thrown, signaling that some handler(s) failed. This exception
+     * may be ignored, is at the discretion of caller.
+     *
+     * @since TBD
+     */
```
kwin, line 337:
```suggestion
     */
     @Override
```

maven-resolver-api/src/main/java/org/eclipse/aether/RepositorySystemSession.java
```diff
@@ -48,6 +51,350 @@
  */
 public interface RepositorySystemSession {
 
+    /**
+     * Immutable session that is closeable, can have onClose handlers registered and should be handled as a resource.
+     * These session instances can be created with {@link SessionBuilder}.
+     *
+     * @since TBD
+     */
+    interface CloseableRepositorySystemSession extends RepositorySystemSession, Closeable {
```
kwin, line 60: why not making the `RepositorySystemSession` always implement `Closeable` with an empty default method `close()`

maven-resolver-api/src/main/java/org/eclipse/aether/RepositorySystem.java
cstamas: is considered "incompatible binary change" by japicmp and i kinda agree with it

maven-resolver-api/src/main/java/org/eclipse/aether/RepositorySystemSession.java
cstamas: because that will suddenly emit warning in all the existing code (think Mojos) that "using closeable reasource without try-with-resource" or alike. Closing session is really objective of "resolver managing" client code (like Maven), but "resolver using (but not managing it)" client code (like Maven Mojos) should really not be annoyed is session closeable or not, they should NEVER close it anyway,

maven-resolver-api/src/main/java/org/eclipse/aether/RepositorySystem.java
kwin: Well IMHO it is more binary backwards incompatible to add a new method without default impl to an interface which is exposed (and can potentially be implemented by anyone). In reality the default impl almost never causes any issues, for details refer to https://docs.oracle.com/javase/specs/jls/se11/html/jls-13.html#jls-13.5.6.

cstamas: well, stop a bit: look at interface top and also read this: https://maven.apache.org/resolver/api-compatibility.html

In short: no client is EVER intended to extend or implement this interface. IF someone shoots in his leg afterwards all these warnings, let it be, but is not our fault.

maven-resolver-transport-jdk-parent/maven-resolver-transport-jdk-11/src/main/java/org/eclipse/aether/transport/jdk/JdkHttpTransporter.java
```diff
@@ -417,7 +421,14 @@ protected PasswordAuthentication getPasswordAuthentication() {
                         });
                     }
 
-                    return builder.build();
+                    HttpClient result = builder.build();
+                    // Only possible in Java21
```
gnodet, line 425: What's the plan here ? Should that be compiled with JDK 21 and with a check on the JDK used at runtime ?

cstamas: There is already another PR that tackles this area: https://github.com/apache/maven-resolver/pull/356
IMO we should merge that, and then this PR should just raise build time JDK requirement to Java 21 and uncomment these lines.

maven-resolver-api/src/main/java/org/eclipse/aether/RepositorySystem.java
gnodet: I've added meta annotations in the Maven api for that:
  https://maven.apache.org/ref/4.0.0-alpha-8/apidocs/org/apache/maven/api/annotations/Provider.html

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultRepositorySystemLifecycle.java
cstamas: that would make `handlers` non-immutable, no? This is within lambda...

maven-resolver-api/src/main/java/org/eclipse/aether/RepositorySystem.java
kwin: Ok, but then I fail to see how a default method calling `shutdown(...)` may break backwards compatibility.

kwin: This is already discussed in https://github.com/apache/maven-resolver/pull/357#discussion_r1386875573 :-)

gnodet (approved):
LGTM

olamy (approved):

PR #359, opened 2023-11-08
[MRESOLVER-422] Fix javadoc configuration

cstamas:
Not much to fix really, but simply put the "minimally required Java" and Javadoc URL needs to be kept in sync. But, as former is a range, latter is a "base version", it goes only manually.

Hence, I left a note in POM for this.

Site related fixes:
* Javadoc craps out on 19 and 21 JDK transport modules (nothing to document).
* PMD does not supports Java 21.

---

https://issues.apache.org/jira/browse/MRESOLVER-422


jira-importer:
Resolve #1103

PR #365, opened 2023-11-15
[MRESOLVER-372] Rework the FileUtils collocated temp file

cstamas:
Fixes:
* move() call should NOT perform the move, as writer stream to tmp file may still be open
* move the file move logic to close, make it happen only when closing collocated temp file
* perform fsync before atomic move to ensure there is no OS dirty buffers related to newly written file
* on windows go with old code that for some reason works (avoid NIO2)
* on non-Win OS fsync the parent directory as well.

---

https://issues.apache.org/jira/browse/MRESOLVER-372

Backport to 1.9.x branch of the https://github.com/apache/maven-resolver/pull/364


slawekjaranowski (approved):

slawekjaranowski (approved):

jira-importer:
Resolve #1049

PR #373, opened 2023-11-20
[MRESOLVER-437][MRESOLVER-438] Do not force set Expect-Continue setting

cstamas:
But allow users to override default (that comes from transport being in use). Also, ignore this setting on Java versions below 20. Finally, align class names for jdk transport as they were inconsistent.

---

https://issues.apache.org/jira/browse/MRESOLVER-437
https://issues.apache.org/jira/browse/MRESOLVER-438


rmannibucau (approved):
LGTM with a small side note (not blocking)

side note: think default http version should be 1.1 (until we enhance our protocol to use it somehow) cause http 2 has several issues (a lot of servers and some which can impact clients). Boost is not that huge once keep alive is on for us so let's not abuse it in our defaults?

PS: thanks for the hyphen typo fix ;)

jira-importer:
Resolve #1108

jira-importer:
Resolve #1112

jira-importer:
Resolve #1112

jira-importer:
Resolve #1108

PR #379, opened 2023-11-27
[MRESOLVER-429] Emit more informative message on connection refused

cstamas:
By default HTTP Client is quite muted, message is "ConnectException". As this "message enhancement" already happens on a branch that leads to error, the throwing of new exception and "decorating it" really does not matter, as the end result is anyway: resolver invocation failure (in Maven is build failure).

---

https://issues.apache.org/jira/browse/MRESOLVER-429


jira-importer:
Resolve #1104

jira-importer:
Resolve #1104

PR #380, opened 2023-11-27
[MRESOLVER-321] Make collection and visiting interruptable

cstamas:
Currently the two collector implementations are not interrupt-able, do not sense interruption directly, only by some side-effect like IO. Moreover, the BF new collector may enter a "busy loop" as we seen (it was due bug, but nothing prevents us to have more bugs). Make main loop in both collector detect thread interruption and use global (per-collection) Args to carry state of the, interruption effectively the whole ST or MT collection.

Same stands for dependency visitor: it may also enter busy loop in case of cycles, hence, it should be made also  interrupt-able. Visitor OTOH should not reset the interrupted flag, it should just stop when it is set, and let the flag be handled at higher level (for example in collector).

---

https://issues.apache.org/jira/browse/MRESOLVER-321

### DIFF
--- a/maven-resolver-api/src/main/java/org/eclipse/aether/graph/DefaultDependencyNode.java
+++ b/maven-resolver-api/src/main/java/org/eclipse/aether/graph/DefaultDependencyNode.java
@@ -286,7 +286,7 @@ public final class DefaultDependencyNode implements DependencyNode {
     }
 
     public boolean accept(DependencyVisitor visitor) {
-        if (Thread.interrupted()) {
+        if (Thread.currentThread().isInterrupted()) {
             throw new RuntimeException("Thread interrupted");
         }
         if (visitor.visitEnter(this)) {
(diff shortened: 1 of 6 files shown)


maven-resolver-api/src/main/java/org/eclipse/aether/graph/DefaultDependencyNode.java
```diff
@@ -286,6 +286,9 @@ public void setData(Object key, Object value) {
     }
 
     public boolean accept(DependencyVisitor visitor) {
+        if (Thread.currentThread().isInterrupted()) {
+            throw new RuntimeException("Thread interrupted");
```
gnodet, line 290: We need a way for the caller to know that the exception was caused by the thread being interrupted.
At least `new RuntimeException(new InterruptedException("Thread interrupted"))` ?

gnodet (changes requested):

maven-resolver-api/src/main/java/org/eclipse/aether/graph/DefaultDependencyNode.java
cstamas: fixed

gnodet (approved):

jira-importer:
Resolve #995

jira-importer:
Resolve #995

PR #382, opened 2023-11-28
[MRESOLVER-440] Addendum

cstamas:
Seems the "atypical" modules like supplier and
demo snippets were completely omitted from the
transport rename, and were still using renamed
transport-http instead the new transport-apache.

---

https://issues.apache.org/jira/browse/MRESOLVER-440


rmannibucau (approved):

jira-importer:
Resolve #1114

PR #385, opened 2023-12-04
[MRESOLVER-443] Integrate tool into build

cstamas:
Simply invoke it as very last step of build to generate the documentation. Ideally, this will produce _same output_ hence there will be no side effect (no local change). But in case page is outdated (needs update) the local checkout becomes dirty and would prevent release for example.

This is not "final" solution, but would work IMHO. Current issue with tool is that:
* it must have access to up-to-date class files (to javap them), hence it runs as last module of multi module build
* it must process WHOLE project, so to say "aggregate" output, it cannot run per module

---

https://issues.apache.org/jira/browse/MRESOLVER-443

### DIFF
--- a/maven-resolver-tools/pom.xml
+++ b/maven-resolver-tools/pom.xml
@@ -73,4 +73,30 @@
       <scope>runtime</scope>
     </dependency>
   </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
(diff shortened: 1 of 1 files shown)


cstamas:
@michael-o anything to add here?

jira-importer:
Resolve #1118

PR #386, opened 2023-12-04
[MRESOLVER-335] Proposal for better errors

cstamas:
Basically drop the method building text message (and build something very basic instead). It is the application that should build proper output, as it does have access to org.eclipse.aether.resolution.ArtifactResolutionException#getResults and can and should build meaningful error instead.

This PR does not fixes the issue for Maven, it merely does these things:
* alters the message built by ArtifactResolutionException that does lack information to be complete
* but IMHO Maven is not meant to use this exception message as is (as it happens today), instead
* this PR makes possible that in Maven code build proper message, as it now exposes everything (maps remote repository to exception)

---

https://issues.apache.org/jira/browse/MRESOLVER-335


cstamas:
@slawekjaranowski as reporter, anything to add?

cstamas:
One thing am unsure about: `Map<ArtifactRepository, Exception> exceptions` in ArtifactResult... I think it should be `Map<ArtifactRepository, List<Exception>> exceptions` instead...

cstamas:
Reworked it and is much simpler change (and is even backward compat, so no tampering with japicm needed)

jira-importer:
Resolve #1006

jira-importer:
Resolve #1006

PR #388, opened 2023-12-05
Add a flattenDependencyNodes method accepting a DependencyFilter

gnodet:


maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultRepositorySystem.java
```diff
@@ -303,11 +299,19 @@ public DependencyResult resolveDependencies(RepositorySystemSession session, Dep
...
     @Override
     public List<DependencyNode> flattenDependencyNodes(RepositorySystemSession session, DependencyNode root) {
+        return flattenDependencyNodes(session, root, null);
+    }
+
+    @Override
+    public List<DependencyNode> flattenDependencyNodes(
+            RepositorySystemSession session, DependencyNode root, DependencyFilter filter) {
         validateSession(session);
         requireNonNull(root, "root cannot be null");
```
cstamas, line 309: this should happen only in public facing method (but maybe even there make it nullable?) As originally it CAN be null...  See original code:

```
        if (result.getRoot() != null) {
            result.getRoot().accept(visitor);
        }
```

PR #391, opened 2023-12-07
[MRESOLVER-449] Step toward testing all HTTP transports

cstamas:
Right now, the HTTP test suite is pulled out and reused for all HTTP transports. It passes for ApacheTransporter (as it was pulled from it), but Jetty and JDK show some signs of some problems...

Current results (note: base class HttpTransporterTest has 70 tests, while some transport specific UTs add more):
* apache :heavy_check_mark: 74 test OK
* jdk :red_circle: 62 test OK, 9 test FAIL
* jetty :red_circle: 64 test OK, 6 FAIL

The "basics" are overall good (basics pass everywhere), failures stems from small differences.

Jetty: it seems it consumes the body (pulls bytes out of it) even before all the HTTP auth happens, this causes that `TransportListener` gets notified "transport started" but in fact is not and test asserts it to not. Also, Jetty throws in tests where server is set up to abruptly close the connection. Retries are not properly done (low level, like abrupt connection drop).

JDK: same problem with body (pulls bytes beforehand) and similarly as with Jetty, "transport started" events fails the test. Preemptive auth is not done (as it seems it is either "manual" handling of auth or Authenticator is used but then preemptive auth becomes impossible as headers are removed, no way to set them). It also throws in tests where server abruptly drops connection. Same for retries (in case of TCP issue).

---

https://issues.apache.org/jira/browse/MRESOLVER-449


cstamas:
I am about to merge this PR as:
* it covers all HTTP transports with tests
* contains substantial bug fixes in two new transports
* disables the failing tests for now....

We can always reiterate on this, IMHO the bug fixes are important (and should go into alpha-4)

jira-importer:
Resolve #1123

jira-importer:
Resolve #1123

PR #394, opened 2023-12-11
[MRESOLVER-335] Proposal for better errors (addendum)

cstamas:
Undo most of the changes to ArtifactResolutionException, as Maven Core depends on it's logic (is flawed for message, but is not to drive logic by checking cause).

Previos main PR did expose all the needed things anyway, so basically it is now Maven turn to instead reusing the (flawed) message to emit it to end user/console, build some better message using new methods.

---

https://issues.apache.org/jira/browse/MRESOLVER-335


jira-importer:
Resolve #1006

PR #395, opened 2023-12-12
[MRESOLVER-451] Tidy up version filters

cstamas:
Tidy up the classes and add "lowest" and "predicate" filter as well. For "exposing" not much needed, is rather client stuff to do: set it on session.

---

https://issues.apache.org/jira/browse/MRESOLVER-451


jira-importer:
Resolve #1124

PR #400, opened 2023-12-19
[MRESOLVER-459] Rework session to be usable in tests

cstamas:
Without touching deprecated code. Also, remove the redundancy by duplicating this class in test utils.

---

https://issues.apache.org/jira/browse/MRESOLVER-459


jira-importer:
Resolve #1135

PR #401, opened 2023-12-19
[MRESOLVER-460] Implement top H/L

cstamas:
Extend filters to implement top/bottom

---

https://issues.apache.org/jira/browse/MRESOLVER-460


jira-importer:
Resolve #1134

PR #404, opened 2024-01-10
[MRESOLVER-463] sort checksum summary file per path and not checksum to be human friendly

rmannibucau:


cstamas (approved):

cstamas:
If am not mistaken, this can easily be backported to Resolver 1.x as well... There are no issues scheduled for 1.9.18, but still...

rmannibucau:
Yes, no strong opinion on that since I'm more aiming at mvn 4 for this one but if you feel it is worth it also happy to backport (can be delayed of +-1week that said due to personal constraints)

PR #407, opened 2024-01-17
[MRESOLVER-465] Make room for alternate version scheme

cstamas:
Since eons Resolver had the "generic" scheme implementing the `Version`, `VersionRange` and `VersionConstraint` API interfaces.

Interesting fact: `GenericVersionRange` and `GenericVersionConstraint` are NOT tied in any way to `GenericVersion` implementation, they are happy with plain `Version` interface as well, they are truly "generic".

This means, that a scheme needs really only to provide `Version` implementation (currently in form of `GenericVersion`) and a "factory" for it (currently in a form of `GenericVersionScheme`).

Refactor a bit, and make room for alternate Version implementations.

Fact: the change can be done in source and binary compatible way, and is enforced by japicmp.

---

https://issues.apache.org/jira/browse/MRESOLVER-465

### DIFF
--- a/maven-resolver-util/src/main/java/org/eclipse/aether/util/version/GenericVersionConstraint.java
+++ b/maven-resolver-util/src/main/java/org/eclipse/aether/util/version/GenericVersionConstraint.java
@@ -28,6 +28,8 @@ import static java.util.Objects.requireNonNull;
 
 /**
  * A constraint on versions for a dependency.
+ * <p>
+ * Despite its name, this class is generic in a sense it works with any {@link Version}.
  */
 final class GenericVersionConstraint implements VersionConstraint {
 
@@ -88,9 +90,9 @@ final class GenericVersionConstraint implements VersionConstraint {
             return false;
         }
 
-        GenericVersionConstraint that = (GenericVersionConstraint) obj;
+        VersionConstraint that = (VersionConstraint) obj;
 
-        return Objects.equals(range, that.range) && Objects.equals(version, that.getVersion());
+        return Objects.equals(range, that.getRange()) && Objects.equals(version, that.getVersion());
     }
 
     @Override
--- a/maven-resolver-util/src/main/java/org/eclipse/aether/util/version/GenericVersionRange.java
+++ b/maven-resolver-util/src/main/java/org/eclipse/aether/util/version/GenericVersionRange.java
@@ -23,13 +23,17 @@ import java.util.Objects;
 import org.eclipse.aether.version.InvalidVersionSpecificationException;
 import org.eclipse.aether.version.Version;
 import org.eclipse.aether.version.VersionRange;
+import org.eclipse.aether.version.VersionScheme;
 
 import static java.util.Objects.requireNonNull;
 
 /**
  * A version range inspired by mathematical range syntax. For example, "[1.0,2.0)", "[1.0,)" or "[1.0]".
+ * <p>
+ * Despite its name, this class is generic in a sense it works with any {@link Version}
  */
 final class GenericVersionRange implements VersionRange {
+    private final VersionScheme versionScheme;
 
     private final Bound lowerBound;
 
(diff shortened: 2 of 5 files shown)


maven-resolver-util/src/main/java/org/eclipse/aether/util/version/GenericVersionConstraint.java
```diff
@@ -28,6 +28,8 @@
 
 /**
  * A constraint on versions for a dependency.
+ * <p>
+ * Despite its name, this class is generic in a sense it works with any {@link Version}.
```
gnodet, line 32: Who cares if the classes are protected and final ?
This only makes sense if these implementation classes are made `public`.

maven-resolver-util/src/main/java/org/eclipse/aether/util/version/GenericVersionRange.java
```diff
@@ -23,13 +23,17 @@
...
 import org.eclipse.aether.version.Version;
 import org.eclipse.aether.version.VersionRange;
+import org.eclipse.aether.version.VersionScheme;
 
 import static java.util.Objects.requireNonNull;
 
 /**
  * A version range inspired by mathematical range syntax. For example, "[1.0,2.0)", "[1.0,)" or "[1.0]".
+ * <p>
+ * Despite its name, this class is generic in a sense it works with any {@link Version}
```
gnodet, line 33: Same

maven-resolver-util/src/main/java/org/eclipse/aether/util/version/VersionSchemeSupport.java
```diff
@@ -0,0 +1,101 @@
...
+ * @since 2.0.0
+ */
+public abstract class VersionSchemeSupport implements VersionScheme {
+    @Override
+    public GenericVersionRange parseVersionRange(final String range) throws InvalidVersionSpecificationException {
+        return new GenericVersionRange(this, range);
+    }
+
+    @Override
+    public GenericVersionConstraint parseVersionConstraint(final String constraint)
```
gnodet, line 44: Same.
This should even cause a warning as `GenericVersionConstraint` is exported outside it's visibility.

maven-resolver-util/src/main/java/org/eclipse/aether/util/version/VersionSchemeSupport.java
```diff
@@ -0,0 +1,101 @@
...
+            result = new GenericVersionConstraint(parseVersion(constraint));
+        } else {
+            result = new GenericVersionConstraint(UnionVersionRange.from(ranges));
+        }
+
+        return result;
+    }
+
+    @Override
+    public boolean equals(final Object obj) {
```
gnodet, line 89: Does it make sense at all ? This isn't a data object.  If there's a need, they should be identified by a name imho.

maven-resolver-util/src/main/java/org/eclipse/aether/util/version/VersionSchemeSupport.java
```diff
@@ -0,0 +1,101 @@
...
+        if (this == obj) {
+            return true;
+        }
+
+        return obj != null && getClass().equals(obj.getClass());
+    }
+
+    @Override
+    public int hashCode() {
```
gnodet, line 98: Useless

gnodet (changes requested):

cstamas:
@gnodet agreed with all, but _I did not change any of these_, all I did is split the class (support + generic scheme). The idea is to have new version scheme provided by Resolver in this very same package (not by "third party" as that would open up a can of worms).

gnodet:
> @gnodet agreed with all, but _I did not change any of these_, all I did is split the class (support + generic scheme). The idea is to have new version scheme provided by Resolver in this very same package (not by "third party" as that would open up a can of worms).

Then `VersionSchemeSupport` should not be public.  That will remove almost all those problems.

cstamas:
Oh, yes, true, oversight on my part. Fixed

gnodet:
> Oh, yes, true, oversight on my part. Fixed

Much better now.  Do we need to keep the hashCode / equals on version scheme ? They really look weird on a pure service class.

cstamas:
Removed them.

gnodet (approved):

jira-importer:
Resolve #1142

jira-importer:
Resolve #1142

PR #411, opened 2024-01-24
[MRESOLVER-470] Clean up Artifact properties use

cstamas:
Introduce new Artifact ctor w/ ArtifactType. Clean up some Javadoc, make it clear what we mean here.

---

https://issues.apache.org/jira/browse/MRESOLVER-470


jira-importer:
Resolve #1144

PR #419, opened 2024-01-27
[MRESOLVER-483] Fix path concatenation logic

cstamas:
As iterator may have more, but no file check is done

---

https://issues.apache.org/jira/browse/MRESOLVER-483


cstamas:
Superseded by https://github.com/apache/maven-resolver/pull/421

jira-importer:
Resolve #1156

PR #421, opened 2024-01-27
[MRESOLVER-483] Classpath generation bugfix

cstamas:
Rework whole class to use Streams

---

https://issues.apache.org/jira/browse/MRESOLVER-483


jira-importer:
Resolve #1156

PR #425, opened 2024-02-06
[MRESOLVER-491] Resolver should not have any "predefined" scope

cstamas:
It is duty of consumer application to define those. Also, no need for SPI either, as essentialy ONLY the dependency managers were using the "system" scope definition so far.

Now, client code, that instantianates these anyway are able to pass in predicate.

Note: as util is japicmp checked, the def ctors MUST stay, but are deprecated.

---

https://issues.apache.org/jira/browse/MRESOLVER-491

### DIFF
--- a/maven-resolver-impl/src/test/java/org/eclipse/aether/internal/impl/collect/DependencyCollectorDelegateTestSupport.java
+++ b/maven-resolver-impl/src/test/java/org/eclipse/aether/internal/impl/collect/DependencyCollectorDelegateTestSupport.java
@@ -19,15 +19,7 @@
 package org.eclipse.aether.internal.impl.collect;
 
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.concurrent.atomic.AtomicReference;
 
 import org.eclipse.aether.DefaultRepositorySystemSession;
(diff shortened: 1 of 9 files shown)


maven-resolver-impl/src/test/java/org/eclipse/aether/internal/impl/collect/DependencyCollectorDelegateTestSupport.java
```diff
@@ -387,7 +379,7 @@ void testManagedVersionScope() throws DependencyCollectionException {
         Dependency dependency = newDep("managed:aid:ext:ver");
         CollectRequest request = new CollectRequest(dependency, singletonList(repository));
 
-        session.setDependencyManager(new ClassicDependencyManager());
+        session.setDependencyManager(new ClassicDependencyManager(s -> Objects.equals(s, "system")));
```
gnodet, line 382: Why not `"system"::equals` ?

cstamas: done

slawekjaranowski (approved):

maven-resolver-util/src/test/java/org/eclipse/aether/util/graph/manager/DependencyManagerTest.java
```diff
@@ -40,6 +41,8 @@
  */
 public class DependencyManagerTest {
 
+    private final SystemScopePredicate systemScopePredicate = "system"::equals;
```
gnodet, line 44: Would it make sense to have this constant shared for the whole codebase (and deprecated) instead of having this predicate  defined dozens of time in various places ?

gnodet (approved):
I've added a constant definition for the system predicate in tests, otherwise LGTM

cstamas:
@gnodet looks good

jira-importer:
Resolve #1169

PR #428, opened 2024-02-10
[MRESOLVER-494] The LOCAL_PATH system property is client app property

cstamas:
This ArtifactProperty is specific to "system" scope, tied to it. Introduce a "system scope handler" that does all, and let client app control all. Also, this component belongs to session, so expose it via session.

---

https://issues.apache.org/jira/browse/MRESOLVER-494


maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/DependencyCollectorDelegate.java
```diff
@@ -394,8 +393,8 @@ protected VersionRangeResult cachedResolveRangeResult(
         return rangeResult;
     }
 
-    protected static boolean isLackingDescriptor(Artifact artifact) {
-        return artifact.getProperty(ArtifactProperties.LOCAL_PATH, null) != null;
+    protected static boolean isLackingDescriptor(RepositorySystemSession session, Artifact artifact) {
+        return session.getSystemScopeHandler().getSystemPath(artifact.getProperties()) != null;
```
cstamas, line 397: This makes me think that "system" scope and artifact with LOCAL_PATH set are one and same. We may want to handle that as one, instead to have them separated?

gnodet: I'm not really sure to understand what you're suggesting here...

cstamas: If you look at "system" scope definition, there is this sentence "Dependency in this scope does not have artifact descriptor either.". And this here is the check "does it have descriptor. And since Resolver 1.0 the actual check is done like this: `artifact.getProperty( ArtifactProperties.LOCAL_PATH, null ) != null` (presence of property and not actual scope).

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/manager/AbstractDependencyManager.java
```diff
@@ -43,7 +42,7 @@ public abstract class AbstractDependencyManager implements DependencyManager {
      * @deprecated To be removed when deprecated constructors are removed.
      */
     @Deprecated
-    protected static final SystemScopePredicate SYSTEM_PREDICATE = "system"::equals;
+    protected static final SystemScopeHandler SYSTEM_PREDICATE = SystemScopeHandler.LEGACY;
```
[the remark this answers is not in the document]
cstamas: Fixed

maven-resolver-api/src/main/java/org/eclipse/aether/SystemScopeHandler.java
```diff
@@ -0,0 +1,96 @@
...
+     *
+     * @return the system path from passed in properties, or {@code null} if not present.
+     */
+    String getSystemPath(Map<String, String> properties);
+
+    /**
+     * Sets system path in properties. The passed in {@code systemPath} can be {@code null}, in which case this is
+     * "remove" operation (or "unset").
+     */
+    void setSystemPath(Map<String, String> properties, String systemPath);
```
gnodet, line 71: This method does not make much sense to me. It should not be needed, as the code calling the resolver is now responsible for the system scope.  The resolver should not alter what has been given.

maven-resolver-api/src/main/java/org/eclipse/aether/SystemScopeHandler.java
```diff
@@ -0,0 +1,96 @@
...
+ * the backing file should reside. Resolution of artifacts in this scope fails, if backing file does not exist
+ * (no property set, or property contains invalid path, or the path points to a non-existent file).
+ *
+ * @since 2.0.0
+ */
+public interface SystemScopeHandler {
+    /**
+     * Returns {@code true} if given scope label is "system" scope.
+     */
+    boolean isSystemScope(String scope);
```
gnodet, line 44: Is `scope` supposed to be nullable ? It seems the code is given `null`.  If that's really expected we should indicate that on the javadoc.

maven-resolver-api/src/main/java/org/eclipse/aether/SystemScopeHandler.java
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * Returns {@code true} if given dependency is in "system" scope.
+     */
+    default boolean isSystemScope(Dependency dependency) {
+        return isSystemScope(dependency.getScope());
```
gnodet, line 50: Is `dependency` nullable or not ?
If yes, we should test for a null `dependency` and have:
```
    return dependency != null && isSystemScope(dependency.getScope());
```

maven-resolver-api/src/main/java/org/eclipse/aether/SystemScopeHandler.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    /**
+     * Returns {@code true} if given dependency node dependency is in "system" scope.
+     */
+    default boolean isSystemScope(DependencyNode dependencyNode) {
+        return dependencyNode.getDependency() != null && isSystemScope(dependencyNode.getDependency());
```
gnodet, line 57: `dependencyNode` may be null here too

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/manager/AbstractDependencyManager.java
```diff
@@ -199,25 +199,27 @@ public DependencyManagement manageDependency(Dependency dependency) {
...
-                if (!systemScopePredicate.test(scope)
-                        && dependency.getArtifact().getProperty(ArtifactProperties.LOCAL_PATH, null) != null) {
+                if (!systemScopeHandler.isSystemScope(scope)
+                        && systemScopeHandler.getSystemPath(
+                                        dependency.getArtifact().getProperties())
+                                != null) {
                     Map<String, String> properties =
                             new HashMap<>(dependency.getArtifact().getProperties());
-                    properties.remove(ArtifactProperties.LOCAL_PATH);
+                    systemScopeHandler.setSystemPath(properties, null);
```
gnodet, line 208: I fail to see how this code is relevant. 
If this is really needed ? that would mean that the scope isn't solely determined by the `dependency.getScope()`.

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/manager/AbstractDependencyManager.java
```diff
@@ -199,25 +199,27 @@ public DependencyManagement manageDependency(Dependency dependency) {
...
                     management.setProperties(properties);
                 }
             }
 
-            if ((systemScopePredicate.test(scope))
-                    || (scope == null && systemScopePredicate.test(dependency.getScope()))) {
+            if ((systemScopeHandler.isSystemScope(scope))
+                    || (scope == null && systemScopeHandler.isSystemScope(dependency.getScope()))) {
                 String localPath = managedLocalPaths.get(key);
```
gnodet, line 215: The managedLocalPaths should be integrated into the systemScopeHandler imho.

gnodet (changes requested):

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/manager/AbstractDependencyManager.java
cstamas: Obviously is not, that was my concern [here](https://github.com/apache/maven-resolver/pull/428#discussion_r1485317539).

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/manager/AbstractDependencyManager.java
cstamas: systemScopeHandler is session scoped, while managedLocalPaths is per mgr, and mgr is created as needed (see "deriveChildManager"), so unsure how could we do that.

maven-resolver-api/src/main/java/org/eclipse/aether/SystemScopeHandler.java
cstamas: Problem is how `Artifact` is immutable, so it does not have `setProperty` method. Also, due this, `Artifact#setProperties` returns new Artifact instance (as almost every mutator method in Resolver).

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/manager/AbstractDependencyManager.java
gnodet: I don't really see a use case in the code where `scope == "system"` and `localPath != null` are not interchangeable.  So I think we don't need to manage the local paths, they'll just be retrieved from the `SystemScopeHandler` when needed using `getSystemPath(Artifact)`.

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/manager/AbstractDependencyManager.java
gnodet: Not needed or not only determined by the scope ?

maven-resolver-api/src/main/java/org/eclipse/aether/SystemScopeHandler.java
gnodet: I don't think we need this method.  We can just delay until the very end of the resolution, then call `SystemScopeManager#getSystemPath()` and put it as the resolved path of the artifact.
We don't need to carry the path for the duration of the resolution imho.

gnodet:
Here's what I have in mind: https://github.com/gnodet/maven-resolver/commit/791c07e3fc05a86fa383607d28db0dcdf7fcc995

maven-resolver-api/src/main/java/org/eclipse/aether/SystemScopeHandler.java
cstamas: This method is used in two cases:
* when dependency scope is managed, and is not "system", unset the LOCAL_PATH if set
* when dependency scope is managed, and is "system", make sure LOCAL_PATH is set

Given this may happen on any (n-th level of depth), am again unsure from where would you get (and which) artifact to call with getSystemPath()? As now, especially with introduction of "transitive dep mgr", we should track introduced depMgr sections during construction of whole graph (unlike it was in Maven3, only 2 depths).

maven-resolver-api/src/main/java/org/eclipse/aether/SystemScopeHandler.java
cstamas: I think javadoc is quite clear on this matter. If you consider `null` "system" scope, then yet, return `true` here.

cstamas:
> @cstamas Here's what I have in mind: [gnodet@791c07e](https://github.com/gnodet/maven-resolver/commit/791c07e3fc05a86fa383607d28db0dcdf7fcc995) The local path is only managed by the scope handler. The assumption is that for a given artifact, if it has a system scope, it also has a local path. Maybe I'm missing something though ...

You do: i.e. you have a dep w/o scope, but you define a depMgr w/ system scope (and system path), who will apply then that mgmt onto dependency? Also, this may come from n-th level now, as we introduced "transitive depmgr" etc.

gnodet (approved):

slawekjaranowski (approved):

jira-importer:
Resolve #1164

PR #432, opened 2024-02-19
[MRESOLVER-301] Artifact generators

cstamas:
Artifact generators.

---

https://issues.apache.org/jira/browse/MRESOLVER-301


slawekjaranowski:
At first time looks a complicated a bit.

I can test it in next weekend - I need more time for it. 

first questions:
 - What is difference about gpg singnature and md5, sha checksums - why not use the same same existing implementation - we simply want generate a checksum or signature
 
- What future will be for m-gpg-p when we move generating of signatures to resolver and finaly to Maven core ... yes I see that it is a extension but ...

By the way I think we should discuss such (for me big) architecture change on ML.

cstamas:
Q1: checksums: they are _tightly coupled_ into "basic connector". Historically, "connector" was the transport abstraction, but later "connector + transport" was introduced, and connector deals with checksums ONLY. Original issue was "deal with signatures as with checksums" but this is a no-go, as connector is (theoretically) swappable, while true, we have only "basic" connector and no other implementation. IMHO, we should eventually remove the abstraction from connector and accept that "basic" is THE connector we use.
Q2: First, it is an extension, yes. Second, I see no problem with resolver signing, as to me, it seems like "natural fit". If you think about, signer plugins usually do something and "attach" the result. Am fine with both, and as far I am concerned, the two can coexists even (maybe add some logic to signer to remain dormant if signatures discovered).

cstamas:
Also, "artifact generator" is one thing, and its use for "signing" is another.

If we remain at "publishing to Central" domain, where PGP signature is enforced, and signing, I am not satisfied with any of existing solutions:
* maven-sign-plugin uses gpg executable
* takari-sign-plugin cannot do ED25519 (but have cool ideas)
* s4u sign plugin unused in ASF (but have cool ideas)

So I just "brought" the best of all here. At least, that was my intent. And yes, IMO, "signing" is natural fit for "artifact generator" and IMO we should not complicate our build/POMs for something that _is an expected requirement_ (is like we'd need to add a plugin to POM to create checksums, something also required to publish to Central).

Also, "signer" is extensible, so it does not have to get GnuPG, it could be something else as well... so in this way, it is _not in Maven Core_ (wired in), but can progress and change, maybe as an extension.

michael-o (changes requested):

maven-resolver-generator-signer/src/main/java/org/eclipse/aether/generator/signer/gpg/GpgAgentPasswordLoader.java
```diff
@@ -0,0 +1,110 @@
...
+        }
+        return null;
+    }
+
+    private String load(long keyId, Path socketPath) throws IOException {
+        try (SocketChannel sock = SocketChannel.open(StandardProtocolFamily.UNIX)) {
+            try {
+                sock.connect(UnixDomainSocketAddress.of(socketPath));
+            } catch (SocketException e) {
+                System.out.println("Cannot connect to server on socket:" + socketPath);
```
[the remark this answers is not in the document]
cstamas: reworked this: dropped use of sout, using logger proper, and exposed socket config.

michael-o (commented):
I don't have any opinion and cannot object.

michael-o (commented):
I'll leave the decision to others.

cstamas:
> By the way I think we should discuss such (for me big) architecture change on ML.

@slawekjaranowski this is not "big architectural change" at all, just another option. I modified current PR to:
* enabled GnuPG signer step back, if signatures detected among deployable artifacts (ie bui;d uses plugin that produces `.asc` signatures)
* by default signer is disabled (even if present as extension)

maven-resolver-generator-signer/src/main/java/org/eclipse/aether/generator/signer/SignerArtifactGenerator.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+
+import org.eclipse.aether.artifact.Artifact;
+import org.eclipse.aether.spi.artifact.generator.ArtifactGenerator;
+
+import static java.util.Objects.requireNonNull;
+
+final class SignerArtifactGenerator implements ArtifactGenerator {
```
gnodet, line 32: Why not `public` instead of `final` with a public constructor and protected fields ?  Why would not people be allowed to extend it ?

maven-resolver-generator-signer/src/main/java/org/eclipse/aether/generator/signer/gpg/GpgEnvPasswordLoader.java
```diff
@@ -0,0 +1,48 @@
...
+
+import org.eclipse.aether.RepositorySystemSession;
+import org.eclipse.aether.util.ConfigUtils;
+import org.eclipse.sisu.Priority;
+
+@Singleton
+@Named(GpgEnvPasswordLoader.NAME)
+@Priority(50)
+@SuppressWarnings("checkstyle:magicnumber")
+public final class GpgEnvPasswordLoader implements GpgSignerFactory.KeyPasswordLoader {
```
gnodet, line 32: Is the `final` keyword needed ?

maven-resolver-generator-signer/src/main/java/org/eclipse/aether/generator/signer/gpg/GpgSigner.java
```diff
@@ -0,0 +1,132 @@
...
+import org.eclipse.aether.artifact.Artifact;
+import org.eclipse.aether.generator.signer.Signer;
+import org.eclipse.aether.spi.connector.layout.RepositoryLayout;
+import org.eclipse.aether.util.artifact.SubArtifact;
+
+/**
+ * GnuPG {@link Signer} implementation that is kept and reused across session.
+ */
+@SuppressWarnings("checkstyle:magicnumber")
+public final class GpgSigner implements Signer {
```
gnodet, line 49: Same...

maven-resolver-generator-signer/src/main/java/org/eclipse/aether/generator/signer/gpg/GpgSigner.java
```diff
@@ -0,0 +1,132 @@
...
+        signatureTempFiles.forEach(p -> {
+            try {
+                Files.deleteIfExists(p);
+            } catch (IOException e) {
+                p.toFile().deleteOnExit();
+            }
+        });
+    }
+
+    private void sign(InputStream content, OutputStream signature) throws IOException {
```
gnodet, line 114: `protected` ?

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultDeployer.java
```diff
@@ -151,7 +162,23 @@ private DeployResult deploy(SyncContext syncContext, RepositorySystemSession ses
             throw new DeploymentException("Failed to deploy artifacts/metadata: " + e.getMessage(), e);
         }
 
+        final List<Artifact> artifacts = new ArrayList<>(request.getArtifacts());
```
gnodet, line 165: I don't see how `final` is useful.  The static analysis very well knows if a variable is final or not.

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultInstaller.java
```diff
@@ -95,53 +106,96 @@ private InstallResult install(SyncContext syncContext, RepositorySystemSession s
 
         RequestTrace trace = RequestTrace.newChild(request.getTrace(), request);
 
-        List<? extends MetadataGenerator> generators = getMetadataGenerators(session, request);
+        final List<Artifact> artifacts = new ArrayList<>(request.getArtifacts());
```
gnodet, line 109: Once again ?

gnodet (approved):
Mostly coding style issues

cstamas:
For other a general answer: I do agree with all your comments, but
* Resolver whole codebase is done as this, and we may want to change all these, but for now there are a lot of package protected and final classes, I did not want to disturb this coherence. But I do agree about non-final/public/protected fields.
* BUT, extending JSR330 annotated component is a no-no (hence final)
* I like how these classes are being prevented to be tampered with. If someone needs this, can copy, but I don't want to be hindered a year after, as somecome tells me I broke "compatibility" with his app (where he uses or extends these classes).

So, yes, I agree, but this is why not IMO.

cstamas:
Added UT for signer (for now does not assert much). Also verified locally the whole signing workflow along with agent, and it all works nicely, while verified produced signatures with `gpg` CLI.

maven-resolver-generator-signer/src/main/java/org/eclipse/aether/generator/signer/SignerArtifactGeneratorFactory.java
```diff
@@ -0,0 +1,83 @@
...
+import org.eclipse.aether.artifact.Artifact;
+import org.eclipse.aether.deployment.DeployRequest;
+import org.eclipse.aether.installation.InstallRequest;
+import org.eclipse.aether.repository.ArtifactRepository;
+import org.eclipse.aether.spi.artifact.generator.ArtifactGenerator;
+import org.eclipse.aether.spi.artifact.generator.ArtifactGeneratorFactory;
+
+@Singleton
+@Named(SignerArtifactGeneratorFactory.NAME)
+public final class SignerArtifactGeneratorFactory implements ArtifactGeneratorFactory {
```
slawekjaranowski, line 39: It is only one implementation of  `SignerArtifactGeneratorFactory`
why we put it in new module ... maybe should be in `maven-resolver-impl`

maven-resolver-generator-signer/src/main/java/org/eclipse/aether/generator/signer/SignerFactory.java
```diff
@@ -0,0 +1,40 @@
...
+
+import org.eclipse.aether.RepositorySystemSession;
+import org.eclipse.aether.repository.ArtifactRepository;
+
+/**
+ * Component creating {@link Signer} instances.
+ *
+ * @since 2.0.0
+ */
+public interface SignerFactory {
```
slawekjaranowski, line 29: next factory similar to `ArtifactGeneratorFactory`

maven-resolver-generator-signer/src/main/java/org/eclipse/aether/generator/signer/SignerArtifactGenerator.java
```diff
@@ -0,0 +1,73 @@
...
+        try {
+            artifacts.addAll(generatedArtifacts);
+            ArrayList<Artifact> signatures = new ArrayList<>();
+            for (Signer signer : signers) {
+                Collection<Artifact> s = signer.sign(artifacts);
+                signatures.addAll(s.stream()
+                        .map(a -> {
+                            HashMap<String, String> properties = new HashMap<>(a.getProperties());
+                            properties.put(
+                                    SignerArtifactGeneratorFactory.ARTIFACT_SIGNER_ID,
```
slawekjaranowski, line 57: This property `ARTIFACT_SIGNER_ID` is only used in this place .... why we add it?

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultDeployer.java
```diff
@@ -151,7 +162,23 @@ private DeployResult deploy(SyncContext syncContext, RepositorySystemSession ses
             throw new DeploymentException("Failed to deploy artifacts/metadata: " + e.getMessage(), e);
         }
 
+        List<Artifact> artifacts = new ArrayList<>(request.getArtifacts());
+        List<? extends ArtifactGenerator> artifactGenerators = getArtifactGenerators(session, request);
         try {
+            List<Artifact> generatedArtifacts = new ArrayList<>();
+            for (ArtifactGenerator artifactGenerator : artifactGenerators) {
+                Collection<? extends Artifact> generated = artifactGenerator.generate(generatedArtifacts);
```
slawekjaranowski, line 170: `generatedArtifacts` is an empty list at beginning - so how we discover artifacts to sign in `artifactGenerator.generate`

maven-resolver-demos/maven-resolver-demo-snippets/pom.xml
```diff
@@ -34,8 +34,8 @@
 
   <properties>
     <Automatic-Module-Name>org.apache.maven.resolver.demo.snippets</Automatic-Module-Name>
-    <!-- To make Jetty work -->
-    <javaVersion>11</javaVersion>
+    <!-- To make Jetty + generator-signer work -->
+    <javaVersion>17</javaVersion>
```
slawekjaranowski, line 38: I only see that `UnixDomainSocketAddress` used in `GpgAgentPasswordLoader` requres JDK 16+ 
so I don't see connections with Jetty

maven-resolver-generator-signer/src/main/java/org/eclipse/aether/generator/signer/gpg/GpgSignerFactory.java
```diff
@@ -0,0 +1,206 @@
...
+import static org.eclipse.aether.generator.signer.gpg.GpgConfigurationKeys.CONFIG_PROP_KEY_ID;
+
+/**
+ * GnuPG signer factory implementation, that is active only for {@link RemoteRepository}ies.
+ */
+@Singleton
+@Named(GpgSignerFactory.NAME)
+public final class GpgSignerFactory implements SignerFactory {
+    public static final String NAME = GpgConfigurationKeys.NAME;
+    private static final String SIGNER_KEY = GpgSignerFactory.class.getName() + ".signer";
```
slawekjaranowski, line 63: unused constant

maven-resolver-generator-signer/src/main/java/org/eclipse/aether/generator/signer/gpg/GpgSignerFactory.java
```diff
@@ -0,0 +1,206 @@
...
+            if (keyRingMaterial != null) {
+                break;
+            }
+        }
+        if (keyRingMaterial == null) {
+            throw new IllegalArgumentException("Key ring material not found");
+        }
+
+        Long keyId = null;
+        String keyIdStr = ConfigUtils.getString(session, null, CONFIG_PROP_KEY_ID);
```
slawekjaranowski, line 134: `keyId` should also be provided by environment value

slawekjaranowski (commented):
There are many new Factories ... do we need all of them?

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultDeployer.java
slawekjaranowski: ok - I see we heve materialized artifact list in `org.eclipse.aether.generator.signer.SignerArtifactGenerator`
It is look too complicated

maven-resolver-demos/maven-resolver-demo-snippets/pom.xml
cstamas: Originally it was 11 as Jetty 10.x we use as client and as server (in http-test) required 11, but now upped to 17.
Will change it: "jetty needs 11, generator-signer needs 17"

cstamas: fixed

maven-resolver-generator-signer/src/main/java/org/eclipse/aether/generator/signer/gpg/GpgSignerFactory.java
cstamas: fixed

maven-resolver-generator-signer/src/main/java/org/eclipse/aether/generator/signer/gpg/GpgSignerFactory.java
cstamas: fixed, now only exists "loaders" (3 of them: env, conf and agent)

cstamas:
@slawekjaranowski general remark:
* spi defined ArtifactGenerators
* generator-signer hooks into this, and define Signers
* gpg is one Signer implementation

But also:
* there may be multiple generators, see how they are ordered, and the idea is that "signer generator" is last, and it should sign all the deployed but also "so far generated" artifacts.
* gpg may not be the only one signer either
* moreover, a signer may sign differently than gpp. Gpg signs each file, but another signer may provide one signature for all "relevant" artifacts.

maven-resolver-generator-signer/src/main/java/org/eclipse/aether/generator/signer/gpg/GpgSignerFactory.java
```diff
@@ -0,0 +1,215 @@
...
+
+        /**
+         * Returns the key password, or {@code null}.
+         */
+        default char[] loadPassword(RepositorySystemSession session, long keyId) throws IOException {
+            return null;
+        }
+    }
+
+    private final RepositoryLayoutProvider repositoryLayoutProvider;
```
cstamas, line 92: This is wrong: signer have nothing to do with remote repository layout. Am reworking this.

cstamas: fixed

cstamas:
@michael-o @slawekjaranowski @gnodet pls review again.

Reworked, as layout was "improper" for signer (is transport and remote repository related). What was _actually missing_ was the ArtifactPredicate (that recognizes artifact hashes and also configured extensions without hashes). This config was never per-remote repository, so it felt right to be moved to new component. Maven2 repo layout updated accordingly. Finally, this change revealed a (doc) bug: remote repo checksums ARE configured per remote repository, but doco did not reflect that.

cstamas:
So to recap, what I tried to explain to Slawek. This PR:
* defines `Artifact Generator` SPI (there may be multiple of those present in system)
* defines `Signer Artifact Generator` that is one provider for SPI, that defines Signer API (there may be multiple of those present in system)
* implements GpgSigner that is one Signer implementation provided out of the box

Naturally, as generators are ordered, signer generator is expected to be "among last" and sign not only artifacts in request (install/deploy) but also "so far generated artifacts" (if they are relevant).

gnodet:
> So to recap, what I tried to explain to Slawek. This PR:
> 
> * defines `Artifact Generator` SPI (there may be multiple of those present in system)
> * defines `Signer Artifact Generator` that is one provider for SPI, that defines Signer API (there may be multiple of those present in system)

Now that you put it that way, I'm not really convinced we need a Signer SPI here. The code of the ArtifactGenerator which calls the signers is trivial, so it seems to me that the GpgSigner could directly implement the Artifact Generator SPI.

> * implements GpgSigner that is one Signer implementation provided out of the box
> 
> Naturally, as generators are ordered, signer generator is expected to be "among last" and sign not only artifacts in request (install/deploy) but also "so far generated artifacts" (if they are relevant).

cstamas:
> Now that you put it that way, I'm not really convinced we need a Signer SPI here. The code of the ArtifactGenerator which calls the signers is trivial, so it seems to me that the GpgSigner could directly implement the Artifact Generator SPI.

Agreed, will simplify it more, but the question begs: "generator-signer" module becomes wrong, no? Should it be maybe "generator-gnupg (or gpg)" instead?

gnodet:
> > Now that you put it that way, I'm not really convinced we need a Signer SPI here. The code of the ArtifactGenerator which calls the signers is trivial, so it seems to me that the GpgSigner could directly implement the Artifact Generator SPI.
> 
> Agreed, will simplify it more, but the question begs: "generator-signer" module name becomes wrong, no? Should it be maybe "generator-gnupg (or gpg)" instead?

Yes, I think that would make more sense.

cstamas:
Collapsed code and renamed to generator-gnupg

gnodet (approved):

jira-importer:
Resolve #978

PR #445, opened 2024-03-15
[MRESOLVER-512] ScopeManager

cstamas:
Resolver ScopeManager.

---

https://issues.apache.org/jira/browse/MRESOLVER-512


gnodet:
I'm *really* reluctant about this PR.  This goes in the exact opposite direction than https://issues.apache.org/jira/browse/MRESOLVER-471, which aimed to make the resolver agnostic about scopes and leave these bits to be defined by Maven.

It basically replaces 250 lines of deprecated code (the `JavaDependencyContextRefiner`, `JavaScopeDeriver` and `JavaScopeSelector` which were deprecated a few weeks ago) with 3800 new lines of code, adds a strong tie between the resolver and maven, whereas the code could be easily moved into maven which already exposes those concepts.

cstamas:
Well, IMO this PR goes exactly in direction of MRESOLVER-471, as:
* scope manager is optional, in a sense, if you want to use Resolver 2.x as you did Resolver 1.x, go for it
* ScopeManager solves the problem of "Resolver should be oblivious about scopes" (MRESOLVER-471), as Resolver 1.x implemented scope related logic scattered across several (dozen?) classes, and actions like "introducing new scope" or "removal of a scope" was impossible exactly due that (as Resolver was not oblivious, but had them cemented as String constants spread across dozen classes). Hence Resolver was "this xor that". 

With ScopeManager Resolver IS really oblivious, as they are defined by client project, the app that is integrating Resolver 2.

gnodet:
@cstamas The whole `SystemScopeHandler` needs has to be removed imho, it's superseded by the `ScopeManager`.

cstamas:
@gnodet applied changes, IMHO it looks even better now :smile:

gnodet (approved):

PR #450, opened 2024-03-30
[MRESOLVER-518] Version selector improvements

cstamas:
Version selector improvements:
* introduce strategy how to select "winner" (so far it was cemented "nearest") and optionally perform some enforcements
* supplier retains same behaviour as before

Big fat note: this is NOT to alter Maven4 behaviour, but to make Resolver 2 more versatile for example for "diagnostic purposes" (while Maven may still benefit from these changes as well, by having some "strict version strategy" mode for example).

---

https://issues.apache.org/jira/browse/MRESOLVER-518


maven-resolver-api/src/main/java/org/eclipse/aether/collection/UnsolvableVersionConflictException.java
```diff
@@ -42,9 +42,25 @@ public class UnsolvableVersionConflictException extends RepositoryException {
      * Creates a new exception with the specified paths to conflicting nodes in the dependency graph.
      *
      * @param paths The paths to the dependency nodes that participate in the version conflict, may be {@code null}.
+     * @deprecated Use {@link #UnsolvableVersionConflictException(String, Collection)} instead.
```
gnodet, line 45: I don't see the point of deprecating this constructor, since it forces the addition of a `null` parameter and lead the user to a change which is useless, given the message has a default value.

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/transformer/NearestVersionSelector.java
```diff
@@ -126,7 +129,7 @@ private UnsolvableVersionConflictException newFailure(final ConflictContext cont
         };
         PathRecordingDependencyVisitor visitor = new PathRecordingDependencyVisitor(filter);
         context.getRoot().accept(new TreeDependencyVisitor(visitor));
-        return new UnsolvableVersionConflictException(visitor.getPaths());
+        return new UnsolvableVersionConflictException(null, visitor.getPaths());
```
gnodet, line 132: No need for this change imho

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/transformer/ConfigurableVersionSelector.java
```diff
@@ -0,0 +1,350 @@
...
+         * This method should determine are versions of two items "compatible" or not. This method is invoked whenever
+         * {@code candidate} is "considered" (does not have to be selected as "winner").
+         */
+        boolean isIncompatibleVersion(ConflictItem candidate, ConflictItem winner)
+                throws UnsolvableVersionConflictException;
+    }
+    /**
+     * If true, this version selector will fail if detects "dependency version divergence" in graph.
+     */
+    protected final boolean enforceVersionConvergence;
```
gnodet, line 79: Should that be a specific selection strategy instead ?
We can't really enforce a single version and select the higher (or nearer) at the same time, right ?

maven-resolver-api/src/main/java/org/eclipse/aether/collection/UnsolvableVersionConflictException.java
cstamas: No existing code uses this deprecated ctor (not counting deprecated class), or new one with `null` param, it is left merely for binary compatibility. In fact, this exception is so special, that nothing else uses it.

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/transformer/NearestVersionSelector.java
cstamas: undone

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/transformer/ConfigurableVersionSelector.java
cstamas: Agreed, is a strategy now.

cstamas:
ping, any other comments to this PR?

rmannibucau (approved):

jira-importer:
Resolve #1185

PR #452, opened 2024-04-01
[MRESOLVER-519] Drop almost equal/same method copies

cstamas:
And introduce one in parent class.

---

https://issues.apache.org/jira/browse/MRESOLVER-519


slawekjaranowski (approved):

jira-importer:
Resolve #1181

PR #455, opened 2024-04-02
[MRESOLVER-521] Improve congested file locks behaviour

cstamas:
Instead to immediately give up and sleep, they will sit a while to enter critical region. This is important for "hot" locks.

Explanation: currently a "loser" will _immediately give up_ and will go to sleep 100ms if cannot enter critical region, even if winner exits critical region within next 5ms. The retry is needed to ensure that it is retried as much as given time/unit takes, that was before consumed by constant retries+sleeps. The logic still works, as if tryLock spends time/unit waiting on criticalRegion (which is possible only on VERY HIGHLY congested locks), there will be no retry happening.

---

https://issues.apache.org/jira/browse/MRESOLVER-521


jira-importer:
Resolve #1209

PR #461, opened 2024-04-10
[1.9.x] [MRESOLVER-522] Improve congested file locks behaviour (#455)

cstamas:
Instead to immediately give up and sleep, they will sit a while to enter critical region. This is important for "hot" locks.

Explanation: currently a "loser" will _immediately give up_ and will go to sleep 100ms if cannot enter critical region, even if winner exits critical region within next 5ms. The retry is needed to ensure that it is retried as much as given time/unit takes, that was before consumed by constant retries+sleeps. The logic still works, as if tryLock spends time/unit waiting on criticalRegion (which is possible only on VERY HIGHLY congested locks), there will be no retry happening.

Backport of 089796ba66d6f5f1a9d4fa7f797f613bae4f5373

---

https://issues.apache.org/jira/browse/MRESOLVER-522


jira-importer:
Resolve #1195

PR #464, opened 2024-04-11
[MRESOLVER-535][MRESOLVER-538] Add decorator ability to graph dumper; show ranges

cstamas:
Ability to pass in any function that is able to "decorate" the print out of the graph. This should have a follow up Pr that enable/disable (or simply move all into decorators). Also, verbose tree should show use of version ranges, for cases when someone wants to detect them in whole transitive hull.

---

https://issues.apache.org/jira/browse/MRESOLVER-535
https://issues.apache.org/jira/browse/MRESOLVER-538


jira-importer:
Resolve #1226

PR #473, opened 2024-04-17
[MRESOLVER-539] Use parent POM 42

cstamas:
Use parent POM 42.

---

https://issues.apache.org/jira/browse/MRESOLVER-539


slawekjaranowski (approved):
dep management for `junit-bom` is in a parent 😄

jira-importer:
Resolve #1197

PR #474, opened 2024-04-18
[MRESOLVER-540] TransferResource can reveal more about resource

cstamas:
Provide extra information about the resource being transferred.

---

https://issues.apache.org/jira/browse/MRESOLVER-540

### DIFF
--- a/maven-resolver-api/src/main/java/org/eclipse/aether/transfer/TransferResource.java
+++ b/maven-resolver-api/src/main/java/org/eclipse/aether/transfer/TransferResource.java
@@ -32,6 +32,8 @@ public final class TransferResource {
 
     /**
      * The type of this resource.
+     *
+     * @since 2.0.0
      */
     public enum ResourceType {
         /**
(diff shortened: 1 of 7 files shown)


cstamas:
Main inspiration is to make stunts like these not required: https://github.com/apache/maven-mvnd/pull/930

maven-resolver-api/src/main/java/org/eclipse/aether/transfer/TransferResource.java
```diff
@@ -23,17 +23,43 @@
...
  * Describes a resource being uploaded or downloaded by the repository system.
  */
 public final class TransferResource {
 
+    /**
+     * The type of this resource.
+     *
+     * @since 2.0.0
+     */
+    public enum ResourceType {
```
gnodet, line 38: I wonder if that one is really needed, as one could use `instanceof` on the resource (and a cast will be needed anyway)...

gnodet (approved):
Not sure about the enum...

maven-resolver-api/src/main/java/org/eclipse/aether/transfer/TransferResource.java
cstamas: I kinda agree with you, moreover as these events are emitted ONLY by connector. For example, Transport API of Maven4 completely circumvents connector, and goes directly to transport... so yeah, due this (connector) type can really only be Artifact or Metadata

cstamas:
Will remove/simplify it

cstamas:
Simplified, removed cruft and is not simple and nice. @gnodet pls review

PR #479, opened 2024-04-25
[MRESOLVER-542] Reduce usage of final classes

cstamas:
This mostly affects utils. As APIs are final for a reason.

And some cleanup as well

---

https://issues.apache.org/jira/browse/MRESOLVER-542


gnodet (approved):

jira-importer:
Resolve #1217

PR #482, opened 2024-04-25
[MRESOLVER-548] Artifact file/path improvements

cstamas:
As original went from File to NIO2 Path, there were some changes left that may disturb code written agains Resolver 1.

The `AbstractArtifact` in resolver is extended in resolver in 3 classes. Still, there may be "outsider" extending it as well. Resolver classes all override properly the file/path related methods, but "outsiders" may not. It is wrong to have here "file -> path" redirection, as "outsider" may and is able only to override file related methods. Hence, `AbstractArtifact` should turn this around and make "path -> file" delegation instead, as 3rd party classes most probably overrode only file related methods.

---

https://issues.apache.org/jira/browse/MRESOLVER-548

### DIFF
--- a/maven-resolver-api/src/main/java/org/eclipse/aether/artifact/AbstractArtifact.java
+++ b/maven-resolver-api/src/main/java/org/eclipse/aether/artifact/AbstractArtifact.java
@@ -82,7 +82,7 @@ public abstract class AbstractArtifact implements Artifact {
      * @param path The resolved file of the artifact, may be {@code null}.
      * @return The new artifact instance, never {@code null}.
      */
-    private Artifact newInstance(String version, Map<String, String> properties, Path path) {
+    protected Artifact newInstance(String version, Map<String, String> properties, Path path) {
         return new DefaultArtifact(
                 getGroupId(), getArtifactId(), getClassifier(), getExtension(), version, path, properties);
     }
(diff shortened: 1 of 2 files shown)


maven-resolver-api/src/main/java/org/eclipse/aether/artifact/AbstractArtifact.java
```diff
@@ -96,30 +96,40 @@ public Artifact setVersion(String version) {
...
+     * this class is extended somewhere.
+     */
     @Override
     public Artifact setPath(Path path) {
         Path current = getPath();
         if (Objects.equals(current, path)) {
             return this;
         }
-        return newInstance(getVersion(), getProperties(), path);
+        return setFile(path != null ? path.toFile() : null);
```
gnodet, line 129: That's wrong. It goes out the NIO api which means a virtual FileSystem cannot be used anymore.

cstamas: This exec path is used ONLY for "outsiders" (ie ProvisioArtifact), all the resolver internal classes override this method. For things like proviso to not break virtual FS will need to be rewritten, as all they know about is File only.

cstamas:
Not doing this

jira-importer:
Resolve #1232

PR #483, opened 2024-04-25
[MRESOLVER-547] Just use setVersion

cstamas:
No need for full copy, Artifact is already immutable. Moreover, the instance may be not DefaultArtifact but something else. And finally, setVersion already
have "optimization" to return this if version is
same as the one we want to copy with.

---

https://issues.apache.org/jira/browse/MRESOLVER-547


jira-importer:
Resolve #1233

PR #484, opened 2024-04-26
[MRESOLVER-549] Parent POM 42

cstamas:
Demos were using implicit sisu index creation that is now disabled in parent, so needs sisu plugin

---

https://issues.apache.org/jira/browse/MRESOLVER-549
https://issues.apache.org/jira/browse/MRESOLVER-550
https://issues.apache.org/jira/browse/MRESOLVER-551


jira-importer:
Resolve #1219

PR #503, opened 2024-06-13
[MRESOLVER-565] Class renames

cstamas:
Rename classes to reflect what they are. These are "default implementations" used on new Session instances but usually not used in "real world".

---

https://issues.apache.org/jira/browse/MRESOLVER-565


michael-o (approved):
This also obsoletes my PR.

jira-importer:
Resolve #1243

PR #505, opened 2024-06-19
[1.9.x] [MRESOLVER-570] Remove excessive strictness of OSGi dependency metadata

HannesWell:
- Mark optional Maven dependencies as optional in OSGi metadata
It is an explicit decision from bndtools to require explicit configuration of a user to make a dependency optional, even if the corresponding maven dependency is optional: https://github.com/bndtools/bnd/issues/2713
- Widen version range for org.slf4j.spi package to [1.7,3)

Fixes [MRESOLVER-570](https://issues.apache.org/jira/browse/MRESOLVER-570)

### DIFF
--- a/maven-resolver-impl/pom.xml
+++ b/maven-resolver-impl/pom.xml
@@ -31,6 +31,18 @@
   <name>Maven Artifact Resolver Implementation</name>
   <description>An implementation of the repository system.</description>
 
+  <properties>
+    <bnd.instructions.additions><![CDATA[
+      # Mark optional Maven dependencies as optional
+      Import-Package: \
+        org.slf4j.spi;version="[1.7,3)", \
+        javax.inject;resolution:=optional,\
+        com.google.inject*;resolution:=optional, \
+        org.eclipse.sisu*;resolution:=optional, \
+        *
+    ]]></bnd.instructions.additions>
+  </properties>
+
   <dependencies>
     <dependency>
       <groupId>org.apache.maven.resolver</groupId>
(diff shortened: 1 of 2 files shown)


cstamas:
please `mvn spotless:apply` and commit changes (formatting is enforced at build)

gnodet (approved):

gnodet:
@HannesWell I wonder if the intra-project imports should require a lower range, i.e. `[1.9,1.10)`.  The reason is that there's no real guarantee on the project's internals, so if someone starts mixing minor versions, it will probably fail.

gnodet:
> @HannesWell I wonder if the intra-project imports should require a lower range, i.e. `[1.9,1.10)`. The reason is that there's no real guarantee on the project's internals, so if someone starts mixing minor versions, it will probably fail.

Ah, just spotted that you came up to the exact same conclusion with #507

gnodet (approved):

HannesWell:
> > @HannesWell I wonder if the intra-project imports should require a lower range, i.e. `[1.9,1.10)`. The reason is that there's no real guarantee on the project's internals, so if someone starts mixing minor versions, it will probably fail.
> 
> Ah, just spotted that you came up to the exact same conclusion with #507

Yes.

Thanks for your review. I created a separate PR to just unify the build configuration regarding the bnd-maven-plugin and maven-jar-plugin via https://github.com/apache/maven-resolver/pull/509. If that is submitted before this one the noise due to moving things around could be reduced.

HannesWell:
This is now based on the changes made in https://github.com/apache/maven-resolver/pull/518 since it allows to implement this a bit more elegant.

HannesWell:

PR #506, opened 2024-06-19
[MRESOLVER-570] Remove excessive strictness of OSGi dependency metadata

HannesWell:
- Mark optional Maven dependencies as optional in OSGi metadata
- Widen version range for org.slf4j.spi package to [1.7,3)
- Move bnd configuration section from execution to plugin configuration in order to simplify extensions in sub-modules.

This is a forward port of https://github.com/apache/maven-resolver/pull/505 to resolver-2.x/master, that skips the update of the bnd-maven-plugin since it is already at 7.0.0.

Fixes [MRESOLVER-570](https://issues.apache.org/jira/browse/MRESOLVER-570)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -609,6 +609,10 @@
               Bundle-SymbolicName: ${Bundle-SymbolicName}
               # Export packages not containing the substring 'internal'
               -exportcontents: ${removeall;${packages};${packages;NAMED;*internal*}}
+              # Mark optional Maven dependencies as optional
+              Import-Package: \
+                javax.inject*;resolution:=optional, \
+                *
               # Reproducible build
               -noextraheaders: true
             ]]></bnd>


gnodet:
@HannesWell Same comment as https://github.com/apache/maven-resolver/pull/505#issuecomment-2180683597
I wonder if we should lower the import range to `[2.0,2.1)` for intra project imports... ?

gnodet (changes requested):
@HannesWell Same comment as https://github.com/apache/maven-resolver/pull/505#issuecomment-2180683597
I wonder if we should lower the import range to [2.0,2.1) for intra project imports... ?

HannesWell:
For resolver-2x this change is actually simpler than for 1.9.x since `maven-resolver-impl` does not reference Sisu, Guice and org.slf4j.spi in its code anymore. The classes `Slf4jLoggerFactory` and `AetherModule` have been dropped for resolver-2.0.

gnodet (approved):
LGTM, however, optional imports should be minimised when possible.

HannesWell:
Thank you for your review and submitting this.

> LGTM, however, optional imports should be minimised when possible.

In general yes. But most of the time problems with it arise if the code is not really prepared to handle the absence of a bundle, since this is not that easy. But since the Maven dependencies are also marked as optional and I assume the code is prepared to handle the absence correctly respectively the use-cases are designed accordingly I think it should be fine.

jira-importer:
Resolve #1238

PR #508, opened 2024-06-20
[MRESOLVER-27] turn api, spi and utils into Java 9 modules

rfscholte:
provided a `module-info` for the 3 mentioned modules. This i,plies, that the automatic module cannot be defined globally, because it is not allowed to have both a module descriptor and an automatic-module-name entry.

@cstamas not sure why the build fails, seems some java parser doesn't understand module descriptors.

### DIFF
--- a/maven-resolver-api/pom.xml
+++ b/maven-resolver-api/pom.xml
@@ -32,8 +32,7 @@
   <description>The application programming interface for the repository system.</description>
 
   <properties>
-    <Automatic-Module-Name>org.apache.maven.resolver</Automatic-Module-Name>
-    <Bundle-SymbolicName>${Automatic-Module-Name}.api</Bundle-SymbolicName>
+    <Bundle-SymbolicName>org.apache.maven.resolver.api</Bundle-SymbolicName>
   </properties>
 
   <dependencies>
@@ -58,6 +57,32 @@
       <plugin>
         <groupId>biz.aQute.bnd</groupId>
         <artifactId>bnd-maven-plugin</artifactId>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>default-compile</id>
+            <configuration>
+              <excludes>
+                <exclude>module-info.java</exclude>
+              </excludes>
+            </configuration>
+          </execution>
+          <execution>
+            <id>module-descriptor</id>
+            <goals>
+              <goal>compile</goal>
+            </goals>
+            <configuration>
+              <release>9</release>
+              <includes>
+                <include>module-info.java</include>
+              </includes>
+            </configuration>
+          </execution>
+        </executions>
       </plugin>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
--- a/maven-resolver-api/src/main/java/module-info.java
+++ b/maven-resolver-api/src/main/java/module-info.java
@@ -1,0 +1,32 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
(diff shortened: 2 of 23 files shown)


cstamas:
Yup, we generate now this page (as in 1.x it is fully out of sync and maintaining it is a hell):
https://maven.apache.org/resolver-archives/resolver-2.0.0-alpha-11/configuration.html

Using this code here, that I smashed just to get away with it... 
https://github.com/apache/maven-resolver/blob/master/maven-resolver-tools/src/main/java/org/eclipse/aether/tools/CollectConfiguration.java

cstamas:
Is executed here
https://github.com/apache/maven-resolver/blob/master/maven-resolver-tools/pom.xml#L85

Probably adding a simple file/path exclusion would do it. The code is really interested in Javadoc.

rfscholte:
awesome, that fixed it.

HannesWell:
You could maybe use the already configured `bnd-maven-plugin` to generate the `modules-info.java/class` based on the exported/imported package also configured in the OSGi MANIFEST.MF:

Maybe a simple `-jpms-module-info` is already sufficient, but configuring it is possible:

If you are interested I can work out a solution.

rfscholte:
I'll leave the preferred approach up to @cstamas . 
I just think that starting with 2.0.0 is a great opportunity to provide these module descriptors when possible. I already noticed that it will be an issue for several third party issues, but if there's a clear separation of libraries and applications, only applications should care about the implementation and libraries could start adopting these api/spi/utils.

rfscholte:
The latest merge broke this PR. And it seems like if we go for my proposal, I need to revert some stuf again. @HannesWell maybe you should give it a try.

HannesWell:
> I just think that starting with 2.0.0 is a great opportunity to provide these module descriptors when possible. 

Agree.

> The latest merge broke this PR. And it seems like if we go for my proposal, I need to revert some stuf again. @HannesWell maybe you should give it a try.

Yes, I expected that. But since #509 was against `resolver-1.9.x` and this is against `master` I'm not sure how that broke this exact PR? Or do you mean if a similar change is applied to the master or this is back-ported to `1.9.x`?
Anyways I can of course try the suggested approach after #509 has been forward-ported.

rfscholte:
Having a second look, we need to put back the maven-jar-plugin back anyway in case there's an automatic module name, that's something that cannot be defined globally

HannesWell:
> Yes, I expected that. But since #509 was against `resolver-1.9.x` and this is against `master` I'm not sure how that broke this exact PR? Or do you mean if a similar change is applied to the master or this is back-ported to `1.9.x`?

Never-mind I didn't saw https://github.com/apache/maven-resolver/pull/512.

> Having a second look, we need to put back the maven-jar-plugin back anyway in case there's an automatic module name, that's something that cannot be defined globally

I have to look into this first. Do you mind if I create an alternative PR using the suggested automatic generation based on the bnd-maven-plugin? Then we can see how both work?

cstamas:
@rfscholte Could you please pull out into separate PR the one liner "tool" change (that leaves our `/module-info.java`)? As that change is good on it's own, and the tool will not want to parse anything else than "plain" java source...

rfscholte:
@cstamas sure, https://github.com/apache/maven-resolver/pull/514

cstamas:
Thanks!

HannesWell:
> > Having a second look, we need to put back the maven-jar-plugin back anyway in case there's an automatic module name, that's something that cannot be defined globally
> 
> I have to look into this first. Do you mind if I create an alternative PR using the suggested automatic generation based on the bnd-maven-plugin? Then we can see how both work?

Please have a look at https://github.com/apache/maven-resolver/pull/516.

rfscholte:
Superceded by https://github.com/apache/maven-resolver/pull/516

jira-importer:
Resolve #820

jira-importer:
Resolve #820

PR #509, opened 2024-06-20
[Build] Unify bnd-maven-plugin and maven-jar-plugin invocations

HannesWell:
Keep the 'bnd' configuration for the `bnd-maven-plugin` at plugin level to simplify overwrites in sub-modules (avoid the need to re-specify the execution there).

Additionally update to `bnd-maven-plugin` 6.4.0.

If this would be applied before https://github.com/apache/maven-resolver/pull/505, it would reduce the 'noise' in that commit due to necessary re-configuraiton.

@cstamas and @gnodet if you think this change is fine, I can also forward port it to master. I also left out a JIRA ticket for such non-functional 'clean-up', let me know if it is still necessary.

PR #510, opened 2024-06-20
[1.9.x] [MRESOLVER-572] Export internal packages in OSGi metadata as x-internal

HannesWell:
This makes maven-resolver-supplier usable as bundle in an OSGi runtime.
At the moment the bundle cannot resolve because the `internal.impl` packages it uses and imports in its OSGi metadata are not exported by `maven-resolver-impl`.

Just exporting internal packages fixes that and, although it is discouraged to use internal/impl packages, gives consumers that can
accept no API-compatibility guarantees the ability to access these packages.
Adding the directive `x-internal` is a convention that, at least when using Eclipse PDE, shows a corresponding warning at call-sides.

Fixes https://issues.apache.org/jira/browse/MRESOLVER-572

This PR is currently a draft since it contains as first commit the build unification of https://github.com/apache/maven-resolver/pull/509 and is built on top of that.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -475,7 +475,10 @@
               Bundle-SymbolicName: org.apache.${replacestring;${project.artifactId};-;.}
               Automatic-Module-Name: ${Bundle-SymbolicName}
               # Export packages not containing the substring 'internal'
-              -exportcontents: ${removeall;${packages};${packages;NAMED;*internal*}}
+              -exportcontents: \
+                *.impl.*;x-internal:=true, \
+                *.internal.*;x-internal:=true, \
+                *
               # Mark optional Maven dependencies as optional
               Import-Package: \
                 javax.inject*;resolution:=optional, \


gnodet:
Please rebase

pom.xml
```diff
@@ -473,7 +473,8 @@
             <bnd><![CDATA[
               Bundle-SymbolicName: ${Bundle-SymbolicName}
               # Export packages not containing the substring 'internal'
-              -exportcontents: ${removeall;${packages};${packages;NAMED;*internal*}}
+              -exportcontents: *.internal.*;x-internal:=true, \
```
HannesWell, line 476: Should `x-internal` also be added to impl packages?
[api-compatibility/package-level-contracts](https://maven.apache.org/resolver/api-compatibility.html#package-level-contracts) says both kind of packages are to be treated as 'non-public'?
```suggestion
              -exportcontents: *.internal.*;x-internal:=true, \
                *.impl.*;x-internal:=true, \
```

HannesWell (commented):
> Please rebase

Done.

pom.xml
gnodet: Yes I think so.

HannesWell:

HannesWell:
Are snapshots deployed for the `maven-resolver-1.9.x` branch too?
I see master is deployed to https://repository.apache.org/content/repositories/snapshots, but I don't see it happening for `maven-resolver-1.9.x`

gnodet:
> Are snapshots deployed for the `maven-resolver-1.9.x` branch too? I see master is deployed to https://repository.apache.org/content/repositories/snapshots, but I don't see it happening for `maven-resolver-1.9.x`

I've deployed one manually.

jira-importer:
Resolve #1236

PR #516, opened 2024-06-21
[MRESOLVER-27] Turn resolver api, spi and utils into Java 9 modules with generated module-info.class

HannesWell:
This PR has the same intend as https://github.com/apache/maven-resolver/pull/508, but uses the `bnd-maven-plugin` to generate the `module-info.class` automatically using the `-jpms-module-info` instruction:

The differences to #508 are that the source `module-info.java` is not generated and that the `module-info.class` is always placed in the jar root. But since JRE's below Java-9 don't recognize that file, I hope that's not a problem.

### DIFF
--- a/maven-resolver-api/pom.xml
+++ b/maven-resolver-api/pom.xml
@@ -33,7 +33,8 @@
 
   <properties>
     <bnd.instructions.additions><![CDATA[
-      Automatic-Module-Name: org.apache.maven.resolver
+      -jpms-module-info: org.apache.maven.resolver
+      -removeheaders Automatic-Module-Name
     ]]></bnd.instructions.additions>
   </properties>
 
--- a/maven-resolver-spi/pom.xml
+++ b/maven-resolver-spi/pom.xml
@@ -31,6 +31,13 @@
   <name>Maven Artifact Resolver SPI</name>
   <description>The service provider interface for repository system implementations and repository connectors.</description>
 
+  <properties>
+    <bnd.instructions.additions><![CDATA[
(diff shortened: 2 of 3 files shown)


HannesWell:
Btw. this is intentionally split into two commits, where the first one only changes the configuration to define all MANIFEST.MF entries through the `bnd-maven-plugin`.

rfscholte:
> But since JRE's below Java-9 don't recognize that file, I hope that's not a problem.

That's not a problem. This spec is making use of the fact that `-` is not allowed in a className, so these files will be ignored.

rfscholte:
@HannesWell I guess by default all packages are exported, right? For these 3 that's fine, for all others we need to verify every package if it should be exported.
I've checked a couple of jars, results look as expected.

rfscholte (approved):

HannesWell:
> @HannesWell I guess by default all packages are exported, right? For these 3 that's fine, for all others we need to verify every package if it should be exported. I've checked a couple of jars, results look as expected.

Thanks for checking/reviewing.
That's right, but with using this approach the JPMS module-info and the OSGi metadata in the MANIFEST.MF export the same packages. So for each module the exported packages can be configured by adjusting the configuration of the bnd-maven-plugin.

FYI this is currently based on (an older state of) https://github.com/apache/maven-resolver/pull/517, which should probably be submitted after https://github.com/apache/maven-resolver/pull/513. Then this is basically a three-line change.

HannesWell:
Now that #517 is submitted, all prerequisites for this from my POV are done.

PR #517, opened 2024-06-21
[Build] Define all MANIFEST.MF entries through the bnd-maven-plugin

HannesWell:
and derive 'Bundle-SymbolicName' header from the artifactId where possible.

This is originally created for https://github.com/apache/maven-resolver/pull/516, but would probably also make https://github.com/apache/maven-resolver/pull/506 simpler.

### DIFF
--- a/maven-resolver-supplier-mvn4/pom.xml
+++ b/maven-resolver-supplier-mvn4/pom.xml
@@ -33,9 +33,9 @@
 
   <properties>
     <javaVersion>17</javaVersion>
-
-    <Automatic-Module-Name>org.apache.maven.resolver.supplier</Automatic-Module-Name>
-    <Bundle-SymbolicName>${Automatic-Module-Name}</Bundle-SymbolicName>
+    <bnd.instructions.additions><![CDATA[
+      Bundle-SymbolicName: org.apache.maven.resolver.supplier
+    ]]></bnd.instructions.additions>
   </properties>
 
   <dependencies>
--- a/maven-resolver-transport-jdk-parent/maven-resolver-transport-jdk/pom.xml
+++ b/maven-resolver-transport-jdk-parent/maven-resolver-transport-jdk/pom.xml
@@ -33,9 +33,9 @@
   <description>Maven Artifact Transport JDK Java 11+.</description>
 
   <properties>
-    <Automatic-Module-Name>org.apache.maven.resolver.transport.jdk</Automatic-Module-Name>
-    <Bundle-SymbolicName>${Automatic-Module-Name}</Bundle-SymbolicName>
-
+    <bnd.instructions.additions><![CDATA[
+      Multi-Release: true
+    ]]></bnd.instructions.additions>
     <javaVersion>11</javaVersion>
   </properties>
 
@@ -148,18 +148,6 @@
             </configuration>
           </execution>
         </executions>
-      </plugin>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-jar-plugin</artifactId>
-        <configuration>
-          <archive>
-            <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
-            <manifestEntries>
-              <Multi-Release>true</Multi-Release>
-            </manifestEntries>
-          </archive>
-        </configuration>
       </plugin>
     </plugins>
   </build>
--- a/maven-resolver-api/pom.xml
+++ b/maven-resolver-api/pom.xml
@@ -32,8 +32,9 @@
   <description>The application programming interface for the repository system.</description>
 
   <properties>
-    <Automatic-Module-Name>org.apache.maven.resolver</Automatic-Module-Name>
-    <Bundle-SymbolicName>${Automatic-Module-Name}.api</Bundle-SymbolicName>
+    <bnd.instructions.additions><![CDATA[
+      Automatic-Module-Name: org.apache.maven.resolver
+    ]]></bnd.instructions.additions>
   </properties>
 
   <dependencies>
--- a/maven-resolver-connector-basic/pom.xml
+++ b/maven-resolver-connector-basic/pom.xml
@@ -31,11 +31,6 @@
   <name>Maven Artifact Resolver Connector Basic</name>
   <description>A repository connector implementation for repositories using URI-based layouts.</description>
 
-  <properties>
-    <Automatic-Module-Name>org.apache.maven.resolver.connector.basic</Automatic-Module-Name>
-    <Bundle-SymbolicName>${Automatic-Module-Name}</Bundle-SymbolicName>
-  </properties>
-
   <dependencies>
     <dependency>
       <groupId>org.apache.maven.resolver</groupId>
--- a/maven-resolver-generator-gnupg/pom.xml
+++ b/maven-resolver-generator-gnupg/pom.xml
@@ -32,9 +32,6 @@
   <description>A generator implementation for GnuPG signatures.</description>
 
   <properties>
-    <Automatic-Module-Name>org.apache.maven.resolver.generator.gnupg</Automatic-Module-Name>
-    <Bundle-SymbolicName>${Automatic-Module-Name}</Bundle-SymbolicName>
-
     <javaVersion>17</javaVersion>
     <bouncycastleVersion>1.78.1</bouncycastleVersion>
   </properties>
--- a/maven-resolver-impl/pom.xml
+++ b/maven-resolver-impl/pom.xml
@@ -31,11 +31,6 @@
   <name>Maven Artifact Resolver Implementation</name>
   <description>An implementation of the repository system.</description>
 
-  <properties>
-    <Automatic-Module-Name>org.apache.maven.resolver.impl</Automatic-Module-Name>
-    <Bundle-SymbolicName>${Automatic-Module-Name}</Bundle-SymbolicName>
-  </properties>
-
   <dependencies>
     <dependency>
       <groupId>org.apache.maven.resolver</groupId>
--- a/maven-resolver-named-locks-hazelcast/pom.xml
+++ b/maven-resolver-named-locks-hazelcast/pom.xml
@@ -31,11 +31,6 @@
   <name>Maven Artifact Resolver Named Locks using Hazelcast</name>
   <description>A synchronization utility implementation using Hazelcast.</description>
 
-  <properties>
-    <Automatic-Module-Name>org.apache.maven.resolver.named.hazelcast</Automatic-Module-Name>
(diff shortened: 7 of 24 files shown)


maven-resolver-transport-jdk-parent/maven-resolver-transport-jdk/pom.xml
```diff
@@ -33,9 +33,10 @@
   <description>Maven Artifact Transport JDK Java 11+.</description>
 
   <properties>
-    <Automatic-Module-Name>org.apache.maven.resolver.transport.jdk</Automatic-Module-Name>
-    <Bundle-SymbolicName>${Automatic-Module-Name}</Bundle-SymbolicName>
-
+    <bnd.instructions.additions><![CDATA[
+      Bundle-SymbolicName: org.apache.maven.resolver.transport.jdk
+      Automatic-Module-Name: ${Bundle-SymbolicName}
+    ]]></bnd.instructions.additions>
```
HannesWell, line 39: Is it correct that from all the `maven-resolver-transport-jdk` modules only `maven-resolver-transport-jdk` is actually to be used and `jdk-8`, `jdk-11` and `jdk-21` are only there fore technical reasons to create multi-release jars?

cstamas: That's right.

HannesWell: Ok, thanks. Then I think it it not really necessary to create a Manifest for those specific bundles or at least it is not necessary to make sure the information is complete, isn't it?

HannesWell: I have no incorporated that and updated the resolver-transport-jdk modules.

HannesWell:
@gnodet or @cstamas could you please have a look at this?

HannesWell:
I have now further enhanced this and added the Automatic-Module-Name by default for all modules (like in the 1.9.x variant of this in #518).
In #516 that header is explicitly removed for all modules that get a JPMS module-info generated.

Due to the naming unification this PR changes the `Bundle-SymbolicName` and `Automatic-Module-Name` header value for the locks modules:
- `org.apache.maven.resolver.named` -> `org.apache.maven.resolver.named.locks`
- `org.apache.maven.resolver.named.redisson` -> `org.apache.maven.resolver.named.locks.redisson`
- `org.apache.maven.resolver.named.hazelcast` -> `org.apache.maven.resolver.named.locks.hazelcast`

I hope this is acceptable?

@gnodet and @cstamas from my POV this is ready and I think it would suit best to have this before my other changes.
Could you please approve the workflows and review this?

maven-resolver-supplier-mvn4/pom.xml
```diff
@@ -33,9 +33,9 @@
 
   <properties>
     <javaVersion>17</javaVersion>
-
-    <Automatic-Module-Name>org.apache.maven.resolver.supplier</Automatic-Module-Name>
-    <Bundle-SymbolicName>${Automatic-Module-Name}</Bundle-SymbolicName>
+    <bnd.instructions.additions><![CDATA[
+      Bundle-SymbolicName: org.apache.maven.resolver.supplier
```
gnodet, line 37: What's the behaviour since maven-resolver-supplier-mvn3 and maven-resolver-supplier-mvn4 end up with the same symbolic name / module name ? Isn't that a problem ?

cstamas: my "non-expert" opinion: IMO no, as only one can be used at one time, or in other words, I see no use case when you want both in the same container?

gnodet: Well, that's OSGi, so it's supposed to support different versions of the same packages/services to be deployed in the same container...

gnodet: @HannesWell is there any real OSGi use case ? like pax-url or something like that ? Or is that just to be able to use Maven inside Eclipse ?
More specifically, is there any use of OSGi services, or just the package wiring layer ?

HannesWell: When using both in the same OSGi runtime it could indeed become a challenge and will probably depend on the runtime configuration how multiple bundles with same ID and version are handled.
If they had different version, e.g. aligned with the targeted maven version it wouldn't be a problem, but I don't think that's what you want and also what's suitable here.

Solutions could be to say that one should only use one of both or we add a corresponding name suffix like for the maven `artifactId`?
I can implement it as you prefer.

HannesWell: But since both bundles also export the package `org.eclipse.aether.supplier` in the same version it will in general be a challenge to use the one for maven 3 and 4 together in one runtime, at least if Import-Package is used.

HannesWell: > @HannesWell is there any real OSGi use case ? like pax-url or something like that ? Or is that just to be able to use Maven inside Eclipse ? More specifically, is there any use of OSGi services, or just the package wiring layer ?

I don't know pax-url much and cannot speak for this. I don't think there is a usage of OSGi services (at least not that I know of at the moment). My primary goal for the moment is actually just to use the Maven-Resolver in an Eclipse application to fetch maven artifacts without the need to repack all its jars. That use case does not even use Maven directly, it just needs the dependencies of the resolver.

So at least for me only the package-wiring layer is of interest. But I cannot tell about the future or the use-cases of others.

> Well, that's OSGi, so it's supposed to support different versions of the same packages/services to be deployed in the same container...

That's right, but due to the split packages in the supplier-mvn3 and mvn4 bundles (and in maven-core/model) it's generally difficult use to use both together. So I would lean towards Tamas stand-point. Although it could be nice in the future to be able to use both, but that might be more work. Many Maven artifacts are difficult to use directly in OSGi runtimes, mainly due to their split packages.

That being said in OSGi it any ways recommended to use Import-Package over Require-Bundle. If that's considered, the Bundle-Symbolic name is not relevant any ways. But due to the mentioned split-package in the resolver I think it would be good to have some kind of distinction in the bundle. So adding a maven-version specific suffix could probably help.

If one wants to operate with Import-Package only, the only solution I see (besides only installing one variant) is to add an attribute to the Package exports that specifies the maven version.

gnodet: Do we really need those two maven-resolver-supplier as bundles ?  They require maven packages which are not exported as OSGi bundles afaik, so anyone wanting to use the resolver will have to package Maven in a bundle, and it will be the place to embed the two classes from those jars imho.

HannesWell: > Do we really need those two maven-resolver-supplier as bundles ? They require maven packages which are not exported as OSGi bundles afaik, so anyone wanting to use the resolver will have to package Maven in a bundle, and it will be the place to embed the two classes from those jars imho.

I assumed using the resolver-supplier is the best way to use the resolver in OSGi since using Eclipse Sisu probably will lead to more struggles:
https://maven.apache.org/resolver/third-party-integrations.html#maven-resolver-supplier

Therefore I need the supplier as bundle. But you are right it needs maven-core/model artifacts that are not bundles.
I use the Eclipse M2E Maven-Targets to generate the OSGi metadata automatically when resolving these artifacts in the Eclipse target-platform.
Of course it would be possible to do the same for the resolver-supplier.

HannesWell: But we could also discuss this in a dedicated issue or PR since this one is actually just about restructuring/simplifying the build not tho change its output.

gnodet (approved):

maven-resolver-supplier-mvn4/pom.xml
gnodet: Fwiw, the maven 4 based supplier is a bit off.  It should now looks like 
https://github.com/apache/maven/blob/master/maven-api-impl/src/test/java/org/apache/maven/internal/impl/standalone/RepositorySystemSupplier.java

HannesWell: > That said, I don't think it's a good idea to tie a resolver jar so closely with maven private implementation classes... It should really be the opposite and the supplier should be provided by maven.

I have to say that I just learned the resolver last week from the nice demos (I new it exists and what it does longer before but have not worked with it) so my opinions are probably not that qualified.
But yes I understood the supplier as 'standalone' integration of the resolver with maven (as you say in the doc, in general the resolver can work with any kind of back-end and Maven is just one, probably the most prominent and driving one).
From that POV it is indeed a back-reference in the dependency graph to have maven-resolver-supplier in maven-resolver.
In maven there is already the maven-resolver-provider, would it make sense to merge the maven-resolver-supplier into that?

HannesWell: > @HannesWell are you aware of https://github.com/maveniverse/mima ? As it's maybe what you really need, as Resolver supplier provides really only the "basic experience".

Not yet. Thanks for the hint. I'll have a closer look at it.

> Also, see here:
> https://issues.apache.org/jira/browse/MNG-8163?focusedCommentId=17856703&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17856703

Yes that's an interesting discussion and I follow it. I just couldn't contribute anything useful (yet).
But after thinking again about this discussion, the argument that the supplier uses resolver-impl also made me wonder if moving that part to maven is really the right step.
But from your comment in that issue I assume that this is a bigger discussion that might only be solved in resolver 3.0.

Btw. should we have another issue or just PR to have different names for the bundles for resolver-supper for maven 3 and 4? Or do you think we should keep it as it is now?

HannesWell:
Thank you @gnodet for your review and submitting this.

maven-resolver-supplier-mvn4/pom.xml
gnodet: > > @HannesWell are you aware of https://github.com/maveniverse/mima ? As it's maybe what you really need, as Resolver supplier provides really only the "basic experience".
> 
> Not yet. Thanks for the hint. I'll have a closer look at it.
> 
> > Also, see here:
> > https://issues.apache.org/jira/browse/MNG-8163?focusedCommentId=17856703&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17856703
> 
> Yes that's an interesting discussion and I follow it. I just couldn't contribute anything useful (yet). But after thinking again about this discussion, the argument that the supplier uses resolver-impl also made me wonder if moving that part to maven is really the right step. But from your comment in that issue I assume that this is a bigger discussion that might only be solved in resolver 3.0.
> 
> Btw. should we have another issue or just PR to have different names for the bundles for resolver-supper for maven 3 and 4? Or do you think we should keep it as it is now?

We also need to take into account the Maven 4 API which kinda hides the resolver API... it may be easier to integrate in the near future.

PR #518, opened 2024-06-21
[Build] Define all MANIFEST.MF entries through the bnd-maven-plugin

HannesWell:
Define all MANIFEST.MF entries through the bnd-maven-plugin and derive `Bundle-SymbolicName` header from the `artifactId`.

This changes the 'Bundle-SymbolicName' and 'Automatic-Module-Name' for the following modules:
- `maven-resolver-named-locks` from `org.apache.maven.resolver.named` to `org.apache.maven.resolver.named.locks`
- `maven-resolver-test-util` from `org.apache.maven.resolver.testutil` to `org.apache.maven.resolver.test.util`

This is effectively the back-port of https://github.com/apache/maven-resolver/pull/517 to the `maven-resolver-1.9.x` branch.
The main difference is that unlike in #517, this change does not 'prepare' the introduction of JPMS `module-info.java/class` files.

### DIFF
--- a/maven-resolver-api/pom.xml
+++ b/maven-resolver-api/pom.xml
@@ -32,8 +32,9 @@
   <description>The application programming interface for the repository system.</description>
 
   <properties>
-    <Automatic-Module-Name>org.apache.maven.resolver</Automatic-Module-Name>
-    <Bundle-SymbolicName>${Automatic-Module-Name}.api</Bundle-SymbolicName>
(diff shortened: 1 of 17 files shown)


HannesWell:
@gnodet or @cstamas could you please have a look at this as well?

HannesWell:
@gnodet now that https://github.com/apache/maven-resolver/pull/517 is submitted for master, can we have this for resolver-1.9.x?

PR #523, opened 2024-07-01
[MRESOLVER-574] Use standard cookie spec

cstamas:
As default one is not standard (w00t?)

---

https://issues.apache.org/jira/browse/MRESOLVER-574


michael-o (approved):

jira-importer:
Resolve #1247

PR #524, opened 2024-07-01
[1.9.x] Tame bnd plugin warnings

cstamas:
To keep things simple, we config the plugin in
one place, but in some modules this config emits
warnings. This change just makes it go away.

No other source or any alike change, no issue either.


gnodet (approved):
LGTM

PR #525, opened 2024-07-01
Tame bnd plugin warnings

cstamas:
To keep things simple, we config the plugin in
one place, but in some modules this config emits
warnings. This change just makes it go away.

No other source or any alike change, no issue either.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -624,6 +624,10 @@
               # Reproducible build
               -noextraheaders: true
               ${bnd.instructions.additions}
+              # Remove warnings (as we keep things simple)
+              -fixupmessages: \\
+                'Unused Import-Package instructions';is:=ignore,\\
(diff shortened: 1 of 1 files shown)


cstamas:
On master another warning in util and spi:
```
[WARNING] /home/cstamas/Worx/apache-maven/maven-resolver/maven-resolver-util/pom.xml [17:74]: No value specified for key: -jpms-module-info. An empty value should be specified as '-jpms-module-info:' or '-jpms-module-info=': <<              -jpms-module-info
      -removeheaders Automatic-Module-Name>>
```

gnodet (approved):
I've slightly updated the instructions

HannesWell:
> On master another warning in util and spi:

Had a quick look, but according to the doc the module name is looked up if not specified:

So it might be a bug, that the validation in bnd does not handle that case.

PR #526, opened 2024-07-02
Use Google JIMFS for demos

cstamas:
This PR showcases how Maven4+Resolver2 can now run on custom (even in memory like JIMFS is) FileSystem


gnodet (approved):
Nice !

PR #530, opened 2024-07-09
[MRESOLVER-579] The TODO part

cstamas:
Adds tests, as with new test:
* JDK transport fails as user explained
* Jetty behaves
* Apache is not HTTP2 capable

---

https://issues.apache.org/jira/browse/MRESOLVER-579


cstamas:
Note: this PR contains the fix as well and will pass OK

PR #532, opened 2024-07-12
[MRESOLVER-584] JDK transport HTTP/2 GOAWAY improvement

cstamas:
Changes:
* drop 21 closer, it abruptly uses "shutdown now"
* in 11 make httpClient NOT stored per session, but per transport instance
* simplify and refactor, put things in places like insecure mode config

---

https://issues.apache.org/jira/browse/MRESOLVER-584


cstamas:
I have a feeling that "swapping" will not even happen. What _actually happened_ in this PR is that HttpClient go _much shorter lifespan_. Originally, HttpClient was kept and reused (for same session+repository) across whole session. Now, the client shares lifecycle with Transport instance, that is OTOH created by connector per invocation.

cstamas:
This fix is wrong.

Tested using reproducer https://issues.apache.org/jira/browse/MNG-8145 (reused same config and all).
Maven 4.0.0-beta-3 fails (reproducer reproduces). And this PR built JDK transport works OK.

BUT, the reason why it works is due this PR "shortening" the lifespan of JDK HttpClient, and not due "client swap".

When I set nginx config from original `keepalive_requests 500;` to `keepalive_requests 100;` (or even 50), this PR fails in very same way.

Still, the PR is correct in a way it keeps HttpClient per Transporter instance, and that is _correct change_.

Will see in some other solution...

cstamas:
Am removing "swap" bit, and will make this pr same as #533

PR #533, opened 2024-07-12
[MRESOLVER-584] Jetty transport HTTP/2 GOAWAY improvement

cstamas:
Just shorten the client lifespan to connector transaction duration vs as before, for whole session duration. Also simplify the code.

---

https://issues.apache.org/jira/browse/MRESOLVER-584


cstamas:
@scholzi100 am about to merge this one, as I changed JIRA goal also from "fix" to "improve", which this PR does. And later we can refine things even more.

jira-importer:
Resolve #1256

PR #534, opened 2024-07-13
[MRESOLVER-586] Avoid copying the type default properties map

gsmet:
When we don't have specific properties for a DefaultArtifact, we can just reuse the type default properties map as it's readonly and immutable.
The patch contains some additional changes as I renamed the method and parameters to make them more specific so that the method wouldn't be used for things that wouldn't be safe.


gsmet:
Backport for branch 1.9 is here: https://github.com/apache/maven-resolver/pull/535

PR #536, opened 2024-07-13
[MRESOLVER-587] Memory usage improvements

cstamas:
Do not use `Object` as descriptor key, use dedicated type. And intern the `List<Dependency>` on artifact descriptors. Also introduce "intern"-ing of ArtifactDescriptor dependencies and managedDependencies that is configurable (speed vs memory).

---

https://issues.apache.org/jira/browse/MRESOLVER-587

### DIFF
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/DataPool.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/DataPool.java
@@ -119,7 +119,7 @@ public final class DataPool {
         internArtifactDescriptorDependencies = ConfigUtils.getBoolean(
                 session, false, CONFIG_PROP_COLLECTOR_POOL_INTERN_ARTIFACT_DESCRIPTOR_DEPENDENCIES);
         internArtifactDescriptorManagedDependencies = ConfigUtils.getBoolean(
-                session, false, CONFIG_PROP_COLLECTOR_POOL_INTERN_ARTIFACT_DESCRIPTOR_MANAGED_DEPENDENCIES);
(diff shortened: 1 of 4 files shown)


cstamas:
This PR causes noticeable slow down: building Quarkus (mvn clean install -Dquickly) went from 9:20 to 11:10.

Ideas:
* make interning of dependencies configurable, default NO
* make interning of dependencyManagement configurable, default YES
* maybe intern only over some size limit?

The most offender is huge count of depMgt inherited from parent POM, so this above would make sensible defaults.

cstamas:
@gsmet this PR now has it all merged changes, plus, introduces config (defaults to "as before" -- false both) to intern or not intern the artifact descriptor deps list and managedDeps lists.

Am tinkering, that maybe feasible default would be false/true (do not intern deps, do intern managedDeps)?

gsmet:
> Am tinkering, that maybe feasible default would be false/true (do not intern deps, do intern managedDeps)?

It looks like a good compromise from my side.

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/DataPool.java
```diff
@@ -257,7 +257,7 @@ public static final class DescriptorKey {
 
         private DescriptorKey(Artifact artifact) {
             this.artifact = artifact;
-            this.hashCode = buildHashCode();
+            this.hashCode = Objects.hashCode(artifact);
```
gsmet, line 260: Ah yeah sorry, I was used to more complex hashCode construction and kept this pattern (we avoid using `Objects.hash` as it allocates an array and can be problematic in some cases so we end up with traditional boilerplate hashCodes).

PR #537, opened 2024-07-13
[MRESOLVER-587] Memory usage improvements

cstamas:
Do not use `Object` as descriptor key, use dedicated type. And intern the `List<Dependency>` on artifact descriptors. Also introduce "intern"-ing of ArtifactDescriptor dependencies and managedDependencies that is configurable (speed vs memory).

---

https://issues.apache.org/jira/browse/MRESOLVER-587


jira-importer:
Resolve #1258

PR #553, opened 2024-08-08
[MRESOLVER-541] S3 Transport using MinIO

cstamas:
This is a trivially simple S3 Transport with basic features, but allows Resolver transport with S3 protocol.

---

https://issues.apache.org/jira/browse/MRESOLVER-541


jira-importer:
Resolve #1202

PR #573, opened 2024-08-29
[MRESOLVER-599] Sigstore Artifact Generator

cstamas:
This is similar to Gpg Artifact Generator, signs installed/deployed artifacts as "batch" at Resolver level. This, combined with "deployAtEnd" means all deployed artifacts are signed in one go, no need OIDC auth per module and 10 minute window may process more (as all artifacts are present when "deployAtend"). Still, the loop may need more, as if many artifacts being deployed, can the 10 minute window still be broken?

---

https://issues.apache.org/jira/browse/MRESOLVER-599


jira-importer:
Resolve #1270

PR #581, opened 2024-10-12
[MRESOLVER-602][MRESOLVER-603] Expose redirect config for apache transport

cstamas:
The Apache transport `maxRedirects` and `followRedirects` exposed. Also return some missing config doco. But all this exposed another issue with doco generator, so fixed that as well, that in turn, required more (mostly Javadoc) fixes.

And sorry for "overloading" this PR with unrelated (mostly side or Javadoc) related fixes, as doco generator fixes made me run site, that on other hand revealed that master site is not buildable :cry: 

---

https://issues.apache.org/jira/browse/MRESOLVER-602
https://issues.apache.org/jira/browse/MRESOLVER-603


jira-importer:
Resolve #1280

jira-importer:
Resolve #1280

jira-importer:
Resolve #1282

PR #590, opened 2024-10-24
Undo P-X raise

cstamas:
It does not matter, but I dislike to pull in alpha of Maven 4 (Resolver should not pull anything of
Maven only in demos).

---

Undo https://issues.apache.org/jira/browse/MRESOLVER-619


jira-importer:
Resolve #1291

PR #595, opened 2024-10-27
[MRESOLVER-614] Fix transitive dependency management

cstamas:
Undo original "clic-clac" solution introduced in first PR, instead, we need more fine-grained control of what and when is applied.

---

https://issues.apache.org/jira/browse/MRESOLVER-614

### DIFF
--- a/maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/manager/AbstractDependencyManager.java
+++ b/maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/manager/AbstractDependencyManager.java
@@ -249,7 +249,7 @@ public abstract class AbstractDependencyManager implements DependencyManager {
             }
 
             // optional is not managed by model builder
-            // if rule exists, apply it
+            // apply only rules coming from "higher" levels
             Holder<Boolean> optional = managedOptionals.get(key);
             if (optional != null && isApplicable(optional)) {
                 if (management == null) {
--- a/maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/manager/AbstractDependencyManager.java
+++ b/maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/manager/AbstractDependencyManager.java
@@ -47,11 +47,12 @@ import static java.util.Objects.requireNonNull;
  * en-route collected ones are carefully applied to proper descendants only, to not override work done by model
  * builder already.
  * <p>
- * Details: Model builder handles version and scope from own dependency management (think "effective POM"). On the other
- * hand it does not handle optional, system paths and exclusions (exclusions are additional information not effective
- * or applied in same POM). Hence, this implementation makes sure, that version and scope are not applied onto same
- * node that actually provided the rules. It achieves this goal by tracking "depth" for each collected rule and ignoring
- * rules coming from same depth as processed dependency node is.
+ * Details: Model builder handles version, scope from own dependency management (think "effective POM"). On the other
+ * hand it does not handle optional for example. System paths are aligned across whole graph, making sure there is
+ * same system path used by same dependency. Finally, exclusions (exclusions are additional information not effective
+ * or applied in same POM) are always applied. This implementation makes sure, that version and scope are not applied
+ * onto same node that actually provided the rules, to no override work that ModelBuilder did. It achieves this goal
+ * by tracking "depth" for each collected rule and ignoring rules coming from same depth as processed dependency node is.
  *
  * @since 2.0.0
  */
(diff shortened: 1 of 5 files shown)


maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/manager/AbstractDependencyManager.java
```diff
@@ -229,37 +231,46 @@ public DependencyManagement manageDependency(Dependency dependency) {
                 }
             }
 
+            // system scope paths always applied to have them aligned
+            // (same artifact > same path) in whole graph
```
gnodet, line 235: why isn't system scope applied locally by ModelBuilder ?

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/manager/AbstractDependencyManager.java
```diff
@@ -229,37 +231,46 @@ public DependencyManagement manageDependency(Dependency dependency) {
...
                             new HashMap<>(dependency.getArtifact().getProperties());
-                    systemDependencyScope.setSystemPath(properties, localPath);
+                    systemDependencyScope.setSystemPath(properties, localPath.getValue());
                     management.setProperties(properties);
                 }
             }
 
-            Boolean optional = managedOptionals.get(key);
+            // optional is not managed by model builder
+            // if rule exists, apply it
```
gnodet, line 252: why isn't `optional` managed by model builder ?
And why don't we consider the depth for managing optional ?

cstamas: I don't know, but model builder method for managing optional is empty. Unsure about reasons.
And we don't consider depth for optional due that above: as model builder is not managing it (so it applied even "onto itself").

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/manager/AbstractDependencyManager.java
cstamas: It cannot: model builder manages model inheritance ("vertical"). It has no idea about "parent" and "child" in graph ("horizontal") direction

cstamas:
```
                  parent-parent
                   |
parent           parent
|                  |
rootDependency -- level1Dependency -- level2Dependency -- ...
```

So this is what I meant by "horizontal" (dependency manager while graph is being built) and model builder (building each node eff model). They are orthogonal.

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/manager/AbstractDependencyManager.java
gnodet: > I don't know, but model builder method for managing optional is empty. Unsure about reasons. 

See https://issues.apache.org/jira/browse/MNG-4600 and https://issues.apache.org/jira/browse/MNG-5227

> And we don't consider depth for optional due that above: as model builder is not managing it (so it applied even "onto itself").

But then shouldn't the rule considered if `holder.getDepth() == 0 || depth >= holder.getDepth() + 1` ? Else a dependency may end up being flagged as optional when it should not ?

cstamas:
Related explanation(s): https://gist.github.com/cstamas/a65f042d14ee8e58c1a6d3c17e3fecaf

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/manager/AbstractDependencyManager.java
cstamas: MNG-4600: It introduced the empty method https://svn.apache.org/viewvc/maven/maven-3/trunk/maven-model-builder/src/main/java/org/apache/maven/model/management/DefaultDependencyManagementInjector.java?r1=925481&r2=925480&pathrev=925481 
MNG-5227: there is comment "is solved in resolver but not in ModelBuilder". This clearly says "handled by dependency manager but not model builder"

gnodet: > MNG-4600: It introduced the empty method https://svn.apache.org/viewvc/maven/maven-3/trunk/maven-model-builder/src/main/java/org/apache/maven/model/management/DefaultDependencyManagementInjector.java?r1=925481&r2=925480&pathrev=925481 MNG-5227: there is comment "is solved in resolver but not in ModelBuilder". This clearly says "handled by dependency manager but not model builder"

Are you saying that optional can be managed already ?

cstamas: No, am saying that model builder is NOT managing it, and resolver dependency manager WAS managing it (classic, and now transitive as well).

gnodet (approved):

PR #599, opened 2024-11-13
[MRESOLVER-627] Improve transport selection and logging

cstamas:
In previous releases transport selection was really secretive and broken, as it considered invalid configuration as "protocol not supported" and Resolver did not log anything or worse, silently skipped to new transport.

Changes:
* transport provider: was logging ONLY if transport selection failed (no transport was selected), changed to continuously log (in DEBUG) why a transport was ignored and next selection performed. Also, in case of total failure, exceptions are added now as suppressed ones instead just lost.
* multiple transporter factories: smaller fixes for earlier protocol detection (file is exception to support cases like JIMFS), and when client creation is attempted but it fails due runtime exception, fail the whole process (like bad enum in JDK transport config).
* before all transporter "blanket" threw NoTransporterEx and Basic connector was "protecting" caller from it, hence to continue doing so, no basic connector "blankets" all runtime and non-runtime exceptions coming from transport to "early report" problem as it happens in basic connector ctor.

---

https://issues.apache.org/jira/browse/MRESOLVER-627


cstamas:
> I don't have an opinion on this.

Sure, no problem. But let me explain what was initial cause for this PR:
We knew that JDK transport cannot handle HTTP/2 `GOAWAY`, and recently we saw a surge in ITs and other jobs getting it, so we decided to _force_ HTTP/1.1 while keeping JDK transport in use. I did that by creating `.mvn/maven.config` in maven and mvnd projects... then I went to check "effects" using `-X` locally and was **very surprised to see that Maven started using Apache transporter instead**.  

Basically, turned out my config had wrong value (a typo), and even `-X` did not report this, instead all I saw in debug logs is that Maven uses Apache transport.... turned out, this typo was interpreted as "not my protocol" by JDK transport, and it made Maven select next available HTTP Transporter that was Apache. Very confusing. 

Hence, this PR now:
* logs every transport "disqualification" (in DEBUG) instead only when no transport could be selected
* fails the build if any transport is mis-configured

jira-importer:
Resolve #1301

PR #600, opened 2024-11-14
[MRESOLVER-629] JDK transport should default to HTTP/1.1

cstamas:
As it cannot cope well with HTTP/2 GOAWAY frame

---

https://issues.apache.org/jira/browse/MRESOLVER-629


jira-importer:
Resolve #1298

PR #615, opened 2024-12-11
[MRESOLVER-635] Use Instant instead of untyped millis for TransferResource startTime

gnodet:
This also provides a clean for Maven to inject the `MonotonicClock` from https://github.com/apache/maven/pull/1965

---

https://issues.apache.org/jira/browse/MRESOLVER-635

### DIFF
--- a/maven-resolver-api/src/main/java/org/eclipse/aether/transfer/TransferResource.java
+++ b/maven-resolver-api/src/main/java/org/eclipse/aether/transfer/TransferResource.java
@@ -22,7 +22,6 @@ import java.io.File;
 import java.nio.file.Path;
 import java.time.Clock;
 import java.time.Instant;
-import java.time.temporal.Temporal;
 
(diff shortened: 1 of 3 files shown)


michael-o (approved):

maven-resolver-api/src/main/java/org/eclipse/aether/transfer/TransferResource.java
```diff
@@ -234,6 +235,15 @@ public TransferResource setResumeOffset(long resumeOffset) {
      * @return The timestamp when the transfer of this resource was started.
```
cstamas, line 235:
```suggestion
     * @return The timestamp when the transfer of this resource was started.
     * @deprecated
```

gnodet:
@cstamas added the deprecation notice, and provided a `Clock` singleton that will be set by Maven to provide monotonic `Instant` instances.

PR #640, opened 2025-01-26
[MRESOLVER-653] NPE in case of failed result

cstamas:
Obvious typo: result artifact may be null, it is request artifact that wanted to be checked.

---

https://issues.apache.org/jira/browse/MRESOLVER-653


jira-importer:
Resolve #1326

PR #641, opened 2025-01-26
[MRESOLVER-660] Make optional method really optional

cstamas:
Make optional method really optional.

---

https://issues.apache.org/jira/browse/MRESOLVER-660


jira-importer:
Resolve #1332

PR #649, opened 2025-02-01
[MRESOLVER-652] Support legacy 1.9.x split repository config

cstamas:
Added legacy properties for now as private and undocumented keys.

---

https://issues.apache.org/jira/browse/MRESOLVER-652


jira-importer:
Resolve #1329

PR #650, opened 2025-02-01
[MRESOLVER-651] Doco update re split repo

cstamas:
Just add explanation how properties are renamed, but not change needed, as MRESOLVER-652 adds backward compat support.

---
https://issues.apache.org/jira/browse/MRESOLVER-651


michael-o (approved):
Y!

jira-importer:
Resolve #1324

PR #652, opened 2025-02-10
ASCII digits only for versions

elharo:
Everything else is treated alphabetically.


cstamas:
Pls don't forget spec, change it as well if you do updates to implementation that maybe needs to go there as well:
https://github.com/apache/maven-resolver/blob/master/maven-resolver-util/src/main/java/org/eclipse/aether/util/version/package-info.java

cstamas:
@dmlloyd ping

dmlloyd:
As what one might refer to as a "frequent complainer" when it comes to the version scheme, I think this change seems reasonable and shouldn't negatively impact anything that I'm aware of.

elharo (commented):
Thanks. I was unaware of the package.info spec. I added these cases to it.

slachiewicz (approved):

PR #654, opened 2025-02-12
[MNG-8578]add exclusion scope to repo, allow exclude dependencies from build(including parent, dependencies, etc)

XenoAmess:


elharo (changes requested):
This must have docs, both a proposal explaining what you want to do and docs for the end user if it's merged.

elharo (changes requested):
Also requires tests, and those should be written before the model code is written, not after.

PR #655, opened 2025-02-16
Prune plexus dependencies

elharo:


elharo:
Failure looks unrelated and flaky: Server not initialized yet, please try again.

cstamas:
What does this PR buys to us?

elharo:
https://jlbp.dev/JLBP-1

elharo:
If something else depends on plexus-utils, it should declare the dependency on plexus-utils. We wouldn't add it if it weren't already here, so we should take it out. Unless there's some weird dependency between plexus-xml and plexus-utils? But if so, then a test is missing because this PR is passing.

cstamas:
This subproject uses plexus-utils, to be precise Xpp3Dom out of it. 

* plexus-utils 3.x contains it
* plexus-utils 4.x has it removed into plexus-xml
* plexus-xml 3.x is to be used with Maven 3, while plexus-xml 4.x is to be used with Maven 4 (as it in fact redirects to maven-xml)

Hence, I'd just downgrade the plexus-utils here to 3.x and done.

elharo:
Bleah. Another reason one shouldn't move classes between artifacts. Let me think about this.

cstamas:
superseded by https://github.com/apache/maven-resolver/pull/684

PR #659, opened 2025-02-24
[MRESOLVER-680] Disable checksum by default for .sigstore.json

thomasleplus:
The need for this change was confirmed with sigstore team in this [discussion](https://github.com/sigstore/sigstore-java/issues/905#issuecomment-2677275304).

This change was already done in [Maven GPG plugin](https://issues.apache.org/jira/browse/MGPG-130).


maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultArtifactPredicateFactory.java
```diff
@@ -51,7 +51,7 @@ public final class DefaultArtifactPredicateFactory implements ArtifactPredicateF
     public static final String CONFIG_PROP_OMIT_CHECKSUMS_FOR_EXTENSIONS =
             CONFIG_PROPS_PREFIX + "omitChecksumsForExtensions";
 
-    public static final String DEFAULT_OMIT_CHECKSUMS_FOR_EXTENSIONS = ".asc,.sigstore";
+    public static final String DEFAULT_OMIT_CHECKSUMS_FOR_EXTENSIONS = ".asc,.sigstore.json";
```
cstamas, line 54: Just like in referenced GPG PR, I'd leave both: `.sigstore` and `.sigstore.json`.

cstamas (changes requested):
Please append, do not remove old value.

cstamas:
Superseded by https://github.com/apache/maven-resolver/pull/661

Thanks for contribution, but we are in push for Maven 4 rc3 that will include new resolver release as well, so I "took over". Thanks again!

PR #660, opened 2025-02-24
[MRESOLVER-681] Sigstore TCCL bugfix

cstamas:
Seems sigstore depends on TCCL that went unnoticed in test, but explodes when this generator used as Maven extension.

---

https://issues.apache.org/jira/browse/MRESOLVER-681


jira-importer:
Resolve #1356

PR #672, opened 2025-03-17
[MRESOLVER-646] Use Maven 4 internal bits

cstamas:
---

https://issues.apache.org/jira/browse/MRESOLVER-646


cstamas:
This cannot work as is: "mvn4 bits" are Maven4 annotated, and bridge is not here (is in maven-core which we don't want here)...

gnodet:
> This cannot work as is: "mvn4 bits" are Maven4 annotated, and bridge is not here (is in maven-core which we don't want here)...

It may be possible to move the bridge in a separate jar, not sure...
Or do we want to use DI annotations for resolver ?

PR #685, opened 2025-04-13
[MRESOLVER-700] Bundle transport: read support

cstamas:
This PR slightly improves the file transport to fully utilize FileSystem, which now can be ZipFileSystem as well. The "old" part remains unchanged.

For factory just added `bundle:` prefix, that now may be in form `bundle:fileUri`, where `fileUri` is basically a path of a ZIP file that must exist. The ZIP file is read only.

The point of this PR is that user can now use a "bundle" (see https://central.sonatype.org/publish/publish-portal-upload/) just like we did before with staging repositories. Various tools (njord?) can produce bundles. The ZIP file is "mounted" and artifacts from it becomes reachable to Maven just like from any remote repository.

---

https://issues.apache.org/jira/browse/MRESOLVER-700

### DIFF
--- a/maven-resolver-transport-file/src/main/java/org/eclipse/aether/transport/file/FileTransporterFactory.java
+++ b/maven-resolver-transport-file/src/main/java/org/eclipse/aether/transport/file/FileTransporterFactory.java
@@ -98,8 +98,8 @@ public final class FileTransporterFactory implements TransporterFactory {
                 throw new UncheckedIOException(e); // hard failure; most probably user error (ie wrong path or perm)
             }
         } else {
-            // special case in file transport: to support custom FS providers (like JIMFS), we cannot
-            // cover "all possible protocols" to throw NoTransporterEx, but we rely on FS rejecting the URI
+            // special case in file: transport: to support custom FS providers (like JIMFS), we cannot
+            // cover all possible protocols (to throw NoTransporterEx), hence we rely on FS rejecting the URI
             FileTransporter.WriteOp writeOp = FileTransporter.WriteOp.COPY;
             if (repositoryUrl.startsWith("symlink+")) {
                 writeOp = FileTransporter.WriteOp.SYMLINK;
--- a/maven-resolver-transport-file/src/main/java/org/eclipse/aether/transport/file/FileTransporterFactory.java
+++ b/maven-resolver-transport-file/src/main/java/org/eclipse/aether/transport/file/FileTransporterFactory.java
@@ -20,9 +20,16 @@ package org.eclipse.aether.transport.file;
 
 import javax.inject.Named;
 
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.net.URI;
+import java.nio.file.FileSystem;
 import java.nio.file.FileSystemNotFoundException;
+import java.nio.file.FileSystems;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.HashMap;
+import java.util.Map;
 
 import org.eclipse.aether.RepositorySystemSession;
 import org.eclipse.aether.repository.RemoteRepository;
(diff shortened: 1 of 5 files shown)


slawekjaranowski:
Why `bundle` is read-only by default?

cstamas:
Initial idea is really just to disseminate the idea that one can "mount" ZIP files as remote repositories. Like repository staging, and there, the main use case is usually to perform some sort of "testing" with newly baked artifacts. This PR makes Resolver able to _consume_ these bundle ZIP files (essentially ZIP files containing artifacts laid down on remote repo layout).

cstamas:
... aand after I commented, I think figured your point :smile: Yes, ultimately Resolver 2 could be used to _create_ and _read_ ZIP bundles. But, IMO we need to be very careful here, like we should allow "create from scratch" (like ZIP file is created from scratch) or "read created bundle", where you really just consume bundle for testing purposes. Existing ZIP bundle should not be modified, or at least, would be risky business.

Njord enforces this: it does not allow you to "export" to already existing ZIP files for example. But we should somewhere sit down and have a :beers: that may be followed by a discussion about these things...

cstamas:
For example we could achieve this by some extra params:
* by default (as in this PR) the ZIP file must exists and is read obly
* if URL has `?create=true` QP, then Resolver could enforce ZIP file does not exists, and could make it read-write (and would become usable with `altDeployRepository` for example).

maven-resolver-transport-file/src/main/java/org/eclipse/aether/transport/file/FileTransporterFactory.java
```diff
@@ -66,28 +65,61 @@ public FileTransporterFactory setPriority(float priority) {
...
+                Map<String, String> env = new HashMap<>();
+                FileSystem fileSystem = FileSystems.newFileSystem(bundlePath, env);
+                return new FileTransporter(
+                        fileSystem,
+                        true,
+                        false,
+                        fileSystem.getPath(fileSystem.getSeparator()),
+                        FileTransporter.WriteOp.COPY);
+            } catch (IOException e) {
+                throw new UncheckedIOException(e); // hard failure; most probably user error (ie wrong path or perm)
```
slawekjaranowski, line 98: For others FS we have different exception handling .. 
can be the same as in lines 120-122

or one catch?

cstamas: Yes, for "generic" (existing) file we throw NoTransporterException that is "expected" by transport provider, and will make it to move to next factory, and finally will report error like "no transport for repo X". But, in case of "bundle:" transport, the file MUST exists and must be ZIP file and we have to be able to open it... if any of these fail, we do "hard failure" that will fail maven completely, as it is most probably user error, like bad ZIP file path or ZIP is simply not there...

cstamas: This is again along the lines of this:
* reading bundle: URL must point to existing ZIP file and we read only
* creating bundle: URL (once we do it) will have to point to non existing ZIP file

In both cases, we must ensure these are true, and do "hard fail". In case of "file" (and if you look at any other transport factory), the NoTransportException is more like factory stating "nah, I cannot handle this URL, go and try with another or report error"

cstamas:
Just one remark (somewhat related to this PR): While I worked on [Njord 0.1.2](https://github.com/maveniverse/njord/releases/tag/release-0.1.2) release, I figured that **even if we allow bundle creation with Resolver** (like the above mentioned `?create=true` QP), the bundle would be usable for central publishing **only**, which is I think still fine. But for more advanced control and support, like bundle merging, validating for requirements, managing, etc you'd need advanced tools like Njord is. Also, this way created bundle would **not be usable with Njord** (lack of metadata). Still, this functionality would provide one single building block needed for Central Publishing: the ZIP bundle itself. But how user would publish it, etc, would **NOT  be solved**.

jira-importer:
Resolve #1435

PR #723, opened 2025-05-13
[MRESOLVER-701] Support legacy property

cstamas:
For easier transitioning.

Same as MRESOLVER-652, the new property is documented, the old is not, it "just works".

---

https://issues.apache.org/jira/browse/MRESOLVER-701


slawekjaranowski (approved):

jira-importer:
Resolve #1436

PR #749, opened 2025-06-12
Collection hashCode

cstamas:
WIP

### DIFF
--- a/maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/manager/MMap.java
+++ b/maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/manager/MMap.java
@@ -1,0 +1,101 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.eclipse.aether.util.graph.manager;
+
+import java.util.HashMap;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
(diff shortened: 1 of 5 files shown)


gnodet (approved):

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/manager/MMap.java
```diff
@@ -0,0 +1,116 @@
...
+        @Override
+        public int hashCode() {
+            if (hashCode == null) {
+                synchronized (delegate) {
+                    if (hashCode == null) {
+                        hashCode = delegate.hashCode();
+                    }
+                }
+            }
+            return hashCode;
```
mbien: i would probably use the simple "racy" version of this similar like it used to be:
https://github.com/apache/maven-resolver/blob/8535d09660cb68eb55d8e75228d5d35eaae00c88/maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/manager/TransitiveDependencyManager.java#L232-L237

this should be safe since:
 - doc requires managers to be stateless, I assume this means all maps are immutable
 - this will be likely only used from one thread
 - race condition would not be a big deal since all it would to is to introduce a chance that the value is computed more than once under high contention

cstamas: I have other idea, will add `done()` method....

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/manager/MMap.java
```diff
@@ -0,0 +1,114 @@
...
+        return delegate.get(key);
+    }
+
+    public V put(K key, V value) {
+        return delegate.put(key, value);
+    }
+
+    public MMap<K, V> done() {
+        return new DoneMMap<>(delegate);
+    }
```
mbien, line 70:
```java
    return this instanceof DoneMMap ? this : new DoneMMap<>(delegate);
}
```
otherwise this would keep copying the maps even if nothing changed.

mbien (approved):
makes sense to me, left one comment.

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/manager/MMap.java
cstamas: but done method is overridden in DoneMMap? I don't get it

mbien: wait maybe I am wrong here - this likely can't happen since `done()` of the frozen instance is a no-op

mbien: sorry!

mbien:
it will copy the map twice now though.

first:
`MMap.copy(this.managedExclusions);`

second later:
`managedExclusions.done()`

which will hit the same copy constructor again.

PR #1490, opened 2025-06-14
Refine performance

XenoAmess:
refine performance for demo at https://github.com/apache/maven/issues/2481

special warning:
I'm not familiar with your DataPool key design, I just see the codes about it and think we can use that identity-hash there instead, because the hashing be too slow.
So if that can never be replaced to identity hash actually, please just tell me the reason, I would like to help with some other strategy...

### DIFF
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/DataPool.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/DataPool.java
@@ -513,7 +513,8 @@ public final class DataPool {
             this.traverser = traverser;
             this.filter = filter;
 
-            hashCode = Objects.hash(artifact, repositories, selector, manager, traverser, filter);
+            hashCode =
+                    Objects.hash(artifact, repositories, selector, System.identityHashCode(manager), traverser, filter);
         }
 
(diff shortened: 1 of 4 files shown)


maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/DataPool.java
```diff
@@ -513,7 +513,8 @@ static final class GraphKey {
             this.traverser = traverser;
             this.filter = filter;
 
-            hashCode = Objects.hash(artifact, repositories, selector, manager, traverser, filter);
+            hashCode =
+                    Objects.hash(artifact, repositories, selector, System.identityHashCode(manager), traverser, filter);
```
cstamas, line 517: identityHashCode is wrong here: what we want in fact is "state" of manager (ie. "same conditions for same artifacts gives same siblings").

XenoAmess: If so, then this pr shall be re-designed.
sigh.
If really wanna keep a state of the manager, then we have to analyze if the hashing process can be splited/re-used, which is very complicated then.

PR #1492, opened 2025-06-14
Refine performance (2nd)

XenoAmess:
refine performance for demo at https://github.com/apache/maven/issues/2481

including 3 parts:

1.  fix the bug about using MMap when value be collection, but the old codes didn't copy the collection, thus may cause appending to a list value in a DoneMMap, thus makes infinity long value, thus makes the run slow in super multi module project.
2.  BfDependencyCollector can share executorService between resolvers, to reduce the create/close/create/close of the executorService, which is expensive.
3.  intern some strings to make use of the new hash cache in String class in jdk.

### DIFF
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/bf/BfDependencyCollector.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/bf/BfDependencyCollector.java
@@ -119,11 +119,11 @@ public class BfDependencyCollector extends DependencyCollectorDelegate {
     public static final String CONFIG_PROP_THREADS = CONFIG_PROPS_PREFIX + "threads";
 
     /**
-     * The default value for {@link #CONFIG_PROP_THREADS}, default value {@code Runtime.getRuntime().availableProcessors()}.
+     * The default value for {@link #CONFIG_PROP_THREADS}, default value 5.
      *
      * @since 1.9.0
      */
-    public static final int DEFAULT_THREADS = Runtime.getRuntime().availableProcessors();
+    public static final int DEFAULT_THREADS = 5;
 
     @Inject
     public BfDependencyCollector(
@@ -470,7 +470,7 @@ public class BfDependencyCollector extends DependencyCollectorDelegate {
     }
 
     static class ParallelDescriptorResolver implements Closeable {
-        private static volatile ExecutorService executorService;
+        private final ExecutorService executorService;
 
         /**
          * Artifact ID -> Future of DescriptorResolutionResult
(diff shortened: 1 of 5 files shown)


XenoAmess:
@cstamas please have a look.
the most severe part is the first part.(MMap part)

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/bf/BfDependencyCollector.java
```diff
@@ -119,11 +119,11 @@ public class BfDependencyCollector extends DependencyCollectorDelegate {
     public static final String CONFIG_PROP_THREADS = CONFIG_PROPS_PREFIX + "threads";
 
     /**
-     * The default value for {@link #CONFIG_PROP_THREADS}, default value 5.
+     * The default value for {@link #CONFIG_PROP_THREADS}, default value {@code Runtime.getRuntime().availableProcessors()}.
      *
      * @since 1.9.0
      */
-    public static final int DEFAULT_THREADS = 5;
+    public static final int DEFAULT_THREADS = Runtime.getRuntime().availableProcessors();
```
cstamas: -1: we can run on monsters as well. This is not an "improvement" on alg, just "throwing more" on it. In other words, doing this may hit "remote end" as well, think of https://en.wikipedia.org/wiki/Tragedy_of_the_commons like Central is.

XenoAmess: @cstamas we increase the number because we try to merge the executers. see changes in several lines under that in the same file.
before, each instance have one seperated executor，and thus have 2 problems
1. when running with like T2C, I afraid there might be very many of instance in parallel, thus many small executors in parallel
2. the start and stop of the small executors be time costing.(especially adding worker threads)

Still，if you don't like mixing idea for merging the  executors of every instance, we can remove this part. I just don't think it good to start many small thread executors, and I still think a bigger single executor is better in this case.

rmannibucau: think we should use forkjoin pool by default, it is alsmot certainly created by a build these days and will avoid the overhead of custom pool with almost no benefit

XenoAmess: > think we should use forkjoin pool by default, it is alsmot certainly created by a build these days and will avoid the overhead of custom pool with almost no benefit

yes it would be far better than using many small pools.
But as for forkjoin pool, I have no idea why we should use forkjoin pool but not executer here. IMO fork join pool is use for works who have relationship/dependency, and the tasks here seems not quite dependent on each other..?

cstamas: https://issues.apache.org/jira/browse/MRESOLVER-283

XenoAmess: > https://issues.apache.org/jira/browse/MRESOLVER-283

Oh, I see https://github.com/apache/maven-resolver/pull/213 and the related comments...
Interesting.
Good, now I kind of getting interested in the fork-join-pool solution now.

@cstamas Is there any news since the original executor pr closed? should I start another pr(and split this from this pr) for that?

cstamas: No, nothing new since. It was just a faint idea, and did not move anywhere yet.

XenoAmess: > No, nothing new since. It was just a faint idea, and did not move anywhere yet.

Um. Seems it is good enough a thing that worth my time. Would have a try when next weekends.
I agree on split it out of this pr, it is a bigger thing than my original thought.
Though I do agree it seems a better way than I provided in this pr.
Please focus on reviewing other part of this pr then.

gnodet:
> I think the intern logic (apart from the `context` in trace requests) would be better placed inside Maven when parsing the POM files. It should be fairly simple, as it's just a matter of passing a `ContentTransformer` as an argument [to this call](https://github.com/apache/maven/blob/e80c9f5ac1cbeb3ffc263feaa5e310be5a761982/impl/maven-impl/src/main/java/org/apache/maven/impl/DefaultModelXmlFactory.java#L95). We should intern all enums (scopes, packaging) and at least groupId, artifactId, version. The models are kept in memory throughout the whole build anyway, so it makes sense to intern quite a few values.

I've raised https://github.com/apache/maven/pull/2495.  I was quite sure the work was done already, but after investigation, it had not been plugged in.

cstamas:
@XenoAmess refined it in #1508 please take a look

PR #1504, opened 2025-06-24
Drop Holder from depMgr

cstamas:
As it make no sense: depMgr holds "applicable" information in own state, while Holder added extra layer of "depth" info. Basically Holder and depMgr instance itself were redundant to each other.


cstamas:
This is wrong: makes IT MavenITmng8347TransitiveDependencyManagerTest fail (that in fact covers this exact spot).

The assessment "is redundant" is wrong. There is more to it.

PR #1508, opened 2025-06-25
Refine performance (alt)

cstamas:
Reusing #1492 but simplifying it: copy only the collection that is about to be modified, leave other unchanged (less copy less mem, less cycles).


XenoAmess:
> Reusing #1492 but simplifying it: copy only the collection that is about to be modified, leave other unchanged (less copy less mem, less cycles).

Looks good to me, you get the essence.
Though the hashing part of the creation of HashMap might be reduced, by introducing a specialized list structure for value (instead of arraylist) to avoid recalculating hashing, but it might be not worthy to bring the complexity.

PR #1510, opened 2025-06-26
Rollback #746 with comments

cstamas:
Also the added UTs made no sense from retrospective, as they used HTTP remote repo with HTTPS proxy, something Maven would never do.

Rolls back #746 but leaves traces and some context.


slawekjaranowski:
we can skip both from milestone and release notes

PR #1518, opened 2025-07-04
Ability to augment metadata nature for version range request

cstamas:
The idea is to let caller be able to augment the nature of metadata to resolve version range from.

Resolver already handles this, and is more elegant, than to have caller fiddle with filtering the remote repositories or alike. Also, this way client code (ie. Maven) can expose more fine grained configuration for this.


slawekjaranowski (approved):

cstamas:
This will need Maven changes once this version is integrated

PR #1519, opened 2025-07-09
unused dependency

elharo:


elharo:
OK, I'm looking at it but I'm not sure including the dependency is the correct or reliable way to ensure version alignment. It seems likely to depend on which dependencies are declared where in which order. Someone could still pull maven-resolver-util into their project with a different version before they pull in maven-resolver-supplier. Is there some reason that can't happen?

PR #1523, opened 2025-07-09
More semantic asserts

elharo:
and make sure the expected value always comes first


elharo:
Failure looks falky

Server not initialized yet, please try again.

slawekjaranowski (approved):

PR #1541, opened 2025-08-07
Resolver Thread management

### LINKED ISSUE #1537: would it be considerable to add Multi-Release JAR mechanism? virtual thread seems helpful in this repo.(at least helpful for bf)

1. add something like 
```
    <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-compiler-plugin</artifactId>
        <version>3.8.1</version> <!-- Use a version that supports multi-release -->
        <executions>
            <execution>
                <id>default-compile</id>
                <goals>
                    <goal>compile</goal>
                </goals>
                <configuration>
                    <release>8</release> <!-- Base version -->
                </configuration>
            </execution>
            <execution>
                <id>java9-compile</id>
                <goals>
                    <goal>compile</goal>
                </goals>
                <configuration>
                    <release>19</release>
                    <source>19</source>
                    <target>19</target>
                    <multiReleaseOutput>true</multiReleaseOutput>
                    <outputDirectory>${project.build.outputDirectory}/META-INF/versions/9</outputDirectory>
                    <includes>
                        <include>path/to/java19/specific/classes/**</include>
                    </includes>
                </configuration>
            </execution>
            <!-- Add more executions for other Java versions as needed -->
        </executions>
    </plugin>
```

```
    <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-jar-plugin</artifactId>
        <version>3.2.0</version>
        <configuration>
            <archive>
                <manifestEntries>
                    <Multi-Release>true</Multi-Release>
                </manifestEntries>
            </archive>
        </configuration>
    </plugin>
```

then modify the jdk19 version of that classes
then make ci use jdk21 to compile

cstamas:
Rework thread management.

High level changes:
* introduce simple `SmartExecutor` to hide underlying things; offer minimal surface just enough for consumers
* make 3 consumers use them with try-with-resources consistently (basic connector, metadata resolver, BF collector)
* on Java 21 use virtual threads, as all 3 consumer executes very IO bound tasks (ideal contenders)

Details:
* consumer basic connector logic simplified; instead of going directly for "direct" executor, use one single method, and perform direct executions as needed (as it had already all in place already); basic connector retains the ability to "direct execute" tasks when appropriate (as it knows ahead the task count)
* all 3 consumers use session cached `SmartExecutor` instances (basic connector exceptions for direct execution explained above)

Consequences: the parameters
* for basic connector: `aether.connector.basic.downstreamThreads`, `aether.connector.basic.upstreamThreads` and `aether.connector.basic.threads` (they may be suffixed with `.repoId`)
* for BF collector `aether.dependencyCollector.bf.threads`
* for metadata resolver `aether.metadataResolver.threads`

are from now on **global and reflect the reality for maven session**, in a way they are now truly aplied per-Maven session (limits now represent true per-session limits).

Before, this was not the case, and Maven was able to fully "step over" these limits: consider BF collector, it created always "local" thread pool with size 5. But alas, Maven 4 builds project models already in parallel (!), hence, BF collector (to resolve models for model builder) was already called from several threads, and each thread created "local" pool for it's own use.

This _may imply_ that pool sizes _may need a revisit_, as they _may_ be too conservative now (or they were just not truly applied, as intended).

Fixes #1537


slawekjaranowski (approved):

maven-resolver-util/src/main/java21/org/eclipse/aether/util/concurrency/SmartExecutorUtils.java
```diff
@@ -0,0 +1,103 @@
...
+            }
+            if (tasks == 1 || maxConcurrentTasks == 1) {
+                return DIRECT;
+            }
+        } else {
+            if (maxConcurrentTasks == 1) {
+                return DIRECT;
+            }
+        }
+        return new SmartExecutor.Limited(Executors.newThreadPerTaskExecutor(Thread.ofVirtual().name(namePrefix, 1).factory()),maxConcurrentTasks);
```
cstamas: @slawekjaranowski seems spotless ignores, or is unaware of this class, as this class clearly have formatting issues and build did not fail.

cstamas:
```suggestion
        return new SmartExecutor.Limited(Executors.newThreadPerTaskExecutor(Thread.ofVirtual().name(namePrefix, 1).factory()), maxConcurrentTasks);
```

maven-resolver-util/src/main/java21/org/eclipse/aether/util/concurrency/SmartExecutorUtils.java
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * @since 2.0.11
+ */
+public final class SmartExecutorUtils {
+    private static final SmartExecutor DIRECT = new SmartExecutor.Direct();
+
+    private SmartExecutorUtils() {}
+
+    /**
+     * Returns a smart executor for given parameters. If {@code tasks} is known (non-null), it must be grater than 0.
```
cstamas:
```suggestion
     * Returns a smart executor for given parameters. If {@code tasks} is known (non-null), it must be greater than 0.
```

maven-resolver-util/src/main/java/org/eclipse/aether/util/concurrency/SmartExecutorUtils.java
```diff
@@ -0,0 +1,104 @@
...
+ *
+ * @since 2.0.11
+ */
+public final class SmartExecutorUtils {
+    private static final SmartExecutor DIRECT = new SmartExecutor.Direct();
+
+    private SmartExecutorUtils() {}
+
+    /**
+     * Returns a smart executor for given parameters. If {@code tasks} is known (non-null), it must be grater than 0.
```
cstamas:
```suggestion
     * Returns a smart executor for given parameters. If {@code tasks} is known (non-null), it must be greater than 0.
```

maven-resolver-util/src/main/java21/org/eclipse/aether/util/concurrency/SmartExecutorUtils.java
slawekjaranowski: default value for java sources is:
```
<include>src/main/java/**/*.java</include>
<include>src/test/java/**/*.java</include>
```

https://github.com/diffplug/spotless/tree/main/plugin-maven#java
😄

XenoAmess:
> This may imply that pool sizes may need a revisit, as they may be too conservative now (or they were just not truly applied, as intended).

Yeah you are right...

PR #1543, opened 2025-08-11
Add scope support for trusted checksums

cstamas:
Currently TC operated on all resolved artifacts, but this may not be what user wants. Add scope support with two values for now: "all" (as before, everything resolved is validated) or "project" (only project dependencies are validated).


maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/resolution/TrustedChecksumsArtifactResolverPostProcessor.java
```diff
@@ -147,6 +165,18 @@ protected boolean isEnabled(RepositorySystemSession session) {
         return ConfigUtils.getBoolean(session, false, CONFIG_PROP_ENABLED);
     }
 
+    private boolean inScope(RepositorySystemSession session, ArtifactResult artifactResult) {
+        String scope = ConfigUtils.getString(session, DEFAULT_SCOPE, CONFIG_PROP_SCOPE);
+        if (ALL_SCOPE.equals(scope)) {
+            return artifactResult.isResolved();
+        } else if (PROJECT_SCOPE.equals(scope)) {
+            return artifactResult.isResolved()
+                    && artifactResult.getRequest().getRequestContext().startsWith("project");
```
slawekjaranowski, line 174: Where request context is set ... or by who?
Are we sure that all project artifacts have such value?

What when some of resolving is triggered by plugin?

cstamas: Context is set by core:
* for projects https://github.com/apache/maven/blob/1f97959a11cc4277d42da08ff581cc061849cea8/maven-core/src/main/java/org/apache/maven/project/DefaultProjectDependenciesResolver.java#L95
* for plugins https://github.com/apache/maven/blob/1f97959a11cc4277d42da08ff581cc061849cea8/maven-core/src/main/java/org/apache/maven/plugin/internal/DefaultPluginDependenciesResolver.java#L199
* for rest it is "free game" (usually people leave it alone), like ad hoc collect request from a plugin. For example in Toolbox I consistently use "toolbox": https://github.com/maveniverse/toolbox/blob/6442611992f83a85de5235616c38c05cb180ec8e/shared/src/main/java/eu/maveniverse/maven/toolbox/shared/internal/ToolboxResolverImpl.java#L509

Later on, kicks in the "context refiner": https://github.com/apache/maven-resolver/blob/bf3f23592499104b2f9bff5b9a2d58a1936ca99f/maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/transformer/JavaDependencyContextRefiner.java#L37  that refines context from "project" to something like "project/test".

All in all, only one thing is sure:
* core sets "project" and "plugin" context, and this is 100% true for project dependencies ("project"), and build plugins ("plugin"). Now, if anyone does ad-hoc collect request, that anyway out of scope.

Hence the "all" (as before, context is totally neglected) and "project" scopes, as that manages project dependencies (a plugin doing ad-hoc lookup is not project dependency).

PR #1565, opened 2025-08-29
Fixes regarding nearest/highest strategies and test

### LINKED ISSUE #1562: Follow Up for ConflictResolver improvements
Follow up to #1547 

The PR has been merged, but:
* introduces a change to `ConflictResolver` that was the only implementation and a `final` class, but now is `abstract`.
* introduced two "coexisting" conflict resolver implementations ("classic" and "path")
* ~the superior (faster) "path" needs to be checked against other version selection strategies than "nearest", if fails it may need more~ seems it works as expected *IF* old DF or new BF collector w/ skipper=false is used. If skipper is enabled, BF performs unwanted premature optimizations.
* do we want to do something about this?

### LINKED ISSUE #1566: New BF collector Skipper implementation prematurely optimizes
### Affected version

All with BF 1.9+

### Bug description

And hence, version selection strategy other than "nearest" is not usable with it (to work, one must use old DF or new BF w/ skipper disabled).

cstamas:
And some other cleanup. Fixed highest strategy (introduced in 2.0.0). Added UT with some examples to compare differences of behaviour of two strategies (nearest vs highest). And finally, there is some minor cleanup in demo snippets, better naming local reposes.

This PR now makes Resolver able to:
* dynamically choose conflict resolver based on session config (def is new "path")
* dynamically choose version selection strategy for conflict resolution, based on session config (def is "nearest")
* dynamically choose skipper type: versionless (default) or versioned (must be used with "highest")

Fixes #1562
Fixes #1566

### DIFF
--- a/maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/transformer/ConfigurableVersionSelector.java
+++ b/maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/transformer/ConfigurableVersionSelector.java
@@ -202,7 +202,8 @@ public class ConfigurableVersionSelector extends VersionSelector {
 
     @Override
     public String toString() {
-        return getClass().getSimpleName() + "(" + (selectionStrategy != null ? selectionStrategy.getClass().getSimpleName() : "not inited") + ")";
+        return getClass().getSimpleName() + "("
+                + (selectionStrategy != null ? selectionStrategy.getClass().getSimpleName() : "not inited") + ")";
     }
 
     /**


slawekjaranowski:
We have introduce new definitions as GACE / GACEV - what does ments, how to extends such shortcuts?

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/transformer/ConfigurableVersionSelector.java
```diff
@@ -157,6 +200,11 @@ protected boolean isAcceptableByConstraints(ConflictGroup group, Version version
         return true;
     }
 
+    @Override
+    public String toString() {
+        return getClass().getSimpleName() + "(" + selectionStrategy.getClass().getSimpleName() + ")";
```
cstamas:
```suggestion
        return getClass().getSimpleName() + "(" + (selectionStrategy != null ? selectionStrategy.getClass().getSimpleName() : "not inited") + ")";
```

PR #1570, opened 2025-08-30
Repository ID handling improvement

### LINKED ISSUE #1564: RepositoryID with slash in it
### Affected version

All

### Bug description

I just found a POM that defines a repository with slash (`/`) in repository ID, that is invalid character for Maven ID fields. 

This breaks:
* split repository
* RRF prefix caching

cstamas:
Many feature blindly assume (as Maven validates this) that `repository.id` is "path segment"-safe string.

Still, we found some fairly recent POMs (like maven-core 3.6.1!) that contain in their transitive hull POMs that contain FS unsafe repository IDs.

These IDs makes features like split repository, or metadata caching go hay-way.

This PR contains fix for these "bad" repository.id-s (by "fixing" them to be as should: FS safe) covering all places where repository.id is to be used as path segment.

Fixes #1564

### DIFF
--- a/maven-resolver-util/src/main/java/org/eclipse/aether/util/repository/RepositoryIdHelper.java
+++ b/maven-resolver-util/src/main/java/org/eclipse/aether/util/repository/RepositoryIdHelper.java
@@ -18,9 +18,9 @@
  */
 package org.eclipse.aether.util.repository;
 
-import java.util.Arrays;
 import java.util.Collections;
-import java.util.List;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Function;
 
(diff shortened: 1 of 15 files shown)


cstamas:
Local testing: 
* build this PR fully w/ ITs
* after the build, the "demos" local repository (`maven-resolver-demo-snippets/target/local-repo/.meta/resolver-status.properties` file) will contain "rewritten `oss.sonatype.orgXjboss-snapshots` repo ID `` instead of illegal `oss.sonatype.org/jboss-snapshots`

maven-resolver-util/src/main/java/org/eclipse/aether/util/repository/RepositoryIdHelper.java
```diff
@@ -0,0 +1,108 @@
...
+     * see {@link #cachedIdToPathSegment(RepositorySystemSession)} method.
+     * <p>
+     * This method is visible for testing only, should not be used in any other scenarios.
+     *
+     * @see #cachedIdToPathSegment(RepositorySystemSession)
+     */
+    static String idToPathSegment(ArtifactRepository repository) {
+        if (repository instanceof RemoteRepository) {
+            StringBuilder result = new StringBuilder(repository.getId());
+            for (int illegalIndex = 0; illegalIndex < ILLEGAL_REPO_ID_CHARS.size(); illegalIndex++) {
```
gnodet: This looks overly complicated.
I'd replace all invalid chars with '-' instead of 'X' and remove the possibility to define different chars.  Especially as this is not configurable.
The loop may be simplified once the above is done, maybe using replaceAll? Not sure...

cstamas: The whole point is this happens in cached call, so as javadoc notes, this method is "is (intentionally) simplistic on purpose", but is not public either, calling it is possible _only via cached one_. 

As for "X" vs "-" vs whatever, is really just matter of taste, as I was even toying to do something like this:
```
| -> PIPE
: -> COLON
/ -> SLASH
\ -> BACKSLASH
...
```
so basically it would be clear (but still "path segment safe") that this ID is "weird".

cstamas: Doing that "dict replace" would in fact make different from IDs still unique, as now `repo/id` and `repo:id` both maps to `repoXid`, while in latter case those would be `repoCOLONid` and `repoSLASHid`

gnodet: "X" vs "-" is a matter of readability, more than taste.  Anything that's not a letter will be easier to read to separate groups of letters.

I'd be fine with specific words, but again for readability, I'd move them between `-` or `_`:  
  `foo/bar` -> `foo_SLASH_bar`
  
  That would be fine imho, especially to keep the ID unique as you indicate.

cstamas: pushed "dict", and separators are needed next

cstamas: added separators to replacement strings, hopefully we will never deal with "bad IDs" like in UT: `"\\/:\"<>|?*"` to worry about neighboring separators :smile:

cstamas: In fact: this makes function reversible, so is even better. Again, this is not for "human readability" but more to not make features like split repo and metadata caching "wander" to wrong dirs on your disk.

cstamas: In other word, this feature is used ONLY when resolver is about to use repository.id as path segment... nowhere else. Exceptions and log messages still use "raw" repo IDs.

gnodet (approved):

Issue #1583, opened 2025-09-13
StackOverflowError in ManagedDependencyContextRefiner


slawekjaranowski:
### Affected version

2.0.11

### Bug description

Discovered in enforcer ITs build by 
`Apache Maven 4.1.0-SNAPSHOT (b4ee1aa0450ff87433a978c2989c0ac432496863)`

 - dependency-convergence-cycle
 - MENFORCER-408
 - tree-verbose

https://github.com/apache/maven-enforcer

```
Exception in thread "main" java.lang.StackOverflowError
	at org.eclipse.aether.internal.impl.scope.ManagedDependencyContextRefiner.transformGraph(ManagedDependencyContextRefiner.java:53)
	at org.eclipse.aether.internal.impl.scope.ManagedDependencyContextRefiner.transformGraph(ManagedDependencyContextRefiner.java:62)
	at org.eclipse.aether.internal.impl.scope.ManagedDependencyContextRefiner.transformGraph(ManagedDependencyContextRefiner.java:62)
	at org.eclipse.aether.internal.impl.scope.ManagedDependencyContextRefiner.transformGraph(ManagedDependencyContextRefiner.java:62)
	at org.eclipse.aether.internal.impl.scope.ManagedDependencyContextRefiner.transformGraph(ManagedDependencyContextRefiner.java:62)
...
```

it is ok with:
`Apache Maven 4.0.0-rc-4 (bed0f8174bf728978f86fac533aa38a9511f3872)`

cstamas:
Hm, for me these failed:
```
[ERROR] The following builds failed:
[ERROR] *  dependency-convergence-cycle/pom.xml
[ERROR] *  require-upper-bound-deps_failure_show_scopes/pom.xml
[ERROR] *  MENFORCER-408/pom.xml
```

cstamas:
Confirmed, this is the new "path" conflict resolver, it does not cleans up all cycles as "classic" does:
https://gist.github.com/cstamas/406b84fe7c089db327540ed85f8ac070

PR #1584, opened 2025-09-13
Bug: path conflict resolver leaves in cycles

cstamas:
Fix: remove the cycle. Tests needs to be added, but this one fixes the issue (by running enforcer ITs).

Fixes: #1583


cstamas:
Bug lies in `org.eclipse.aether.util.graph.transformer.PathConflictResolver.Path#push` method: we have the winner, BUT the winner also introduces cycles (cycle cut happens ONLY in loser branch). And we do not have losers as well, only one `com.codeborne:phantomjsdriver` is present in tree.

cstamas:
No, UT makes classic fail, not the "path", so reproducer is not good.... hmm

PR #1588, opened 2025-09-16
Document deployment of artifacts

### LINKED ISSUE #1587: Document deploy API

The default deploy method uses a PUT request to place artifacts in a remote repository (https://github.com/apache/maven-resolver/blob/bcd3eded2cefeb234ae0588f29c55f48be954c12/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultDeployer.java#L86). This should be documented along with possibilities to use a different ReST API like a POST or https://central.sonatype.org/publish/publish-portal-api/#uploading-a-deployment-bundle.

kwin:
This closes #1587

### DIFF
--- a/src/site/markdown/deployment.md
+++ b/src/site/markdown/deployment.md
@@ -1,0 +1,44 @@
+# Deployment
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+Deploying artifacts and related metadata to a (remote) repository can be achieved via Resolver API with method [`org.eclipse.aether.RepositorySystem.deploy(RepositorySystemSession session, DeployRequest request)`](https://github.com/apache/maven-resolver/blob/master/maven-resolver-api/src/main/java/org/eclipse/aether/RepositorySystem.java). This writes/uploads the given artifact(s) including metadata to the given repository leveraging a [`RepositoryConnector`](https://github.com/apache/maven-resolver/blob/master/maven-resolver-spi/src/main/java/org/eclipse/aether/spi/connector/RepositoryConnector.java).
+
+The most prominent consumer of this API is probably [maven-deploy-plugin](https://maven.apache.org/plugins/maven-deploy-plugin/).
+
+# Repository Connector
+
+The default repository connector implementation at [`BasicRepositoryConnectorFactory`](https://github.com/apache/maven-resolver/blob/master/maven-resolver-connector-basic/src/main/java/org/eclipse/aether/connector/basic/BasicRepositoryConnector.java) uses a `RepositoryLayout` to calculate the URL and a `Transporter` to achieve the actual upload/write of artifacts/metadata.
+
+## Repository Layout
+
+The repository layout determines the location to which the artifact is being written/uploaded with its `RepositoryLayout.getLocation(Artifact, true)` or `RepositoryLayout.getLocation(Metadata, true)` method. For [Maven 2 repositories](https://maven.apache.org/repositories/layout.html) the logic is implemented in [`Maven2RepositoryLayoutFactory`](https://github.com/apache/maven-resolver/blob/master/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/Maven2RepositoryLayoutFactory.java).
+
+## Transporter
+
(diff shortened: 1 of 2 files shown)


cstamas:
The "Maven way" of wanted topic "deploy API using PUT by default for HTTP" is true IF following conditions are true:
* last connector in pipeline (mvn3 does not have it, but similar thing happens effectively, while mvn4 has dedicated API for this for programmatic construction) is `basic` connector
* transport being used is some of HTTP implementations (`apache`, `jdk` or `jetty`) -- _usually_ transport selection happens based 0n [`RemoteRepository#getProtocol()`](https://github.com/apache/maven-resolver/blob/master/maven-resolver-api/src/main/java/org/eclipse/aether/repository/RemoteRepository.java#L131-L139) (but as always, there are exceptions)

In that case, one can say that "deploy" call maps to HTTP PUT method for _each artifact and metadata_.

src/site/markdown/deployment.md
```diff
@@ -0,0 +1,44 @@
...
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+Deploying artifacts and related metadata to a (remote) repository can be achieved via Resolver API with method [`org.eclipse.aether.RepositorySystem.deploy(RepositorySystemSession session, DeployRequest request)`](https://github.com/apache/maven-resolver/blob/master/maven-resolver-api/src/main/java/org/eclipse/aether/RepositorySystem.java). This writes/uploads the given artifact(s) including metadata to the given repository leveraging a [`RepositoryConnector`](https://github.com/apache/maven-resolver/blob/master/maven-resolver-spi/src/main/java/org/eclipse/aether/spi/connector/RepositoryConnector.java).
```
kwin, line 21: What about checksums? Are they treated as separate artifacts like signatures?

cstamas: No, checksums are feature of basic connector (check out M2 layout, that one contains the "configured" checksums), and they are opaquely (to client) being uploaded (on deploy) or being asked for (on consume) by basic connector. No client code should ever mangle with them (while deploying or resolving).

Though, as they are treated/cached in same was as artifacts and metadata, they are stored in local repository (on consume), and also, they **can** be explicitly resolved too by same calls as artifacts or metadata is resolved.

In short: checksums are _NOT_ separate artifacts (even if they do behave like "separate artifacts" in some respect), but there are important differences. In general, you should NOT assume they are separate artifacts, you should leave their handling to basic connector: the role of basic connector among others, is that when it returns from a call with downloaded artifact, you KNOW the checksums are checked and are okay (according to session config like checksum policy and configured checksum algs), otherwise basic connector fails the call.

cstamas: And one more important point: if you DO provide checksums, basic connector will let them pass (and deploy in opaque manner) **unless configured to use same checksum algorithm**. Think ASF source bundle deploys, we use a plugin to create SHA512 for them, (while Resolver is using 'standard' checksum config of MD5 and SHA1). IF Resolver would be configured to use SHA512 as well (mvn3: `-Daether.checksums.algorithms=...` and mvn4: `aether.layout.maven2.checksumAlgorithms`), it would _recalculate and replace checksum with its own_ on deploy.

cstamas: Also, important is the "kind" of checksum:
https://github.com/apache/maven-resolver/blob/f737339fd2209044d8e8b2315ccd367cd2a841b4/maven-resolver-spi/src/main/java/org/eclipse/aether/spi/connector/checksum/ChecksumPolicy.java#L65-L93

cstamas:
Maybe just emphasize that "basic connector + HTTP transport" is the "most common way users are using Maven" (given Central is HTTP)? But, that is exactly NOT the typical how they deploy :smile: 

Unsure, what we want really to document here... as today, NOTHING aside of MRMs (that provide "internal" or "private" repositories, and usually proxies as well) support the "basic Maven deploy" (HTTP PUT artifact to remote repo) use case.

cstamas:
Just for ref, the Slack discussion as well:

So, I don't know can we go with it so simple, as for example, new mvn4 feature is this (visible with -X):
```
[DEBUG] Final pipeline: offline( rrf( basic( apache-snapshots-b6ad1b6b4141cf7b4e128785f79cbc45945d1ec4 (https://repository.apache.org/content/groups/snapshots, default, snapshots) ) ) )
```

Maven 4 now "pipeline" connectors, and this is final assembled pipeline in resolver 2:
```
offline( rrf( basic( target-usually remote repo ) ) )
```

Roles:
* offline, if repo offline, "shortcuts" and refuse request
* rrf, based (and if present) RRF rules, bounce/refuse requests
* basic, is our "basic" transport, handles transport (Transport) and checksums (corruption protection)

Transport is implementing actual transport like HTTP, FTP, S3 etc. Also, we plan to add new connectors in pipeline, `like sigcheck()`.

What this doco describe is true ONLY if `basic` connector is used (it implements the well known "roundtrip" maven repo, which focuses on immediate availability), but one could implement other connectors, like for example P2 is (gets stuff from P2 repo), or OBR and those formats are wildly different (btw, the connector/transport separation happened quite late, around maven 3.3.? As before connector+transport was one). For example, Njord is another connector as well (not transport, it in fact reuses file transport from resolver). Mimir is another connector, that implements read-through cache around `basic`.

Resolver is layered on purpose, and at the same time due this is not quite trivial to explain how it works.

Also, important distinction about "abstraction levels": connectors deals with Artifacts, while transport only about URIs. Transport is blind for "which artifact is this URI for"-like questions.

PR #1594, opened 2025-09-18
Go back to atomic

cstamas:
This change was done due Windows, but later the execution path was split for Windows to use oldie pre-NIO2 code to copy as for some reason that works on Windows.

On other platforms, insist on ATOMIC, still make use able to override it, in case atomic is not a viable option.


cstamas:
make it configurable...

PR #1599, opened 2025-09-22
Demote JDK transport

cstamas:
Originally we have this order of transports w/ priorities (present/absent in maven core):
* wagon -1.0 (present)
* apache 5.0 (present)
* jdk 10.0 (present) ("default" transport)
* jetty 15.p (absent) (if user added it via extension or lib/ext it "wins")

But lately we see that JDK one advantage of HTTP/2 support (that is already disabled due various issues) is shaded due several disadvantages (no support for HTTPS proxies, unstable, and unreliable)

Hence, I propose to "demote" it (maybe even remove it from Maven core):
* wagon -1.0 (present)
* jdk 5.0 (present; and maybe remove it from maven core)
* apache 10.0 (present; new "default" transport)
* jetty 15.0 (absent)

This makes Maven 4 not have "modern" HTTP transport by default, but apache transport is well battle tested and robust, with one issue: lack of HTTP/2 support (that does improve download speeds).

### DIFF
--- a/maven-resolver-transport-apache/src/main/java/org/eclipse/aether/transport/apache/ApacheTransporterFactory.java
+++ b/maven-resolver-transport-apache/src/main/java/org/eclipse/aether/transport/apache/ApacheTransporterFactory.java
@@ -39,7 +39,7 @@ import static java.util.Objects.requireNonNull;
 public final class ApacheTransporterFactory implements HttpTransporterFactory {
     public static final String NAME = "apache";
 
-    private float priority = 5.0f;
+    private float priority = 10.0f;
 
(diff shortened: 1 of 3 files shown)


cstamas:
Just FYI, and sorry for the confusion, It was me hastily "demoting" but
then after change in mind doing this:
https://github.com/apache/maven-resolver/pull/1611

So, mvn4-rc-5 did NOT demote JDK transport after all.
Still, this will not reduce deps as Martin writes, as HttpClient transport
is still present in Maven (in fact, HttpClient is a requirement of Wagon,
that in turn, is needed to provide Maven 3 backward compat).
We may drop all these only in Maven 5 I guess.

T

PR #1610, opened 2025-10-01
GH-1609 Bad metadata naming

### LINKED ISSUE #1609: Prefix locking issues
### Affected version

2.0.11

### Bug description

```
Caused by: java.lang.IllegalStateException: Could not acquire shared lock for [] using lock file:///home/runner/work/maven/maven/its/core-it-suite/target/user-home/.m2/repository/.locks/metadata~.lock in 30 SECONDS
    at org.eclipse.aether.internal.impl.synccontext.named.NamedLockFactoryAdapter$AdaptedLockSyncContext.acquire(NamedLockFactoryAdapter.java:260)
    at org.eclipse.aether.internal.impl.DefaultMetadataResolver.resolve(DefaultMetadataResolver.java:163)
    at org.eclipse.aether.internal.impl.DefaultMetadataResolver.resolveMetadata(DefaultMetadataResolver.java:149)
    at org.eclipse.aether.internal.impl.filter.PrefixesRemoteRepositoryFilterSource.lambda$resolvePrefixesFromRemoteRepository$4(PrefixesRemoteRepositoryFilterSource.java:292)
    at org.eclipse.aether.internal.impl.filter.PrefixesRemoteRepositoryFilterSource.ongoingUpdatesGuard(PrefixesRemoteRepositoryFilterSource.java:213)
    at org.eclipse.aether.internal.impl.filter.PrefixesRemoteRepositoryFilterSource.resolvePrefixesFromRemoteRepository(PrefixesRemoteRepositoryFilterSource.java:306)
    at org.eclipse.aether.internal.impl.filter.PrefixesRemoteRepositoryFilterSource.loadPrefixTree(PrefixesRemoteRepositoryFilterSource.java:226)
    at org.eclipse.aether.internal.impl.filter.PrefixesRemoteRepositoryFilterSource.lambda$cachePrefixTree$1(PrefixesRemoteRepositoryFilterSource.java:206)
    at java.util.concurrent.ConcurrentHashMap.computeIfAbsent(ConcurrentHashMap.java:1724)
    at org.eclipse.aether.internal.impl.filter.PrefixesRemoteRepositoryFilterSource.lambda$cachePrefixTree$2(PrefixesRemoteRepositoryFilterSource.java:205)
    at org.eclipse.aether.internal.impl.filter.PrefixesRemoteRepositoryFilterSource.ongoingUpdatesGuard(PrefixesRemoteRepositoryFilterSource.java:213)
    at org.eclipse.aether.internal.impl.filter.PrefixesRemoteRepositoryFilterSource.cachePrefixTree(PrefixesRemoteRepositoryFilterSource.java:203)
    at org.eclipse.aether.internal.impl.filter.PrefixesRemoteRepositoryFilterSource.access$200(PrefixesRemoteRepositoryFilterSource.java:79)
    at org.eclipse.aether.internal.impl.filter.PrefixesRemoteRepositoryFilterSource$PrefixesFilter.acceptPrefix(PrefixesRemoteRepositoryFilterSource.java:344)
    at org.eclipse.aether.internal.impl.filter.PrefixesRemoteRepositoryFilterSource$PrefixesFilter.acceptArtifact(PrefixesRemoteRepositoryFilterSource.java:327)
    at org.eclipse.aether.internal.impl.filter.DefaultRemoteRepositoryFilterManager$Participants.lambda$acceptArtifact$0(DefaultRemoteRepositoryFilterManager.java:93)
    at java.util.stream.Collectors.lambda$uniqKeysMapAccumulator$0(Collectors.java:180)
    at java.util.stream.ReduceOps$3ReducingSink.accept(ReduceOps.java:169)
    at java.util.Collections$UnmodifiableMap$UnmodifiableEntrySet.lambda$entryConsumer$0(Collections.java:1779)
    at java.util.HashMap$EntrySpliterator.forEachRemaining(HashMap.java:1859)
    at java.util.Collections$UnmodifiableMap$UnmodifiableEntrySet$UnmodifiableEntrySetSpliterator.forEachRemaining(Collections.java:1804)
    at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:570)
    at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:560)
    at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921)
    at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:265)
    at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:727)
    at org.eclipse.aether.internal.impl.filter.DefaultRemoteRepositoryFilterManager$Participants.acceptArtifact(DefaultRemoteRepositoryFilterManager.java:92)
    at org.eclipse.aether.internal.impl.DefaultArtifactResolver.resolve(DefaultArtifactResolver.java:262)
    at org.eclipse.aether.internal.impl.DefaultArtifactResolver.resolveArtifacts(DefaultArtifactResolver.java:201)
    at org.eclipse.aether.internal.impl.DefaultRepositorySystem.resolveArtifacts(DefaultRepositorySystem.java:230)
    at org.apache.maven.impl.DefaultArtifactResolver.lambda$doResolve$0(DefaultArtifactResolver.java:111)...
```

cstamas:
It caused in certain cases that lock names for different metadata collapsed onto same name, and hence, congestion with possible timeouts on resource happened. In case of metadata, `Metadata#getType` was totally ignored, which turns out was wrong, as today we have more types than the simple `maven-metadata.xml`.

Fix: We already had means to make sure "string is valid for path segment", but it was buried in `RepositoryIdHelper`. Pull it out, and make it reusable, as this is essentially very same use case. Also, this extra work really "punishes" only non maven metadata, and that is okay (right now we have only repository prefix and archetype catalog files in play).

Fixes #1609

### DIFF
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/synccontext/named/GAVNameMapper.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/synccontext/named/GAVNameMapper.java
@@ -24,6 +24,7 @@ import java.util.TreeSet;
 
 import org.eclipse.aether.RepositorySystemSession;
 import org.eclipse.aether.artifact.Artifact;
+import org.eclipse.aether.metadata.DefaultMetadata;
 import org.eclipse.aether.metadata.Metadata;
 import org.eclipse.aether.named.NamedLockKey;
 
@@ -103,18 +104,46 @@ public class GAVNameMapper implements NameMapper {
                 + suffix;
     }
 
+    private static final String MAVEN_METADATA = "maven-metadata.xml";
+
     private static String getMetadataName(Metadata metadata, String prefix, String separator, String suffix) {
-        String name = prefix;
-        if (!metadata.getGroupId().isEmpty()) {
-            name += metadata.getGroupId();
-            if (!metadata.getArtifactId().isEmpty()) {
-                name += separator + metadata.getArtifactId();
-                if (!metadata.getVersion().isEmpty()) {
-                    name += separator + metadata.getVersion();
+        if (metadata.getType().contains("/")) {
+            String metadataType = metadata.getType();
+            int slash = metadataType.indexOf('/');
+            return metadataType.substring(0, slash)
+                    + separator
+                    + getMetadataName(
+                            new DefaultMetadata(
+                                    metadata.getGroupId(),
+                                    metadata.getArtifactId(),
+                                    metadata.getVersion(),
+                                    metadataType.substring(slash + 1),
+                                    metadata.getNature(),
+                                    metadata.getProperties(),
+                                    metadata.getPath()),
+                            prefix,
+                            separator,
+                            suffix);
+        } else {
+            String name = prefix;
+            if (!metadata.getGroupId().isEmpty()) {
+                name += metadata.getGroupId();
+                if (!metadata.getArtifactId().isEmpty()) {
+                    name += separator + metadata.getArtifactId();
+                    if (!metadata.getVersion().isEmpty()) {
+                        name += separator + metadata.getVersion();
+                    }
+                }
+                if (!MAVEN_METADATA.equals(metadata.getType())) {
+                    name += separator + metadata.getType();
+                }
+            } else {
+                if (!MAVEN_METADATA.equals(metadata.getType())) {
+                    name += metadata.getType();
                 }
             }
+            return name + suffix;
         }
-        return name + suffix;
     }
 
     public static NameMapper gav() {
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/synccontext/named/GAVNameMapper.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/synccontext/named/GAVNameMapper.java
@@ -26,6 +26,7 @@ import org.eclipse.aether.RepositorySystemSession;
 import org.eclipse.aether.artifact.Artifact;
 import org.eclipse.aether.metadata.Metadata;
 import org.eclipse.aether.named.NamedLockKey;
+import org.eclipse.aether.util.PathUtils;
 
 import static java.util.Objects.requireNonNull;
 
@@ -106,10 +107,6 @@ public class GAVNameMapper implements NameMapper {
     private static final String MAVEN_METADATA = "maven-metadata.xml";
 
     private static String getMetadataName(Metadata metadata, String prefix, String separator, String suffix) {
-        String type = metadata.getType();
-        if (type.contains("/")) {
-            type = type.replaceAll("/", "_");
-        }
         String name = prefix;
         if (!metadata.getGroupId().isEmpty()) {
             name += metadata.getGroupId();
@@ -119,12 +116,12 @@ public class GAVNameMapper implements NameMapper {
                     name += separator + metadata.getVersion();
                 }
             }
-            if (!MAVEN_METADATA.equals(type)) {
-                name += separator + type;
+            if (!MAVEN_METADATA.equals(metadata.getType())) {
+                name += separator + PathUtils.stringToPathSegment(metadata.getType());
             }
         } else {
-            if (!MAVEN_METADATA.equals(type)) {
-                name += type;
+            if (!MAVEN_METADATA.equals(metadata.getType())) {
+                name += PathUtils.stringToPathSegment(metadata.getType());
             }
         }
         return name + suffix;
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/synccontext/named/GAVNameMapper.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/synccontext/named/GAVNameMapper.java
@@ -24,7 +24,6 @@ import java.util.TreeSet;
 
 import org.eclipse.aether.RepositorySystemSession;
 import org.eclipse.aether.artifact.Artifact;
-import org.eclipse.aether.metadata.DefaultMetadata;
 import org.eclipse.aether.metadata.Metadata;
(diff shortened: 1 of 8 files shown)


maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/synccontext/named/GAVNameMapper.java
```diff
@@ -103,18 +104,46 @@ private static String getArtifactName(Artifact artifact, String prefix, String s
...
-                name += separator + metadata.getArtifactId();
-                if (!metadata.getVersion().isEmpty()) {
-                    name += separator + metadata.getVersion();
+        if (metadata.getType().contains("/")) {
+            String metadataType = metadata.getType();
+            int slash = metadataType.indexOf('/');
+            return metadataType.substring(0, slash)
+                    + separator
+                    + getMetadataName(
+                            new DefaultMetadata(
```
gnodet: It looks inefficient to create an object just to call the method recursively.  It may be cleaner to define a new method which would have the individual parameters.

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/synccontext/named/GAVNameMapper.java
```diff
@@ -103,18 +104,46 @@ private static String getArtifactName(Artifact artifact, String prefix, String s
...
-        if (!metadata.getGroupId().isEmpty()) {
-            name += metadata.getGroupId();
-            if (!metadata.getArtifactId().isEmpty()) {
-                name += separator + metadata.getArtifactId();
-                if (!metadata.getVersion().isEmpty()) {
-                    name += separator + metadata.getVersion();
+        if (metadata.getType().contains("/")) {
+            String metadataType = metadata.getType();
+            int slash = metadataType.indexOf('/');
+            return metadataType.substring(0, slash)
```
gnodet: Why do we put the first part of the metadata in front while we usually used `g/a/v/t` ?
What are the types containing a `/` ?

cstamas: This needs a bit of history.... originally Maven had no means to (re)use Resolver transport to fetch (or put) non "standard" (not on layout) resources, and similarly, you could not reuse cache/checksum policies to perform such a transport, youd need to redo everything. As everything was either artifact or maven metadata. But with latest Resolver, metadata URI is calculated like this:

Metadata is G, A, V and T. Remote URI is constructed as:
* remote repository `baseUrl`
* if G given, `+G.replaceAll(".", "/")` -- as artifact groupId
* if A given, `+A` -- as artifact artifactId
* if V given, `+V` -- as artifact version
* `+type`

Type was always "maven-metadata.xml". This is how you can make Resolver download the Maven Repository Metadata from G (plugin) level, GA (versions) level, and GAV (snapshot resolution) level.

Later, we realized that type does not have to be always "maven-metadata.xml", but can be something else, so for example Archetype Catalog (that sits in [repo root](https://repo.maven.apache.org/maven2/archetype-catalog.xml)) can be addressed (and hence, use same transport to get artifacts) as this (this is how it is done in m-archetype-p to get and cache the catalog):

`Metadata(G:"", A:"", V:"", Type:"archetype-catalog.xml")` and is cached locally as `archetype-catalog-$repoID.xml` (w/ checksums) by Resolver automatically (this is how "metadata" is handled by resolver).

The subsequent step was to make `Metadata` be able to address **any (out of layout) file** on remote repository, or, in other words, make Maven able to get any file (and cache it, with all the whistle and bells of caching/policies/checksums) that are not addressable by GAV/artifact. This change was done in https://github.com/apache/maven-resolver/pull/1491 and it was needed for prefixes (again, to use same transport and all features of local repo like checksum validation/cache retention/etc and maven metadata tracking by origin). Basically prefix files lie on this path:

`Metadata(G:"", A:"", V:"", Type:".meta/prefixes.txt")` and is cached locally as `.meta/prefixes-$repoId.txt` (w/ checksums) by resolver automatically.

And currently, they are the ONLY Metadata files (in resolver and/or maven) that contain `/` (slash).

The point is, that archetype catalog, but also prefixes _are repository metadata resources_, and they need same checksum validation, cache retention as any other "metadata". And what happens is total reuse of all existing functionality of Resolver to get/cache these files (so `-U` also updates them, etc).

cstamas: One of the typical pain was long time index: indexes are also _published within repository_, [here](https://repo.maven.apache.org/maven2/.index/). And originally, Maven Indexer had to reimplement everything (full transport stack) to get it from same place, from where user was already getting artifacts from (with same possible proxies, mirrors, auth and all the cruft). Now, Resolver would be able even to get indexes for you, no need to reimplement anything, as any file, that is published via repository is reachable to Resolver and Maven (and plugins).

cstamas: And to answer your original question "we put the first part of the metadata in front"... due local repository layout. If I ask from Central following metadata:
* `Metadata(G:"", A:"", V:"", Type:"prefixes.txt")`
* `Metadata(G:"", A:"", V:"", Type:".meta/prefixes.txt")`
* `Metadata(G:"", A:"", V:"", Type:".meta/meta/prefixes.txt")`

They would call conflict, as Resolver handles Metadata file names in local repository by appending origin repoId, so they would be locally then be named `prefix-$repoId.txt` and clash.

Oh, but you are right, these are local repo paths, and we talk about _lock names_. Yes, you are right, we could replace `/` to `_` or anything to keep them "flat", agreed.

PR #1613, opened 2025-10-02
Advised and Enforced Management

cstamas:
Originally, dependency management marked a "subject" as "managed", it was 0/1 true/false only. This PR now adds ability to mark management as "advised" (may be subjected to further management) or "enforced" (equivalent of previously "managed").

Changes:
* `DependencyManager` applies management as "enforced", if they come from root (ie. project POM), otherwise as "advice" (this latter applies to transitive managers only, as for example classic manager have no data from "deeper levels").
* `ConflictManager` backs out for _enforced_ scope/optionality.
* This makes the "isInheritedDerived" hack in dependency manager not needed, as now management rule tells the modality of it (is enforced or just advised)


gnodet:
Any idea about memory/perfs implications of the switch to `Map`?

PR #1614, opened 2025-10-02
Target Platform Support

cstamas:
WIP

Builds upon
https://github.com/apache/maven-resolver/pull/1613


maven-resolver-api/src/main/java/org/eclipse/aether/collection/DependencyManagement.java
```diff
@@ -42,9 +47,27 @@ public final class DependencyManagement {
 
     /**
      * Creates an empty management update.
+     *
+     * @deprecated
```
slawekjaranowski, line 51: We many times missing description on deprecations .... why, what should be used instead of ... and so on

Issue #1644, opened 2025-10-30
Locking issues


cstamas:
### Affected version

master, 1.9.x

### Bug description

Seems that multiple remote repositories adds to locking issues, at least we see similar pattern between artifacts and metadata. Related issues:
* https://github.com/apache/maven-mvnd/issues/836
* https://github.com/apache/maven/issues/11116
* https://github.com/apache/maven/issues/9049

Also in resolver:
* https://github.com/apache/maven-resolver/issues/1641

PR #1655, opened 2025-11-14
Bug: Filter fixes

### LINKED ISSUE #1654: RRF auto discovery is broken (tried to circumvent locking issue)
### Affected version

2.0.13

### Bug description

RRF auto discovery contains hacks that needs to be undone. The hack was added to circumvent #1644 but just causes more issues. The origin issue needs fix instead.

### LINKED ISSUE #1667: RRF: make sure prefix discovery happens only once per session

This was the intent, but I may missed few edge cases...

cstamas:
Three major issues: 
* (prefix) one was that it tried to fix the Maven issue of remote repository uniqueness (globally), and that attempt was in fact wrong, fix is elsewhere. The reason it tried to do this is in fact to circumvent locking issues #1644.
* (manager) other issue was filter lifecycle, manager did it wrongly (data is _inherited on customized sessions_): the bug caused that one session was used to acquire filters and same filter got used for potentially other (maybe even reconfigured) session. This overlook forced filters to implement hoops and loops (to prevent recursion), but also prevented any third party code to have saying in filter operation.
* (both filters) make sure RemoteRepositories used as keys for prefixes are _same_ (normalized as bare)

Changes:
* fix manager to make sure filters and session are aligned
* fix/simplify prefix filter recursion prevention
* introduce public and documented way to inhibit prefix discovery

Note: as this PR now removes  the "hack" (that in fact tried to circumvent #1644) we are now back at state we were before https://github.com/apache/maven-resolver/pull/1575 (the "by make repositories unique" bit). As we see, this change (without locking fix) will cause Maven IT failures, as ITs are executed in parallel, and "prefix discovery" initiated over same local repository will/may cause locking conflicts and hence timeouts (result is sporadically failing ITs due locking timeouts). Hence, this fix is to be followed by fix for #1644 

Fixes #1654 
Fixes #1667

### DIFF
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/filter/RemoteRepositoryFilterSourceSupport.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/filter/RemoteRepositoryFilterSourceSupport.java
@@ -83,6 +83,13 @@ public abstract class RemoteRepositoryFilterSourceSupport implements RemoteRepos
      * used {@link org.eclipse.aether.RepositorySystem#newResolutionRepositories(RepositorySystemSession, List)} beforehand.
      * The hash/equalTo method factors in all these as well, but from our perspective, they do not matter. So we make all
      * key remote repositories back to "bare".
+     * Ignored properties of normalized repositories:
+     * <ul>
+     *     <li>proxy - is environment dependent</li>
+     *     <li>authentication - is environment and/or user dependent</li>
+     *     <li>mirrored repositories - is environment dependent (within same session does not change)</li>
+     *     <li>repository manager - is environment dependent (within same session does not change)</li>
+     * </ul>
      */
     protected RemoteRepository normalizeRemoteRepository(
             RepositorySystemSession session, RemoteRepository remoteRepository) {


maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/filter/PrefixesRemoteRepositoryFilterSource.java
```diff
@@ -273,6 +299,14 @@ private Path resolvePrefixesFromLocalConfiguration(
         }
     }
 
+    private boolean supportedResolvePrefixesForRemoteRepository(
+            RepositorySystemSession session, RemoteRepository remoteRepository) {
+        // TODO: RemoteRepository.isRepositoryManager() is still unused in Maven; once used, factor it in
```
gnodet: It may be a good idea to create an issue in Maven and reference it here, so that we can track more easily in the future.

cstamas: In fact, let me work thru this: despite nothing **sets** this `RemoteRepository` property, it _is queried regularly in resolver codebase_.

PR #1656, opened 2025-11-14
[1.9.x] Mild backport: support same properties as Resolver 2.x

cstamas:
Changes:
* fix manager to make sure filters and session are aligned
* support same (newly introduced) properties

This is "mild backport" of #1655

### DIFF
--- a/.github/workflows/maven-verify.yml
+++ b/.github/workflows/maven-verify.yml
@@ -26,7 +26,8 @@ jobs:
     name: Verify
     uses: apache/maven-gh-actions-shared/.github/workflows/maven-verify.yml@v4
     with:
+      ff-run: false
       ff-site-run: false
-      maven-matrix: '[ "3.9.9" ]'
-
-
+      maven4-enabled: true
+      verify-fail-fast: false
+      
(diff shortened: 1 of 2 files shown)


slawekjaranowski:
> maven-resolver-1.9.x is broken

I will look ... JDK 25 😄

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/filter/RemoteRepositoryFilterSourceSupport.java
```diff
@@ -73,7 +73,8 @@ protected String configPropKey(String name) {
      * Default is {@code false}.
      */
     protected boolean isEnabled(RepositorySystemSession session) {
-        return ConfigUtils.getBoolean(session, false, CONFIG_PROP_PREFIX + this.name);
+        return ConfigUtils.getBoolean(session, false, CONFIG_PROP_PREFIX + this.name)
+                && !ConfigUtils.getBoolean(session, false, CONFIG_PROP_PREFIX + this.name + ".skipped");
```
slawekjaranowski, line 77: should we documented new property 1.9.x ... or only introduce for compatibility with 2.x?

cstamas: Yes, this is "mild" backport and forward compatibility.

PR #1657, opened 2025-11-14
Bugfix: Java 25 broke test

cstamas:
As test was naively using `new File("")` assuming its method `file.list()` will return `null` on cleanup in `teardown()` method.

But, it turns out it was an old Java bug fixed in
Java 25, and once GH matrix got Java 25 introduced, it broke the build.

Refs:
* https://bugs.openjdk.org/browse/JDK-8024695
* https://inside.java/2025/11/03/quality-heads-up/

Thanks to @sormuras

### DIFF
--- a/.github/workflows/maven-verify.yml
+++ b/.github/workflows/maven-verify.yml
@@ -28,6 +28,6 @@ jobs:
     with:
       ff-run: false
       ff-site-run: false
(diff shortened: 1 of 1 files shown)


sormuras (approved):
Looks good to me.

PR #1658, opened 2025-11-14
Update GH CI

cstamas:
Update GH CI and pull in Java 25 as well.

PR #1659, opened 2025-11-14
Fix test DefaultArtifactResolverTest on JDK 25

slawekjaranowski:
use JUnit TempDir instead of manually managed


slawekjaranowski:
fixed in #1658

PR #1662, opened 2025-11-15
[1.9.x] Fix locking issues

cstamas:
Changes:
* introduce dedicated ex (backport from 2)
* up default timeouts (from 30s to 900s)
* improve message by listing all lock subjects
* improve message by mentioning the property that user should use to increase timeouts (there is no one size fits all; we could go with "infinite" timeouts. but am unsure about that)

### DIFF
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/synccontext/named/NamedLockFactoryAdapter.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/synccontext/named/NamedLockFactoryAdapter.java
@@ -218,7 +218,7 @@ public final class NamedLockFactoryAdapter {
                 }
             }
             if (!illegalStateExceptions.isEmpty()) {
-                String message = "Could not acquire " + (shared ? "read" : "write") + " lock for "
+                String message = "Could not acquire " + (shared ? "shared" : "exclusive") + " lock for "
                         + lockSubjects(artifacts, metadatas) + " in " + time + " " + timeUnit
                         + "; consider using '" + TIME_KEY
                         + "' property to increase lock timeout to a value that fits your environment";


maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/synccontext/named/NamedLockFactoryAdapter.java
```diff
@@ -216,12 +218,49 @@ public void acquire(Collection<? extends Artifact> artifacts, Collection<? exten
...
+            }
+            return builder.toString();
+        }
+
+        private String metadataSubjects(Metadata metadata) {
+            String name = "";
+            if (!metadata.getGroupId().isEmpty()) {
+                name += metadata.getGroupId();
+                if (!metadata.getArtifactId().isEmpty()) {
+                    name += "." + metadata.getArtifactId();
```
gnodet: `.` or `:` ?

cstamas: fixed

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/synccontext/named/NamedLockFactoryAdapter.java
```diff
@@ -216,12 +218,50 @@ public void acquire(Collection<? extends Artifact> artifacts, Collection<? exten
                 }
             }
             if (!illegalStateExceptions.isEmpty()) {
-                IllegalStateException ex = new IllegalStateException("Could not acquire lock(s)");
+                String message = "Could not acquire " + (shared ? "read" : "write") + " lock for "
```
cstamas: "shared" vs "exclusive" to align with 2.x

cstamas:
```suggestion
                String message = "Could not acquire " + (shared ? "shared" : "exclusive") + " lock for "
```

PR #1674, opened 2025-11-18
[1.9.x] Name mappers cleanup and new GAECV mapper

cstamas:
Cleanup name mappers usage and introduce new, more selective mapper GAECV.

Changes:
* introduce GAECV next to existing GAV name mapper.
* make default GAECV

### DIFF
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/synccontext/named/NameMappers.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/synccontext/named/NameMappers.java
@@ -67,6 +67,9 @@ public final class NameMappers {
         return gavNameMapper(false);
     }
 
+    /**
+     * @since 1.9.25
+     */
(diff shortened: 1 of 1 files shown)


maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/synccontext/named/NameMappers.java
```diff
@@ -46,11 +64,44 @@ public static NameMapper staticNameMapper() {
     }
 
     public static NameMapper gavNameMapper() {
-        return GAVNameMapper.gav();
+        return gavNameMapper(false);
+    }
+
+    public static NameMapper gavNameMapper(boolean fileSystemFriendly) {
```
gnodet, line 73: Missing `@since`

cstamas: fixed

Issue #996, opened 2023-02-15
[MRESOLVER-323] Investigate transport-http shared state pre-population


jira-importer:
**[Tamas Cservenak](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=cstamas)** opened **[MRESOLVER-323](https://issues.apache.org/jira/browse/MRESOLVER-323?redirect=false)** and commented

The issue MRESOLVER-319 in PR applies a band-aid: sends FIRST transport task fully thru the pipe, only to let the rest in multi threaded way.

Reason for this (as shown by IT MavenITmng4470AuthenticatedDeploymentToProxyTest) was concurrency issue how proxy authCache was populated.

There is either some concurrency issue in there, or, simpler would be to refactor transport-http in some way to "ping" remote peer, and as today, remember the results during session life span.

---

**Issue Links:**
- [MRESOLVER-319](https://issues.apache.org/jira/browse/MRESOLVER-319) Support parallel deploy

slachiewicz:
Resolved: the first-task-through-the-pipe shared-state hack in transport-http was removed by the auth caching rework (PR #1675 on 1.9.x / #1676 on master, followed by the simplification in #1790), which introduced SharingAuthCache and DeferredCredentialsProvider. On 2.x the transport-http module no longer exists at all, replaced by transport-jetty and transport-apache. Closing as fixed.

PR #1675, opened 2025-11-18
[1.9.x] Remove hack from Basic connector

cstamas:
There was an issue https://github.com/apache/maven-resolver/issues/996 that is still open and points to `MavenITmng4470AuthenticatedDeploymentToProxyTest` IT, so lets trigger it and lets fix it. OTOH, it may be very well already fixed...

Unrelated: fix for reproducer... (was wondering whole day why and what 11 things it resolved...)


gnodet:
I'd try to make `SharingAuthCache` more thread safe by using a `ConcurrentHashMap` instead of a `Map` to avoid possible issues.

PR #1679, opened 2025-11-19
Repository Key Function SPI

cstamas:
New opt-in **experimental** feature for enhanced local repository: ability to choose "repository key" function. Local repository (simple and enhanced) by default uses the `simple` key (historically) where other places all used `nid`. Have to note, that `simple` is technically equivalent to `nid` as `RemoteRepository.isRepositoryManager()` is not set anywhere in Maven.

Added key functions:
* `simple` -> the original code in LRM
* `nid` -> `norm(id)` (path friendly)
* `hurl` -> `sha1(url)`
* `nid_hurl` -> `norm(id)-sha1(url)`
* `gurk` -> `norm(id)-sha1(seed)` where "seed" is all config properties of repo
* `ngurk` -> `norm(id)-sha1(seed)` where "seed" is all config sans mirror list (just the string "isMirrored" is added for mirrors)

Introduced `RepositoryKeyFunction` type, that is used consistently in enhanced local repository (availability calculation), prefix composer (split repository prefix calculation) and remote repository manager (remote repository consolidation). This is new SPI, and impl provides default implementation with those above.

### DIFF
--- a/maven-resolver-util/src/test/java/org/eclipse/aether/util/repository/RepositoryIdHelperTest.java
+++ b/maven-resolver-util/src/test/java/org/eclipse/aether/util/repository/RepositoryIdHelperTest.java
@@ -18,13 +18,12 @@
  */
 package org.eclipse.aether.util.repository;
 
+import java.util.Collections;
+
 import org.eclipse.aether.repository.RemoteRepository;
 import org.eclipse.aether.repository.RepositoryKeyFunction;
 import org.eclipse.aether.repository.RepositoryPolicy;
 import org.junit.jupiter.api.Test;
-
-import java.util.Arrays;
-import java.util.Collections;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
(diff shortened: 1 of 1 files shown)


maven-resolver-api/src/main/java/org/eclipse/aether/repository/WorkspaceRepository.java
```diff
@@ -27,11 +28,14 @@
  * the contained artifacts is handled by a {@link WorkspaceReader}.
  */
 public final class WorkspaceRepository implements ArtifactRepository {
+    public static final String ID = "workspace";
 
     private final String type;
 
     private final Object key;
 
+    private final int hashCode;
```
gnodet, line 37: Caching hashCode could be seen as a premature optimisation.  They are more often kept in lists afaik, aren't they ?

cstamas: Yes, they are quite often in lists and collections. Maybe not as often as RemoteRepository instances, but I think this is okay to do.

PR #1691, opened 2025-11-26
Prevent FileNotFoundException when reading .lastUpdated tracking files

vladilm:
In our CI pipelines we are encountering a substantial number of exceptions like:

```
[main] [WARNING] Failed to read tracking file '/home/builduser/.m2/repository/cached/central-proxy/org/apache/maven/wagon/wagon-provider-api/1.0-alpha-6/wagon-provider-api-1.0-alpha-6.jar.lastUpdated'
java.io.FileNotFoundException: /home/builduser/.m2/repository/cached/central-proxy/org/apache/maven/wagon/wagon-provider-api/1.0-alpha-6/wagon-provider-api-1.0-alpha-6.jar.lastUpdated (No such file or directory)
    at java.io.FileInputStream.open0 (Native Method)
    at java.io.FileInputStream.open (FileInputStream.java:219)
    at java.io.FileInputStream.<init> (FileInputStream.java:157)
    at org.eclipse.aether.internal.impl.DefaultTrackingFileManager.read (DefaultTrackingFileManager.java:58)
    ...
```

These errors occur when a `.lastUpdated` tracking file is missing during concurrent repository access, leading the build to fail even though the file absence is harmless.

This patch introduces a safe fallback by catching `FileNotFoundException` in `DefaultTrackingFileManager.read()`. Missing or concurrently removed tracking files are now treated as non-fatal, preventing unnecessary build failures.

### DIFF
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultTrackingFileManager.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultTrackingFileManager.java
@@ -25,6 +25,7 @@ import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.FileInputStream;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.RandomAccessFile;
 import java.io.UncheckedIOException;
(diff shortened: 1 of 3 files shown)


cstamas:
Thanks!
Any particular reason why 3.9.6? Since then there is 3.9.11 with a TON of fixed bugs...

bade7n:
No specific reason, I can upgrade the version to 3.9.11 if you say that.
I might be wrong but I haven't found anything what could fix that issue in recent commits of 3.9.x.

bade7n:
Ignoring the FNF would resolve the issue for us.

Should I remove the tests from this PR and keep only the fix, or should this change go into master/some other branch instead?

cstamas:
Pls take a peek at https://github.com/apache/maven-resolver/pull/1692

IMO, this should fix your issue; I will backport this fix from master to 1.9.x as well, so both, Maven 4-rc-6 and Maven 3.9.12 should have it.

PR #1692, opened 2025-11-26
TrackingFileManager changes

cstamas:
Drop out of band deletion, let TFM delete as well. Also, simplify but do not drop locking, is needed for interoperability with 3.9.11 and 4.0.0-rc-5 and older released versions. Make FNFEx ignored in case of read/delete.

Backport to 1.9.x is here https://github.com/apache/maven-resolver/pull/1695.

### DIFF
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultTrackingFileManager.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultTrackingFileManager.java
@@ -93,11 +93,11 @@ public final class DefaultTrackingFileManager implements TrackingFileManager {
             LOGGER.warn("Failed to create tracking file parent '{}'", path, e);
             throw new UncheckedIOException(e);
         }
-        Properties props = new Properties();
         synchronized (mutex(path)) {
             try (FileChannel fileChannel = FileChannel.open(
                             path, StandardOpenOption.READ, StandardOpenOption.WRITE, StandardOpenOption.CREATE);
                     FileLock unused = fileLock(fileChannel, false)) {
+                Properties props = new Properties();
                 if (fileChannel.size() > 0) {
                     props.load(Channels.newInputStream(fileChannel));
                 }
@@ -119,12 +119,12 @@ public final class DefaultTrackingFileManager implements TrackingFileManager {
                 fileChannel.position(0);
                 int written = fileChannel.write(ByteBuffer.wrap(stream.toByteArray()));
                 fileChannel.truncate(written);
+                return props;
             } catch (IOException e) {
                 LOGGER.warn("Failed to write tracking file '{}'", path, e);
                 throw new UncheckedIOException(e);
             }
         }
-        return props;
     }
 
     @Override


cstamas:
@bade7n ping

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultTrackingFileManager.java
```diff
@@ -61,15 +62,14 @@ public Properties read(File file) {
...
-                    long fileSize = Files.size(path);
-                    try (FileChannel fileChannel = FileChannel.open(path, StandardOpenOption.READ);
-                            FileLock unused = fileLock(fileChannel, Math.max(1, fileSize), true)) {
-                        Properties props = new Properties();
-                        props.load(Channels.newInputStream(fileChannel));
-                        return props;
-                    }
+            synchronized (mutex(path)) {
+                try (FileChannel fileChannel = FileChannel.open(path, StandardOpenOption.READ);
+                        FileLock unused = fileLock(fileChannel, Math.max(1, fileChannel.size()), true)) {
```
gnodet: what's the purpose of using `fileLock(fileChannel, x)` ? Why not using `fileLock(fileChannel, 0)` ?
According to the javadoc: 
> A value of zero means to lock all bytes from the specified starting position to the end of the file, regardless of whether the file is subsequently extended or truncated

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultTrackingFileManager.java
```diff
@@ -87,64 +87,71 @@ public Properties update(File file, Map<String, String> updates) {
...
-                try {
-                    fileSize = Files.size(path);
-                } catch (IOException e) {
-                    fileSize = 0L;
+        Properties props = new Properties();
+        synchronized (mutex(path)) {
+            try (FileChannel fileChannel = FileChannel.open(
+                            path, StandardOpenOption.READ, StandardOpenOption.WRITE, StandardOpenOption.CREATE);
+                    FileLock unused = fileLock(fileChannel, Math.max(1, fileChannel.size()), false)) {
+                if (fileChannel.size() > 0) {
```
gnodet, line 101: why is there a check here and not while reading ?

cstamas: Reading will be unable to open absent file, here we may create it (and will be 0 byte)

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultTrackingFileManager.java
cstamas: Basically we lock with `size` 0 or 1 (so all or [0..1] bytes), based on file was just created or existed.

gnodet: Yes, that's what the code does.  I don't understand why.  What we want is a lock on the full file, so why just requesting 1 byte instead of the whole length (which will be extended automatically when written) ?
This parameter seems useless to me.

gnodet: And it forces to add a call to retrieve the file size...

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultTrackingFileManager.java
```diff
@@ -87,72 +87,79 @@ public Properties update(File file, Map<String, String> updates) {
...
         }
-
-        synchronized (getMutex(path)) {
-            try {
-                long fileSize;
-                try {
-                    fileSize = Files.size(path);
-                } catch (IOException e) {
-                    fileSize = 0L;
+        Properties props = new Properties();
```
gnodet: The `props` variable should be defined inside the `try` block, with the `return` statement moved at the end of the same block.

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultTrackingFileManager.java
cstamas: I think you looked at outdated PR? As this morning I removed the 0/1 dance fully (and hence the file size call as well)

PR #1696, opened 2025-11-28
Locking inhibitor SPI

### LINKED ISSUE #1641: Lock congestion on metadata (ie prefix files)
### Affected version

All 2.0.x

### Bug description

There is a lock congestion happening when prefix files are being discovered.

Problem: as many repositories are in play, all will be asking for `.meta/prefixes.txt` metadata, and named locks does not have "context" like repository is (see https://github.com/apache/maven-resolver/pull/1610 )

This is most visible in parallely executed Maven ITs:
```
Error:  Error executing Maven.
Error:  Could not acquire shared lock for [.meta-SLASH-prefixes.txt] using lock file:///D:/a/maven/maven/its/core-it-suite/target/user-home/.m2/repository/.locks/metadata~.meta-SLASH-prefixes.txt.lock in 30 SECONDS
```

### LINKED ISSUE #1663: Prefix file discovery locking conflicts
### Affected version

2.0.13

### Bug description

Prefix file discovery conflicts, when multiple threads/processes are about to discover prefix files (ie parallel tests)

cstamas:
A new SPI that offers ability to augment Resolver locking and inhibit locking on certain resources.

Out of the box RRF prefix file inhibition is provided.

Fixes #1663
Fixes #1641


maven-resolver-spi/src/main/java/org/eclipse/aether/spi/locking/LockingInhibitor.java
```diff
@@ -0,0 +1,47 @@
...
+ * Locking inhibitor may prevent Resolver locking to happen on certain resources.
+ *
+ * @since 2.0.14
+ */
+public interface LockingInhibitor {
+    /**
+     * May provide a predicate for artifacts that needs lock inhibition. <em>Warning: you do not want to override
+     * this method, or if you do, think twice.</em>
+     */
+    default Optional<Predicate<Artifact>> inhibitArtifactLocking() {
```
gnodet: That looks over engineered.
Why not a simple
```
    default boolean inhibitLock(Artifact artifact) {
        return false;
    }

    default boolean inhibitLock(Metadata metadata) {
        return false;
    }
```

gnodet: or `preventLock`... ?

cstamas: Simplified

PR #1709, opened 2025-12-11
Enhance RRF

cstamas:
Changes:
* prefix: introduce "resolvePrefixFiles" config that controls auto-discovery (def: true; as before)
* both: introduce "noInputOutcome" config that controls filtering outcome when enabled but no input provided (def: true; as before)
* both: cleanup of configuration and use of helper classes

### DIFF
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultArtifactResolver.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultArtifactResolver.java
@@ -312,9 +312,10 @@ public class DefaultArtifactResolver implements ArtifactResolver {
                             new LocalArtifactRequest(
                                     artifact, filteredRemoteRepositories, request.getRequestContext()));
                     result.setLocalArtifactResult(local);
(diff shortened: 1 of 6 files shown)


maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/filter/GroupIdRemoteRepositoryFilterSource.java
```diff
@@ -125,14 +122,31 @@ public final class GroupIdRemoteRepositoryFilterSource extends RemoteRepositoryF
 
     public static final boolean DEFAULT_SKIPPED = false;
 
+    /**
+     * The "no input outcome": if filter is enabled, but no input is available. By default, RRF filters without
+     * input would "stand aside" and just "allow", resulting in same outcome for "disabled" and "enabled but no input
+     * available" situations. This configuration allows to alter outcome for "enabled but no input available" case.
```
gnodet:
```suggestion
     * Determines what happens when the filter is enabled but has no prefix data to work with. 
     * When set to `true` (default), the filter allows all requests to proceed. When set to `false`, 
     * the filter blocks requests when no prefix data is available. 
```

PR #1752, opened 2026-01-15
Expose SeekableByteChannel from PutTask

### LINKED ISSUE #1751: Resolver SPI: Expose SeekableByteChannel from PutTask

Currently a PUT tasks either receives the artifact as [byte array](https://github.com/apache/maven-resolver/blob/2340c676f3aaf920ea7bbb5120b0ade4829a737c/maven-resolver-spi/src/main/java/org/eclipse/aether/spi/connector/transport/PutTask.java#L135) or as [Path](https://github.com/apache/maven-resolver/blob/2340c676f3aaf920ea7bbb5120b0ade4829a737c/maven-resolver-spi/src/main/java/org/eclipse/aether/spi/connector/transport/PutTask.java#L123) but exposes it for both only as [InputStream](https://github.com/apache/maven-resolver/blob/2340c676f3aaf920ea7bbb5120b0ade4829a737c/maven-resolver-spi/src/main/java/org/eclipse/aether/spi/connector/transport/PutTask.java#L58). Although there is an individual method to get the [Path](https://github.com/apache/maven-resolver/blob/2340c676f3aaf920ea7bbb5120b0ade4829a737c/maven-resolver-spi/src/main/java/org/eclipse/aether/spi/connector/transport/PutTask.java#L58) there is none which exposes the byte array.

In order to implement retries without being forced to copy to a temp file (like in https://github.com/apache/maven-resolver/blob/2340c676f3aaf920ea7bbb5120b0ade4829a737c/maven-resolver-transport-jdk-parent/maven-resolver-transport-jdk11/src/main/java/org/eclipse/aether/transport/jdk/JdkTransporter.java#L439) it is important to expose an object which can be seeked to beginning. It would make sense to expose `SeekableByteChannel newSeekableByteChannel()` which would work equally well on top of both byte array and path.

kwin:
This closes #1751

### DIFF
--- a/maven-resolver-spi/src/main/java/org/eclipse/aether/spi/connector/transport/PutTask.java
+++ b/maven-resolver-spi/src/main/java/org/eclipse/aether/spi/connector/transport/PutTask.java
@@ -58,9 +58,8 @@ public final class PutTask extends TransportTask {
      *
      * @return The input stream for the data, never {@code null}. The stream is unbuffered.
      * @throws IOException If the stream could not be opened.
-     * @deprecated Use {@link #newByteChannel()} instead.
+     * @see #newByteChannel()
      */
-    @Deprecated
     public InputStream newInputStream() throws IOException {
         if (dataPath != null) {
             return Files.newInputStream(dataPath);
@@ -71,7 +70,7 @@ public final class PutTask extends TransportTask {
     /**
      * Opens a seekable byte channel for the data to be uploaded. The length of the channel can be queried via
      * {@link #getDataLength()}. It's the responsibility of the caller to close the provided channel.
-     * Write operations on the returned channel will throw {@link java.nio.channels.NonWritableChannelException}.
+     * Write operations are not supposed to be performed on the returned channel and will have undefined behaviour.
      *
      * @return The seekable byte channel for the data, never {@code null}.
      * @throws IOException If the channel could not be opened.
@@ -114,12 +113,12 @@ public final class PutTask extends TransportTask {
 
     /**
      * Gets the file (if any) with the data to be uploaded.
+     * Consider using the data agnostic {@link #newInputStream()} or {@link #newByteChannel()} methods to access the data to not deal with {@code null} return values.
      *
      * @return The data file or {@code null} if the data resides in memory.
      * @since 2.0.0
-     * @deprecated Use {@link #newByteChannel()} instead.
+     * @see #getDataBytes()
      */
-    @Deprecated
     public Path getDataPath() {
         return dataPath;
     }
@@ -149,6 +148,18 @@ public final class PutTask extends TransportTask {
         this.dataPath = dataPath;
         dataBytes = EMPTY;
         return this;
+    }
+
+    /**
+     * Gets the binary data to be uploaded.
+     * Consider using the data agnostic {@link #newInputStream()} or {@link #newByteChannel()} methods to access the data to not deal with {@code null} return values.
+     *
+     * @return The binary data, may be {@code null} if the data resides in a file.
+     * @since 2.1.0
+     * @see #getDataPath()
+     */
+    public byte[] getDataBytes() {
+        return dataBytes;
     }
 
     /**


maven-resolver-spi/pom.xml
```diff
@@ -47,6 +47,11 @@
       <groupId>com.google.code.gson</groupId>
       <artifactId>gson</artifactId>
     </dependency>
+    <dependency>
```
kwin, line 50: @cstamas I don't like this additional dependency but unfortunately there is no impl in Java for a SeekableByteChannel on top of a byte array...

maven-resolver-spi/pom.xml
```diff
@@ -47,6 +47,11 @@
...
+      <groupId>commons-io</groupId>
+      <artifactId>commons-io</artifactId>
+      <version>2.21.0</version>
```
cstamas, line 53: Pull up, as we already have commons-io as dependency somewhere, so let them have aligned. Same for commons-codec, so even if this PR goes in as-is (which is fine) we need another pass to pull these up.

maven-resolver-spi/src/main/java/org/eclipse/aether/spi/connector/transport/PutTask.java
```diff
@@ -54,14 +58,33 @@ public PutTask(URI location) {
...
+     * Opens a seekable byte channel for the data to be uploaded. The length of the channel can be queried via
+     * {@link #getDataLength()}. It's the responsibility of the caller to close the provided channel.
+     * Write operations on the returned channel will throw {@link java.nio.channels.NonWritableChannelException}.
+     *
+     * @return The seekable byte channel for the data, never {@code null}.
+     * @throws IOException If the channel could not be opened.
+     * @since 2.1.0
+     * @see #getInputStream()
+     */
+    public SeekableByteChannel newByteChannel() throws IOException {
```
cstamas, line 80: And why not just expose the `byte[]`? And leave to caller to do with it what it wants? Just like `Path` is exposed, and payload is one of those two...

kwin: In order to not make the caller distinguish between the two types.

kwin: I am fine though with exposing the byte array through a dedicated getter for some edge cases as well

maven-resolver-spi/pom.xml
kwin: I couldn't find any other usage of commons-io yet.

cstamas: In my IDE it shows 

Image

So probably not direct deps, but we may want to pull them up (and also figure out what commons-logging does here at all)

maven-resolver-spi/src/main/java/org/eclipse/aether/spi/connector/transport/PutTask.java
cstamas: with byte array exposed this method is not needed, is it?

maven-resolver-spi/pom.xml
kwin: Right, transitive dep of commons-codec: https://commons.apache.org/proper/commons-codec/dependencies.html

maven-resolver-spi/src/main/java/org/eclipse/aether/spi/connector/transport/PutTask.java
kwin: I still think this is beneficial as only newInputStream() and newByteChannel() can be used no matter how the data has been provided.

cstamas: I just find it awkward to pull in whole library for this one single class....

kwin: ok, since my use case for having that is gone, I am fine with not merging this right now.

PR #1755, opened 2026-01-16
JDK Transport: Do no longer leverage temp file for transfering artifact

kwin:
with PUT

Directly transfer based on the PutTask's InputStream. Improve retry handling to not retry (some) BodyPublisher exceptions.

### DIFF
--- a/maven-resolver-spi/src/test/java/org/eclipse/aether/spi/connector/transport/TransportListenerNotifyingInputStreamTest.java
+++ b/maven-resolver-spi/src/test/java/org/eclipse/aether/spi/connector/transport/TransportListenerNotifyingInputStreamTest.java
@@ -1,0 +1,252 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.eclipse.aether.spi.connector.transport;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InterruptedIOException;
+import java.nio.ByteBuffer;
+
+import org.eclipse.aether.transfer.TransferCancelledException;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * Tests for {@link TransportListenerNotifyingInputStream}.
+ */
+class TransportListenerNotifyingInputStreamTest {
+
+    private static final byte[] TEST_DATA = new byte[] {1, 2, 3, 4, 5, 6, 7, 8, 9, 10};
+    private static final long TEST_SIZE = TEST_DATA.length;
+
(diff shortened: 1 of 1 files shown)


maven-resolver-test-http/src/main/java/org/eclipse/aether/internal/test/util/http/HttpTransporterTest.java
```diff
@@ -991,7 +992,7 @@ protected void testPut_ProxyAuthenticated() throws Exception {
         transporter.put(task);
         assertEquals(0L, listener.getDataOffset());
         assertEquals(6L, listener.getDataLength());
-        assertEquals(1, listener.getStartedCount());
+        assertTrue(listener.getStartedCount() > 0 && listener.getStartedCount() <= 2, "Started count: " + listener.getStartedCount());
```
kwin: There are some more tests which are strict with regard to started count. In general depending on the HTTP client it may be started multiple times (due to internal retries). Can we also relax the other assertions @cstamas?

kwin: Not sure yet why they only fail with Java < 25...

cstamas: That was the exact reason why it used tmp file, I wanted higher level APIs (ie listeners most importantly) protect from low level stuff, and make them behave _in same way_ (re emitted events).

kwin: But that way you don't get real results, e.g. progress reported on the listener is not upload progress but copy progress to the temp file. I consider this worse than having different values for the started count. Not sure what the started count is used for potentially by consumers, but obviously retries may differ i.e. it is expected to have different numbers depending on the transport.

kwin: @cstamas Are you fine with relaxing the other ITs assertions with respect to num started counts?

cstamas: Yup, is fine.

maven-resolver-spi/src/main/java/org/eclipse/aether/spi/connector/transport/TransportListenerNotifyingInputStream.java
```diff
@@ -0,0 +1,98 @@
...
+    @Override
+    public int read(byte[] b) throws IOException {
+        int numBytesRead = super.read(b);
+        if (numBytesRead != -1) {
+            if (!isStarted) {
+                notifyStarted();
+            }
+            notifyProgress(b, 0, numBytesRead);
+        }
+        return super.read(b);
```
cstamas: ? is invoked 2 times, no?

kwin: Good catch, gonna fix this

kwin: Done in https://github.com/apache/maven-resolver/pull/1755/commits/dce23de6a12c2cbc84c93c4da2d0afbac18bbca1.

cstamas: Looks good, but one thing bothers me: how did double invocation slipped thru UT? Why did not it fail? We obviously don't test this path? Or we don't validate what we got?

kwin: Usually only the read with 3 arguments is used (e.g. when copying to an output stream). Would require dedicated unit tests to cover this

kwin: Added a unit test with a little help from Claude.

PR #1758, opened 2026-01-25
Add CycleAwareDependencyGraphDumper to visualize cycles (resolves #1561)

### LINKED ISSUE #1561: Create a dependency visitor that can visualize cycles

Right now we have no dependency visitor (like DependencyGraphDumper is) that is able to visualize cycles. This means that graphs in FULL verbosity mode cannot be used to be visualized (all visitors will fail with StackOverflow).

Jimin15:

## Summary
This PR adds `CycleAwareDependencyGraphDumper`, a new dependency visitor that visualizes cycles in dependency graphs while preventing StackOverflow errors. This addresses issue #1561.

## What
Added `CycleAwareDependencyGraphDumper` class that wraps `DependencyGraphDumper` and adds cycle detection capabilities- When a cycle is detected, it displays the cycle with `^N` notation where N is the index of the node in the path that it cycles back to- The visitor stops traversing children of cycle nodes to prevent infinite recursion

## Why
Currently, `DependencyGraphDumper` fails with `StackOverflowError` when visualizing dependency graphs in FULL verbosity mode that contain cycles. While `TreeDependencyVisitor` can prevent the error, it doesn't visualize cycles. This new visitor provides both cycle visualization and StackOverflow prevention.

## How
The implementation uses a `Deque` to track the current path during traversal- Cycle detection is performed by comparing versionless artifact IDs using `ArtifactIdUtils.equalsVersionlessId()`- When a cycle is detected, custom indentation formatting is applied to match the tree structure- The visitor delegates to `DependencyGraphDumper` for normal nodes and handles cycle nodes separately
## Testing
Added comprehensive unit tests in `CycleAwareDependencyGraphDumperTest`- Tests verify: cycle detection, visualization format, StackOverflow prevention, and compatibility with non-cycle graphs- All tests pass successfully

## Related Issue
Resolves #1561
- [x] You have run the integration tests successfully (`mvn -Prun-its verify`).

### DIFF
--- a/maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/visitor/CycleAwareDependencyGraphDumper.java
+++ b/maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/visitor/CycleAwareDependencyGraphDumper.java
@@ -1,0 +1,240 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.eclipse.aether.util.graph.visitor;
+
+import java.util.ArrayDeque;
+import java.util.Collection;
+import java.util.Deque;
+import java.util.Iterator;
+import java.util.List;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+import org.eclipse.aether.artifact.Artifact;
+import org.eclipse.aether.graph.DependencyNode;
+import org.eclipse.aether.graph.DependencyVisitor;
+import org.eclipse.aether.util.artifact.ArtifactIdUtils;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * A dependency visitor that visualizes cycles in dependency graphs while preventing StackOverflow errors.
+ * This visitor wraps a {@link DependencyGraphDumper} and adds cycle detection and visualization capabilities.
+ * <p>
+ * When a cycle is detected (a node with the same versionless artifact ID appears again in the current path),
+ * it is displayed with a reference notation like {@code ^N} where N is the index of the node in the path that
+ * it cycles back to. The visitor then stops traversing children of the cycle node to prevent infinite recursion.
+ * </p>
+ * <p>
+ * This visitor is particularly useful for visualizing dependency graphs in FULL verbosity mode where cycles
+ * are preserved in the graph structure.
+ * </p>
+ *
+ * @since 2.0.0
+ */
+public class CycleAwareDependencyGraphDumper implements DependencyVisitor {
+
+    private final Consumer<String> consumer;
+    private final DependencyGraphDumper dumper;
+    private final Deque<DependencyNode> currentPath;
+    private final Stack<Boolean> isCycleStack;
+
+    /**
+     * Creates a new cycle-aware dependency graph dumper with the specified consumer.
+     *
+     * @param consumer the string consumer, must not be {@code null}
+     */
+    public CycleAwareDependencyGraphDumper(Consumer<String> consumer) {
+        this.consumer = requireNonNull(consumer, "consumer cannot be null");
+        this.dumper = new DependencyGraphDumper(consumer);
+        this.currentPath = new ArrayDeque<>();
+        this.isCycleStack = new Stack<>();
+    }
+
+    /**
+     * Creates a new cycle-aware dependency graph dumper with the specified consumer and decorators.
+     *
+     * @param consumer the string consumer, must not be {@code null}
+     * @param decorators the decorators to apply, must not be {@code null}
+     */
+    public CycleAwareDependencyGraphDumper(
+            Consumer<String> consumer, Collection<Function<DependencyNode, String>> decorators) {
+        this.consumer = requireNonNull(consumer, "consumer cannot be null");
+        this.dumper = new DependencyGraphDumper(consumer, decorators);
+        this.currentPath = new ArrayDeque<>();
+        this.isCycleStack = new Stack<>();
+    }
+
+    @Override
+    public boolean visitEnter(DependencyNode node) {
+        requireNonNull(node, "node cannot be null");
+
+        // Check for cycle BEFORE adding current node to path
+        int cycleIndex = findCycleInPath(node);
+        boolean isCycle = cycleIndex >= 0;
+
+        // Add node to path after checking (for formatting purposes)
+        currentPath.push(node);
+        isCycleStack.push(isCycle);
+
+        if (isCycle) {
+            // Format the cycle line with reference notation
+            // Use custom formatting for cycle nodes since they might be references
+            String indentation = formatCycleIndentation(currentPath);
+            String nodeStr = dumper.formatNode(currentPath);
+            String line = indentation + nodeStr + " ^" + cycleIndex;
+            consumer.accept(line);
+            return false; // Stop traversing children to prevent infinite recursion
+        }
+
+        // Delegate to the wrapped dumper for normal nodes
+        return dumper.visitEnter(node);
+    }
+
+    @Override
+    public boolean visitLeave(DependencyNode node) {
+        if (!currentPath.isEmpty() && currentPath.peek() == node) {
+            currentPath.pop();
+        }
+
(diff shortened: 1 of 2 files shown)


gnodet (commented):
## AI Review — PR #1758: Add CycleAwareDependencyGraphDumper

Hi @Jimin15, thanks for tackling cycle visualization in dependency graphs — this addresses a real need (issue #1561)!

After review and independent verification, a few issues stood out:

### ⚠️ CI is red — needs rebase (high)
All 12 Verify jobs fail. The PR branch is significantly behind master. A local build/test run of the PR code itself passes (413 tests, 0 failures), so the failure is likely caused by the stale branch base rather than your code. A rebase onto current master should fix CI.

### ⚠️ Cycle index semantics are misleading (medium)
`findCycleInPath` returns an index where 0 is the **parent** (most recent ancestor), not the root — because `ArrayDeque.push` adds to the front. The code comment on line 144 says "0-based, root is 0" which is incorrect. The `^N` notation in output is internally consistent but could confuse users who expect a root-relative index. The test assertion (`cycleIndex >= 0 && cycleIndex < output.size()`) is also too weak to validate semantic correctness of the index.

### ⚠️ `isLastChild` false positive risk (medium)
The `equalsVersionlessId` fallback in `isLastChild` compares groupId:artifactId:extension:classifier but **not version**. If siblings share the same GA but differ in version (plausible in FULL verbosity mode), `isLastChild` could return true for a non-last child, producing incorrect tree indentation (`\-` instead of `+-`).

### 💡 Design: composition with tight coupling (medium)
The class uses composition (`new DependencyGraphDumper(consumer)`) but calls `protected` methods (`dumper.formatNode`) and relies on knowledge of internal deque semantics. This works because both classes are in the same package, but it's fragile — if `formatNode` were changed to reference internal state beyond `peek()`, this would break. Consider inheritance or a cleaner decorator pattern.

**What works well:**
- The core cycle detection logic is correct — the verifier confirmed that `formatNode(currentPath)` correctly peeks the cycle node
- The enter/leave delegation is properly paired (both skipped for cycle nodes, both called for non-cycle nodes)
- Test coverage for basic scenarios is solid

---
*🤖 This review was generated by [ForgeBot](https://github.com/gnodet/forgebot-skills) using a maker/checker pattern (reviewer + independent verifier). 2 of 6 original findings were removed as false positives after verification.*

gnodet (commented):
## Review Summary

Thanks for working on this, @Jimin15! A cycle-aware graph dumper is a useful addition for debugging dependency resolution. The composition-based approach wrapping `DependencyGraphDumper` is a reasonable design choice.

After review and independent verification, I found several confirmed issues:

### Medium Severity

1. **Misleading cycle index semantics** — The comment on `findCycleInPath` says "0-based, root is 0" but `ArrayDeque.push()` adds to the head, and the for-each iterator starts from the head. So index 0 actually corresponds to the most recently pushed node (direct parent), not the root. The `^N` notation in output will show an index relative to the top of the stack, which may confuse users.

2. **`isLastChild` versionless ID fallback** — The `equalsVersionlessId()` fallback ignores version, so if two siblings share the same `groupId:artifactId` but differ in version (possible in FULL verbosity mode), the wrong sibling could match, producing incorrect tree indentation.

3. **Incorrect `@since 2.0.0` tag** — Version 2.0.0 was released long ago (latest is 2.0.20). Recently added classes use accurate version tags like `@since 2.0.19`. This should be updated to the next release version.

4. **Weak test assertion in `cycleReferencePointsToCorrectIndex`** — The test checks `cycleIndex >= 0 && cycleIndex < output.size()`, comparing a path-depth index against the number of output lines. These are unrelated quantities, so the test would pass even with an incorrect cycle index as long as it's a small positive number.

### Low Severity

5. **Test simplification** — The element-by-element loop comparison in `dumpSimple` could be simplified to `assertEquals(standardOutput, cycleAwareOutput)` for clearer failure messages.

6. **Missing test coverage** — The two-argument constructor accepting decorators has no test coverage.

### Notes

- The reviewer initially flagged a potential "stack desynchronization bug" in `visitLeave`, but independent verification confirmed this is **not a bug** — the `DependencyVisitor` contract guarantees matching `visitEnter`/`visitLeave` pairs, so both stacks remain synchronized.
- The composition approach using `protected` methods from `DependencyGraphDumper` is a valid design choice since both classes are co-located in the same package.

---

gnodet (commented):
Thanks for working on cycle visualization — this addresses a real need (issue #1561). The implementation has a few issues that should be addressed before merge.

**Correctness bug:**
- `formatCycleIndentation` (line 172) uses identity comparison `child == cycleNode` to determine tree-drawing characters. In cycle graphs, `DependencyGraphParser` reuses the same `DependencyNode` Java object for `^N` references, so the cycle node IS the same object as an earlier ancestor in the path. This causes `end=true` prematurely at the wrong depth, producing incorrect indentation (e.g., `+- ` instead of `|  `). Fix: replace with a positional check like `depth == path.size() - 2`.

**Other issues:**
- `isLastChild` versionless ID fallback via `equalsVersionlessId()` (line 205) ignores version. In FULL verbosity mode, siblings with the same `groupId:artifactId` but different versions would be misidentified as the last child, producing incorrect `\-` instead of `+-`.
- The comment "0-based, root is 0" in `findCycleInPath` (line 144) is incorrect — `ArrayDeque.push()` adds to the head, so index 0 is the direct parent, not the root.
- `@since 2.0.0` should be `@since 2.0.22` (2.0.0 was released long ago; current dev is 2.0.22-SNAPSHOT).
- Test `cycleReferencePointsToCorrectIndex` compares a path-depth index against `output.size()` (line count) — these are unrelated quantities, so the assertion is vacuously true. Tests should validate against expected exact output strings.
- All 12 CI Verify jobs are failing — a rebase onto current master is needed.
- The two-argument constructor accepting decorators has no test coverage.

gnodet (changes requested):
The cycle-aware graph dumper has a correctness bug and several issues that should be addressed:

**High severity:**
- 🐛 **Correctness bug** in `formatCycleIndentation` (line 183): The identity comparison `child == cycleNode` produces incorrect indentation when the same Java object appears at multiple positions in the path (the normal case for cycle graphs). **Fix**: Replace `boolean end = child == cycleNode;` with `boolean end = !iter.hasNext();`.

**Medium severity:**
- `@since 2.0.0` tag is incorrect — the current development version is `2.0.22-SNAPSHOT`. Should be `@since 2.0.22`.
- Comment at line 154 says "0-based, root is 0" but `ArrayDeque`'s for-each iterator goes from head (most recently pushed = direct parent) to tail, so index 0 is the direct parent, not the root. The comment is factually wrong.
- `isLastChild` versionless ID fallback via `equalsVersionlessId()` ignores version. In FULL verbosity mode, siblings with the same `groupId:artifactId` but different versions would be misidentified as the last child.
- `cycleReferencePointsToCorrectIndex` test compares `cycleIndex` (path-depth index) against `output.size()` (total line count) — these are unrelated quantities, making the assertion vacuously true.

**Low severity:**
- Element-by-element loop comparison in `dumpSimple` can be simplified to `assertEquals(standardOutput, cycleAwareOutput)` for clearer failure messages.
- Two-argument constructor accepting decorators has no test coverage.
- `dumpCycles` test only checks pattern match, not exact output format. A test with expected output strings would have caught the indentation bug.

gnodet (commented):
The PR adds valuable cycle visualization for dependency graphs (issue #1561). However, there are some issues that should be addressed:

### Identity-comparison bug in `formatCycleIndentation`

When `DependencyGraphParser` creates a cycle via `^N`, it reuses the same Java object instance. In `formatCycleIndentation`, the descending iterator traverses from root upward, and when it encounters the original node at an earlier depth, `child == cycleNode` is `TRUE` because they are the same object instance. This causes `end=true` prematurely, producing garbled indentation. The fix requires comparing by index position in the deque rather than object identity.

### `@since` tag

The `@since 2.0.0` tag should be `@since 2.0.22` (the project version is `2.0.22-SNAPSHOT` and existing classes follow this convention, e.g. `@since 2.0.12`, `@since 2.0.17`).

### Weak test assertions

The `dumpCycles` test only asserts that some output line matches `.*\\^\\d+.*`, and `cycleReferencePointsToCorrectIndex` asserts `cycleIndex >= 0 && cycleIndex < output.size()` which is trivially true for any small non-negative path depth. Neither test verifies correct indentation or that the right nodes are marked as cycles — they would pass even with the garbled output from the identity-comparison bug.

### Design: extend vs. wrap

`DependencyGraphDumper` exposes `formatIndentation` and `formatNode` as `protected` methods. Extending it (rather than wrapping with a parallel `currentPath` deque) would eliminate the dual-state management, reduce the code duplication in `formatCycleIndentation`, and avoid the identity-comparison bug by reusing the parent's `nodes` deque.

### `isLastChild` versionless comparison

The `ArtifactIdUtils.equalsVersionlessId` fallback in `isLastChild` ignores versions — two children with the same `groupId:artifactId:extension:classifier` but different versions would both match, potentially producing incorrect indentation. The original `DependencyGraphDumper.formatIndentation` uses only identity comparison.

slachiewicz:
The last CI run on this branch was 2026-01-27, and GitHub has deleted the logs since, so there's no way to tell whether the change still builds against master.

A rebase triggers a fresh run and makes the state readable again. If you'd rather not carry it forward, closing is equally useful — the PR isn't being ignored, it's unreadable.

*This comment was created with AI assistance.*

slachiewicz:
Closing due to inactivity following review feedback on cyclic node formatting. The feature request remains tracked under Issue #1561.

PR #1771, opened 2026-01-28
Expose transporter details per protocol

### LINKED ISSUE #1770: Expose transporter details per protocol

Currently it is impossible to tell in advance which transporter implementation is used for which kind of (remote) repository resolver/upload request. In order to ease debugging, this should be exposed via API (without the need for having a concrete RemoteRepository).

kwin:
Instead of purely relying on exception handling when a transporter cannot deal with a certain kind of repository an in advance check on the repository protocol is introduced

This closes #1770

### DIFF
--- a/maven-resolver-spi/src/main/java/org/eclipse/aether/spi/connector/transport/TransporterFactory.java
+++ b/maven-resolver-spi/src/main/java/org/eclipse/aether/spi/connector/transport/TransporterFactory.java
@@ -18,6 +18,8 @@
  */
 package org.eclipse.aether.spi.connector.transport;
 
+import java.util.Map;
+
 import org.eclipse.aether.RepositorySystemSession;
 import org.eclipse.aether.repository.RemoteRepository;
 import org.eclipse.aether.transfer.NoTransporterException;
@@ -32,8 +34,38 @@ import org.eclipse.aether.transfer.NoTransporterException;
 public interface TransporterFactory {
 
     /**
+     * A key for transporter properties.
+     * @see #getProperties()
+     */
+    public interface TransporterPropertyKey {}
+
+    /**
+     * Indicates whether this factory can handle the specified repository protocol.
+     * Even if {@code true} is returned the factory may still refuse to create a transporter for the given protocol.
+     *
+     * @param repositoryProtocol The repository protocol to check, may be {@code null}.
+     * @return {@code true} if this factory can potentially handle the specified repository protocol, {@code false} otherwise.
+     * @see #newInstance(RepositorySystemSession, RemoteRepository)
+     */
+    default boolean canHandle(String repositoryProtocol) {
+        return true;
+    }
+
+    /**
+     * Gets the properties exposing information about this transporter factory.
+     * Some are mandatory and some are transporter-specific.
+     *
+     * @return A map of transporter property keys with values, never {@code null}.
+     * @throws UnsupportedOperationException if the transporter factory does not implement this method.
+     */
+    default Map<TransporterPropertyKey, Object> getProperties() {
+        throw new UnsupportedOperationException("getProperties not implemented");
+    }
+
+    /**
      * Tries to create a transporter for the specified remote repository. Typically, a factory will inspect
      * {@link RemoteRepository#getProtocol()} to determine whether it can handle a repository.
+     * This should be preceded by a call to {@link #canHandle(String)}.
      *
      * @param session The repository system session from which to configure the transporter, must not be {@code null}.
      *            In particular, a transporter should obey the timeouts configured for the session.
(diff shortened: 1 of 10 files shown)


maven-resolver-spi/src/main/java/org/eclipse/aether/spi/connector/transport/TransporterFactory.java
```diff
@@ -31,9 +33,39 @@
...
+
+    /**
+     * Indicates whether this factory can handle the specified repository protocol.
+     * Even if {@code true} is returned the factory may still refuse to create a transporter for the given protocol.
+     *
+     * @param repositoryProtocol The repository protocol to check, may be {@code null}.
+     * @return {@code true} if this factory can potentially handle the specified repository protocol, {@code false} otherwise.
+     * @see #newInstance(RepositorySystemSession, RemoteRepository)
+     */
+    default boolean canHandle(String repositoryProtocol) {
```
kwin, line 50: Maybe to ease listing all transporters it would be better to expose a list of supported protocols (potentially including a wildcard). WDYT @cstamas?

cstamas: Isn't requesting an instance for RemoteRepository (and getting NoTransportEx) same?

cstamas: Especially as sometimes you cannot tell ahead of time. for HTTP yes, but for file?

kwin: It is different in the following aspect: This just requires a protocol not a full blown remote repository. It doesn't rely on exceptions but just returns a boolean.

cstamas:
We have cases like `dav+http` or `symlink+file` etc. And as said above, we have cases where it is just empty string, but RemoteRepository would "still work", etc

kwin:
I looked at all transporters we have, and while it is true that some evaluate more than just protocol, they all evaluate the protocol first. Therefore having this aspect separated is IMHO useful for this feature (but also for determining the correct transporter) as in the best case you don't have the overhead of throwing an exception as you bail out for a non-matching protocol first.

cstamas:
Also, some transport may support or not support something based on config... and again, protocol is just not enough to decide all this. Either use repo url or whole repo instance. Hence, my 5 cents would be:
* make the method `canHandle(RepositorySystemSession, RemoteRepository)` instead. But, the erasure of this method now became same as that of `newInstance`....

gnodet (commented):
## AI Review — PR #1771: Expose transporter details per protocol

Hi @kwin, thanks for working on improving protocol handling in the transporter layer!

After review and independent verification, several significant issues were confirmed:

### 🔴 `FileTransporterFactory.canHandle` breaks compound protocols (high)
`canHandle` returns `true` only for `"bundle"` and `"file"`, but `newInstance` also handles `symlink+file`, `hardlink+file`, and custom FS providers (JIMFS). The code comments in `FileTransporterFactory` explicitly state: *"to support custom FS providers (like JIMFS), we cannot cover all possible protocols."* With the PR's `DefaultTransporterProvider` changes, `canHandle("symlink+file")` returns `false`, causing the factory to be skipped entirely — **breaking existing functionality** for symlink, hardlink, and JIMFS-based file transport.

### 🔴 Protocol validation removed from `newInstance` (high)
On master, all transport factory `newInstance()` methods throw `NoTransporterException` for unsupported protocols. The PR removes these checks with a comment "repository protocol already checked in canHandle." This breaks the SPI contract for **direct callers** who invoke `newInstance()` without going through `DefaultTransporterProvider`. The Javadoc still declares `@throws NoTransporterException` but the implementations no longer honor it.

### 🔴 `getProperties()` default throws `UnsupportedOperationException` (high)
Throwing from a default method in an SPI interface defeats the purpose of default methods (backward-compatible no-op). Any caller that invokes it gets a runtime exception. A default returning `Collections.emptyMap()` would be the standard Java idiom.

### ⚠️ Error accumulation degraded (medium)
On master, every factory that rejects a protocol contributes to the final diagnostic error message. With `canHandle`, factories returning `false` are silently skipped with only a DEBUG log — making the "Cannot access ... using the registered transporter factories" error incomplete and harder to diagnose.

### ⚠️ Test coverage regression (medium)
`testInit_BadProtocol` was changed to only test `canHandle` instead of `newInstance`, removing coverage for `newInstance` behavior with unsupported protocols.

Note: The concern about `canHandle(String)` being too narrow for compound protocols was evaluated as a **design discussion** (not a bug), since the default `return true` implementation allows factories to fall through to `newInstance()` as before. This aligns with @cstamas's feedback in the PR discussion.

CI is also failing across all platforms.

---
*🤖 This review was generated by [ForgeBot](https://github.com/gnodet/forgebot-skills) using a maker/checker pattern (reviewer + independent verifier). 1 of 7 original findings was removed as a false positive after verification.*

PR #1815, opened 2026-03-27
Advised and Enforced Management

cstamas:
Originally, dependency management marked a "subject" as "managed", it was 0/1 true/false only. This PR now adds ability to mark management as "advised" (may be subjected to further management) or "enforced" (equivalent of previously "managed").

Changes:
* `DependencyManager` applies management as "enforced", if they come from root (ie. project POM), otherwise as "advice" (this latter applies to transitive managers only, as for example classic manager have no data from "deeper levels").
* `ConflictManager` backs out for _enforced_ scope/optionality.
* This makes the "isInheritedDerived" hack in dependency manager not needed, as now management rule tells the modality of it (is enforced or just advised)

### DIFF
--- a/maven-resolver-api/src/main/java/org/eclipse/aether/collection/DependencyManagement.java
+++ b/maven-resolver-api/src/main/java/org/eclipse/aether/collection/DependencyManagement.java
@@ -19,7 +19,7 @@
 package org.eclipse.aether.collection;
 
 import java.util.Collection;
-import java.util.HashMap;
+import java.util.EnumMap;
 import java.util.Map;
 
 import org.eclipse.aether.graph.Exclusion;
@@ -43,15 +43,15 @@ public final class DependencyManagement {
         PROPERTIES
     }
 
-    private final Map<Subject, Object> managedValues;
-    private final Map<Subject, Boolean> managedEnforced;
+    private final EnumMap<Subject, Object> managedValues;
+    private final EnumMap<Subject, Boolean> managedEnforced;
 
     /**
      * Creates an empty management update.
      */
     public DependencyManagement() {
-        this.managedValues = new HashMap<>();
-        this.managedEnforced = new HashMap<>();
+        this.managedValues = new EnumMap<>(Subject.class);
+        this.managedEnforced = new EnumMap<>(Subject.class);
     }
 
     /**
--- a/maven-resolver-api/src/main/java/org/eclipse/aether/collection/DependencyManagement.java
+++ b/maven-resolver-api/src/main/java/org/eclipse/aether/collection/DependencyManagement.java
@@ -19,6 +19,7 @@
 package org.eclipse.aether.collection;
 
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.Map;
 
 import org.eclipse.aether.graph.Exclusion;
@@ -29,22 +30,46 @@ import org.eclipse.aether.graph.Exclusion;
  * @see DependencyManager#manageDependency(org.eclipse.aether.graph.Dependency)
  */
 public final class DependencyManagement {
+    /**
+     * Enumeration of manageable attributes, attributes that can be subjected to dependency management.
+     *
+     * @since 2.0.17
+     */
+    public enum Subject {
+        VERSION,
+        SCOPE,
+        OPTIONAL,
+        EXCLUSIONS,
+        PROPERTIES
+    }
 
-    private String version;
-
-    private String scope;
-
-    private Boolean optional;
-
-    private Collection<Exclusion> exclusions;
-
-    private Map<String, String> properties;
+    private final Map<Subject, Object> managedValues;
+    private final Map<Subject, Boolean> managedEnforced;
 
     /**
      * Creates an empty management update.
      */
     public DependencyManagement() {
-        // enables default constructor
+        this.managedValues = new HashMap<>();
+        this.managedEnforced = new HashMap<>();
+    }
+
+    /**
+     * Returns {@code true} if passed in subject is managed.
+     *
+     * @since 2.0.17
+     */
+    public boolean isManagedSubject(Subject subject) {
+        return managedValues.containsKey(subject);
+    }
+
+    /**
+     * Returns {@code true} if passed in subject is managed and is enforced.
+     *
+     * @since 2.0.17
+     */
+    public boolean isManagedSubjectEnforced(Subject subject) {
+        return isManagedSubject(subject) && managedEnforced.getOrDefault(subject, false);
     }
 
     /**
@@ -54,7 +79,7 @@ public final class DependencyManagement {
      *         remain unchanged.
      */
     public String getVersion() {
-        return version;
+        return (String) managedValues.get(Subject.VERSION);
     }
 
     /**
@@ -62,9 +87,29 @@ public final class DependencyManagement {
      *
      * @param version The new version, may be {@code null} if the version is not managed.
      * @return This management update for chaining, never {@code null}.
+     * @deprecated Use {@link #setVersion(String, boolean)} instead.
      */
+    @Deprecated
     public DependencyManagement setVersion(String version) {
-        this.version = version;
+        return setVersion(version, true);
+    }
+
+    /**
+     * Sets the new version to apply to the dependency.
+     *
+     * @param version The new version, may be {@code null} if the version is not managed.
+     * @param enforced The enforcement of new value.
+     * @return This management update for chaining, never {@code null}.
+     * @since 2.0.17
+     */
+    public DependencyManagement setVersion(String version, boolean enforced) {
+        if (version == null) {
+            this.managedValues.remove(Subject.VERSION);
+            this.managedEnforced.remove(Subject.VERSION);
+        } else {
+            this.managedValues.put(Subject.VERSION, version);
+            this.managedEnforced.put(Subject.VERSION, enforced);
+        }
         return this;
     }
 
@@ -75,7 +120,7 @@ public final class DependencyManagement {
      *         unchanged.
      */
     public String getScope() {
-        return scope;
+        return (String) managedValues.get(Subject.SCOPE);
     }
 
     /**
@@ -83,9 +128,29 @@ public final class DependencyManagement {
      *
      * @param scope The new scope, may be {@code null} if the scope is not managed.
      * @return This management update for chaining, never {@code null}.
+     * @deprecated Use {@link #setScope(String, boolean)} instead.
      */
+    @Deprecated
     public DependencyManagement setScope(String scope) {
-        this.scope = scope;
+        return setScope(scope, true);
+    }
+
+    /**
+     * Sets the new scope to apply to the dependency.
+     *
+     * @param scope The new scope, may be {@code null} if the scope is not managed.
+     * @param enforced The enforcement of new value.
+     * @return This management update for chaining, never {@code null}.
+     * @since 2.0.17
+     */
+    public DependencyManagement setScope(String scope, boolean enforced) {
+        if (scope == null) {
+            this.managedValues.remove(Subject.SCOPE);
+            this.managedEnforced.remove(Subject.SCOPE);
+        } else {
+            this.managedValues.put(Subject.SCOPE, scope);
+            this.managedEnforced.put(Subject.SCOPE, enforced);
+        }
         return this;
     }
 
@@ -96,7 +161,7 @@ public final class DependencyManagement {
      *         dependency should remain unchanged.
      */
     public Boolean getOptional() {
-        return optional;
+        return (Boolean) managedValues.get(Subject.OPTIONAL);
     }
 
     /**
@@ -104,9 +169,29 @@ public final class DependencyManagement {
      *
      * @param optional The optional flag, may be {@code null} if the flag is not managed.
      * @return This management update for chaining, never {@code null}.
+     * @deprecated Use {@link #setOptional(Boolean, boolean)} instead.
      */
+    @Deprecated
     public DependencyManagement setOptional(Boolean optional) {
-        this.optional = optional;
+        return setOptional(optional, true);
+    }
+
+    /**
+     * Sets the new optional flag to apply to the dependency.
+     *
+     * @param optional The optional flag, may be {@code null} if the flag is not managed.
+     * @param enforced The enforcement of new value.
+     * @return This management update for chaining, never {@code null}.
+     * @since 2.0.17
+     */
+    public DependencyManagement setOptional(Boolean optional, boolean enforced) {
+        if (optional == null) {
+            this.managedValues.remove(Subject.OPTIONAL);
+            this.managedEnforced.remove(Subject.OPTIONAL);
+        } else {
+            this.managedValues.put(Subject.OPTIONAL, optional);
+            this.managedEnforced.put(Subject.OPTIONAL, enforced);
+        }
         return this;
     }
 
@@ -118,8 +203,9 @@ public final class DependencyManagement {
      * @return The new exclusions or {@code null} if the exclusions are not managed and the existing dependency
      *         exclusions should remain unchanged.
      */
+    @SuppressWarnings("unchecked")
     public Collection<Exclusion> getExclusions() {
-        return exclusions;
+        return (Collection<Exclusion>) managedValues.get(Subject.EXCLUSIONS);
     }
 
     /**
@@ -129,9 +215,31 @@ public final class DependencyManagement {
      *
      * @param exclusions The new exclusions, may be {@code null} if the exclusions are not managed.
      * @return This management update for chaining, never {@code null}.
+     * @deprecated Use {@link #setExclusions(Collection, boolean)} instead.
      */
+    @Deprecated
     public DependencyManagement setExclusions(Collection<Exclusion> exclusions) {
-        this.exclusions = exclusions;
+        return setExclusions(exclusions, true);
+    }
+
+    /**
+     * Sets the new exclusions to apply to the dependency. Note that this collection denotes the complete set of
+     * exclusions for the dependency, i.e. the dependency manager controls whether any existing exclusions get merged
+     * with information from dependency management or overridden by it.
+     *
+     * @param exclusions The new exclusions, may be {@code null} if the exclusions are not managed.
+     * @param enforced The enforcement of new value.
+     * @return This management update for chaining, never {@code null}.
+     * @since 2.0.17
+     */
+    public DependencyManagement setExclusions(Collection<Exclusion> exclusions, boolean enforced) {
+        if (exclusions == null) {
+            this.managedValues.remove(Subject.EXCLUSIONS);
+            this.managedEnforced.remove(Subject.EXCLUSIONS);
+        } else {
+            this.managedValues.put(Subject.EXCLUSIONS, exclusions);
+            this.managedEnforced.put(Subject.EXCLUSIONS, enforced);
+        }
         return this;
     }
 
@@ -143,8 +251,9 @@ public final class DependencyManagement {
      * @return The new artifact properties or {@code null} if the properties are not managed and the existing properties
      *         should remain unchanged.
      */
+    @SuppressWarnings("unchecked")
     public Map<String, String> getProperties() {
-        return properties;
+        return (Map<String, String>) managedValues.get(Subject.PROPERTIES);
     }
 
     /**
@@ -154,9 +263,31 @@ public final class DependencyManagement {
      *
      * @param properties The new artifact properties, may be {@code null} if the properties are not managed.
      * @return This management update for chaining, never {@code null}.
+     * @deprecated Use {@link #setProperties(Map, boolean)} instead.
      */
+    @Deprecated
     public DependencyManagement setProperties(Map<String, String> properties) {
-        this.properties = properties;
+        return setProperties(properties, true);
+    }
+
+    /**
+     * Sets the new properties to apply to the dependency. Note that this map denotes the complete set of properties,
+     * i.e. the dependency manager controls whether any existing properties get merged with the information from
+     * dependency management or overridden by it.
+     *
+     * @param properties The new artifact properties, may be {@code null} if the properties are not managed.
+     * @param enforced The enforcement of new value.
+     * @return This management update for chaining, never {@code null}.
+     * @since 2.0.17
+     */
+    public DependencyManagement setProperties(Map<String, String> properties, boolean enforced) {
+        if (properties == null) {
+            this.managedValues.remove(Subject.PROPERTIES);
+            this.managedEnforced.remove(Subject.PROPERTIES);
+        } else {
+            this.managedValues.put(Subject.PROPERTIES, properties);
+            this.managedEnforced.put(Subject.PROPERTIES, enforced);
+        }
         return this;
     }
 }
--- a/maven-resolver-api/src/main/java/org/eclipse/aether/collection/DependencyManagement.java
+++ b/maven-resolver-api/src/main/java/org/eclipse/aether/collection/DependencyManagement.java
@@ -20,6 +20,7 @@ package org.eclipse.aether.collection;
 
 import java.util.Collection;
 import java.util.EnumMap;
+import java.util.EnumSet;
 import java.util.Map;
 
 import org.eclipse.aether.graph.Exclusion;
@@ -33,7 +34,7 @@ public final class DependencyManagement {
     /**
      * Enumeration of manageable attributes, attributes that can be subjected to dependency management.
      *
-     * @since 2.0.17
+     * @since 2.0.19
      */
     public enum Subject {
         VERSION,
@@ -44,20 +45,20 @@ public final class DependencyManagement {
     }
 
     private final EnumMap<Subject, Object> managedValues;
-    private final EnumMap<Subject, Boolean> managedEnforced;
+    private final EnumSet<Subject> managedEnforced;
 
     /**
      * Creates an empty management update.
      */
     public DependencyManagement() {
         this.managedValues = new EnumMap<>(Subject.class);
-        this.managedEnforced = new EnumMap<>(Subject.class);
+        this.managedEnforced = EnumSet.noneOf(Subject.class);
     }
 
     /**
      * Returns {@code true} if passed in subject is managed.
      *
-     * @since 2.0.17
+     * @since 2.0.19
      */
     public boolean isManagedSubject(Subject subject) {
         return managedValues.containsKey(subject);
@@ -66,10 +67,10 @@ public final class DependencyManagement {
     /**
      * Returns {@code true} if passed in subject is managed and is enforced.
      *
-     * @since 2.0.17
+     * @since 2.0.19
      */
     public boolean isManagedSubjectEnforced(Subject subject) {
-        return isManagedSubject(subject) && managedEnforced.getOrDefault(subject, false);
+        return isManagedSubject(subject) && managedEnforced.contains(subject);
     }
 
     /**
@@ -100,17 +101,10 @@ public final class DependencyManagement {
      * @param version The new version, may be {@code null} if the version is not managed.
      * @param enforced The enforcement of new value.
      * @return This management update for chaining, never {@code null}.
-     * @since 2.0.17
+     * @since 2.0.19
      */
     public DependencyManagement setVersion(String version, boolean enforced) {
-        if (version == null) {
-            this.managedValues.remove(Subject.VERSION);
-            this.managedEnforced.remove(Subject.VERSION);
-        } else {
-            this.managedValues.put(Subject.VERSION, version);
-            this.managedEnforced.put(Subject.VERSION, enforced);
-        }
-        return this;
+        return set(Subject.VERSION, version, enforced);
     }
 
     /**
@@ -141,17 +135,10 @@ public final class DependencyManagement {
      * @param scope The new scope, may be {@code null} if the scope is not managed.
      * @param enforced The enforcement of new value.
      * @return This management update for chaining, never {@code null}.
-     * @since 2.0.17
+     * @since 2.0.19
      */
     public DependencyManagement setScope(String scope, boolean enforced) {
-        if (scope == null) {
-            this.managedValues.remove(Subject.SCOPE);
-            this.managedEnforced.remove(Subject.SCOPE);
-        } else {
-            this.managedValues.put(Subject.SCOPE, scope);
-            this.managedEnforced.put(Subject.SCOPE, enforced);
-        }
-        return this;
+        return set(Subject.SCOPE, scope, enforced);
     }
 
     /**
@@ -182,17 +169,10 @@ public final class DependencyManagement {
      * @param optional The optional flag, may be {@code null} if the flag is not managed.
      * @param enforced The enforcement of new value.
      * @return This management update for chaining, never {@code null}.
-     * @since 2.0.17
+     * @since 2.0.19
      */
     public DependencyManagement setOptional(Boolean optional, boolean enforced) {
-        if (optional == null) {
-            this.managedValues.remove(Subject.OPTIONAL);
-            this.managedEnforced.remove(Subject.OPTIONAL);
-        } else {
-            this.managedValues.put(Subject.OPTIONAL, optional);
-            this.managedEnforced.put(Subject.OPTIONAL, enforced);
-        }
-        return this;
+        return set(Subject.OPTIONAL, optional, enforced);
     }
 
     /**
@@ -230,17 +210,10 @@ public final class DependencyManagement {
      * @param exclusions The new exclusions, may be {@code null} if the exclusions are not managed.
      * @param enforced The enforcement of new value.
      * @return This management update for chaining, never {@code null}.
-     * @since 2.0.17
+     * @since 2.0.19
      */
     public DependencyManagement setExclusions(Collection<Exclusion> exclusions, boolean enforced) {
-        if (exclusions == null) {
-            this.managedValues.remove(Subject.EXCLUSIONS);
-            this.managedEnforced.remove(Subject.EXCLUSIONS);
-        } else {
-            this.managedValues.put(Subject.EXCLUSIONS, exclusions);
-            this.managedEnforced.put(Subject.EXCLUSIONS, enforced);
-        }
-        return this;
+        return set(Subject.EXCLUSIONS, exclusions, enforced);
     }
 
     /**
@@ -278,15 +251,26 @@ public final class DependencyManagement {
      * @param properties The new artifact properties, may be {@code null} if the properties are not managed.
      * @param enforced The enforcement of new value.
      * @return This management update for chaining, never {@code null}.
-     * @since 2.0.17
+     * @since 2.0.19
      */
     public DependencyManagement setProperties(Map<String, String> properties, boolean enforced) {
-        if (properties == null) {
-            this.managedValues.remove(Subject.PROPERTIES);
-            this.managedEnforced.remove(Subject.PROPERTIES);
+        return set(Subject.PROPERTIES, properties, enforced);
+    }
+
+    /**
+     * Generic but private setter that applies common logic.
+     */
+    private DependencyManagement set(Subject subject, Object value, boolean enforced) {
+        if (value == null) {
+            this.managedValues.remove(subject);
+            this.managedEnforced.remove(subject);
         } else {
-            this.managedValues.put(Subject.PROPERTIES, properties);
-            this.managedEnforced.put(Subject.PROPERTIES, enforced);
+            this.managedValues.put(subject, value);
+            if (enforced) {
+                this.managedEnforced.add(subject);
+            } else {
+                this.managedEnforced.remove(subject);
+            }
         }
         return this;
     }
--- a/maven-resolver-api/src/main/java/org/eclipse/aether/graph/DefaultDependencyNode.java
+++ b/maven-resolver-api/src/main/java/org/eclipse/aether/graph/DefaultDependencyNode.java
@@ -26,6 +26,7 @@ import java.util.List;
 import java.util.Map;
 
 import org.eclipse.aether.artifact.Artifact;
+import org.eclipse.aether.collection.DependencyManagement;
 import org.eclipse.aether.repository.RemoteRepository;
 import org.eclipse.aether.version.Version;
 import org.eclipse.aether.version.VersionConstraint;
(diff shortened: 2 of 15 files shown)


gnodet (commented):

## In-depth review of "Advised and Enforced Management"

Thanks for working on this @cstamas. The concept of distinguishing "advised" vs "enforced" management is a valuable improvement — it addresses a real pain point where transitive dependency management would mark nodes as "managed" and prevent proper scope/optional inheritance during graph transformation. Here's a detailed review.

---

### Summary

This PR replaces the binary "managed or not" bit field (`managedBits`) with a richer model that tracks both *whether* and *how* (advised vs enforced) each attribute is managed. The key behavioral change is:
- Root-level (project POM) management → **enforced** (conflict resolver won't override)
- Transitive management → **advised** (conflict resolver may still derive scope/optional)
- This eliminates the `isInheritedDerived()` hack in `TransitiveDependencyManager`

---

### Concerns

#### 1. Breaking change: `setManagedBits()` throws at runtime

`DefaultDependencyNode.setManagedBits(int)` now throws `IllegalArgumentException("bits are not supported")`. While `@noextend`/`@noimplement` hints exist on the interface, `DefaultDependencyNode` is a public API class and consumers (Maven plugins, IDE integrations like IntelliJ/Eclipse, test frameworks) may call `setManagedBits()` on existing node instances. A runtime exception is a hard break. Consider instead:
- Mapping the bits to the new `managedSubjects` map (all as enforced for backwards compat)
- Or at minimum, deprecating with a no-op + warning log rather than throwing

#### 2. Type safety regression in `DependencyManagement`

The refactoring from typed fields (`String version`, `String scope`, `Boolean optional`, etc.) to `Map<Subject, Object>` loses compile-time type safety. The getters now rely on unchecked casts:
```java
return (String) managedValues.get(Subject.VERSION);
return (Collection<Exclusion>) managedValues.get(Subject.EXCLUSIONS);
```
This means a bug like `managedValues.put(Subject.VERSION, 42)` would compile but fail at runtime. Consider keeping the typed fields alongside the enforcement map, or using a typed wrapper/record per Subject.

#### 3. No new tests for the core semantic change

The PR introduces the advised/enforced distinction but adds **no new test cases** exercising this behavior. The existing `testDependencyManagement_VerboseMode` still tests via the deprecated `getManagedBits()` API. Tests should verify:
- That transitive management produces "advised" (`enforced=false`) management
- That root-level management produces "enforced" (`enforced=true`) management  
- That `ClassicConflictResolver` and `PathConflictResolver` correctly skip scope/optional derivation for enforced nodes but apply it for advised nodes
- That the `TestDependencyManager` in tests exercises both paths

#### 4. New interface methods without `default` implementations

`DependencyNode.isManagedSubject()` and `isManagedSubjectEnforced()` are added without `default` methods. While `@noimplement` signals intent, adding `default` implementations (delegating to `getManagedBits()` for `isManagedSubject` and returning `false` for `isManagedSubjectEnforced`) would be more resilient and maintain backwards compat for any existing (even if unsupported) implementations.

#### 5. Inconsistent semantics of `isManagedSubjectEnforced` between classes

In `DependencyManagement`:
```java
public boolean isManagedSubjectEnforced(Subject subject) {
    return isManagedSubject(subject) && managedEnforced.getOrDefault(subject, false);
}
```
In `DefaultDependencyNode`:
```java
public boolean isManagedSubjectEnforced(DependencyManagement.Subject subject) {
    return managedSubjects.getOrDefault(subject, false);
}
```
The `DefaultDependencyNode` version doesn't first check `isManagedSubject()`. For an unmanaged subject, both return `false`, but the *semantics* diverge: one says "not managed so not enforced", the other says "default to false". This could lead to subtle bugs if the map ever contained `null` values. Suggest aligning the implementations.

#### 6. Method rename is a binary-breaking change

`createDependencyNode(...)` (the cycle variant) → `createDependencyNodeCycle(...)` in `DependencyCollectorDelegate`. This is a `protected` method, and while `DependencyCollectorDelegate` is internal, any custom collector subclass will break at compile time. The non-static promotion of `createDependencyNode` (removing `static`) is also a binary-level change.

#### 7. Memory overhead

`DependencyManagement()` now always allocates two `HashMap` instances even when empty. In large dependency graphs, `manageDependency()` is called for every dependency at every level. Most calls will produce a `DependencyManagement` with 0-2 managed subjects. Consider lazy initialization or using `EnumMap` (which is optimized for enum keys and uses less memory than `HashMap`).

#### 8. Missing `@param` in Javadoc

The new interface methods `isManagedSubject` and `isManagedSubjectEnforced` in `DependencyNode.java` are missing `@param subject` documentation.

---

### Minor observations

- The `@Override` annotations added to `DefaultDependencyNode` methods that already existed (like `getChildren()`, `getDependency()`, etc.) are a nice cleanup but are unrelated to this PR's purpose — consider splitting them out.
- The copy constructor in `DefaultDependencyNode` now iterates all `Subject.values()` to build the managed subjects map. This is fine for 5 enum values but is more allocation-heavy than the previous `managedBits = (byte)(...)` one-liner. Not a real concern, just noting the trade-off.
- The `managedSubjects` field in `DefaultDependencyNode` is initialized to `Collections.emptyMap()` but `setManagedSubjects(null)` also sets it to `Collections.emptyMap()`. This is consistent but the field could just be declared without the initializer since the constructor paths all set it.

---

### Questions for discussion

1. **Exclusions are always "not enforced"** (`management.setExclusions(result, false)` in `AbstractDependencyManager`). Is this intentional? If root-level exclusion management should also be enforced, this might need the same `path.isEmpty()` check.

2. **Properties from system path handling are always "not enforced"** (`management.setProperties(properties, false)`). Same question — is this by design?

3. The determination of "enforced" via `versionOwner.path.isEmpty()` is clever but implicit. Would it be clearer to have an explicit `isRoot()` method or pass the enforcement flag more directly?

maven-resolver-api/src/main/java/org/eclipse/aether/graph/DefaultDependencyNode.java
```diff
@@ -201,34 +223,67 @@ public void setVersion(Version version) {
...
 
+    @Override
     public void setOptional(Boolean optional) {
         if (dependency == null) {
             throw new IllegalStateException("node does not have a dependency");
         }
         dependency = dependency.setOptional(optional);
     }
 
+    @Override
```
gnodet, line 245: _Claude Code on behalf of Guillaume Nodet_

This is a hard runtime break for any consumer code that calls `setManagedBits()` on `DefaultDependencyNode`. Even though the interface says `@noextend`, consumers do instantiate and configure `DefaultDependencyNode` directly.

Suggestion: map the bits to the new model for backwards compat:
```java
@Deprecated
public void setManagedBits(int managedBits) {
    HashMap<DependencyManagement.Subject, Boolean> subjects = new HashMap<>();
    if ((managedBits & DependencyNode.MANAGED_VERSION) != 0) {
        subjects.put(DependencyManagement.Subject.VERSION, true);
    }
    if ((managedBits & DependencyNode.MANAGED_SCOPE) != 0) {
        subjects.put(DependencyManagement.Subject.SCOPE, true);
    }
    if ((managedBits & DependencyNode.MANAGED_OPTIONAL) != 0) {
        subjects.put(DependencyManagement.Subject.OPTIONAL, true);
    }
    if ((managedBits & DependencyNode.MANAGED_PROPERTIES) != 0) {
        subjects.put(DependencyManagement.Subject.PROPERTIES, true);
    }
    if ((managedBits & DependencyNode.MANAGED_EXCLUSIONS) != 0) {
        subjects.put(DependencyManagement.Subject.EXCLUSIONS, true);
    }
    setManagedSubjects(subjects.isEmpty() ? null : subjects);
}
```

maven-resolver-api/src/main/java/org/eclipse/aether/collection/DependencyManagement.java
```diff
@@ -29,22 +30,46 @@
...
+     * Enumeration of manageable attributes, attributes that can be subjected to dependency management.
+     *
+     * @since 2.0.17
+     */
+    public enum Subject {
+        VERSION,
+        SCOPE,
+        OPTIONAL,
+        EXCLUSIONS,
+        PROPERTIES
```
gnodet, line 44: _Claude Code on behalf of Guillaume Nodet_

Replacing typed fields with `Map<Subject, Object>` loses compile-time type safety. A mistyped `managedValues.put(Subject.VERSION, 42)` would compile but fail at runtime with a `ClassCastException` here.

Consider using `EnumMap` (better memory for enum keys) and keeping typed fields alongside the enforcement tracking, or introducing a small typed record:

```java
private record ManagedValue<T>(T value, boolean enforced) {}
```

maven-resolver-api/src/main/java/org/eclipse/aether/collection/DependencyManagement.java
```diff
@@ -29,22 +30,46 @@
...
-
-    private String scope;
-
-    private Boolean optional;
-
-    private Collection<Exclusion> exclusions;
-
-    private Map<String, String> properties;
+    private final Map<Subject, Object> managedValues;
+    private final Map<Subject, Boolean> managedEnforced;
```
gnodet: _Claude Code on behalf of Guillaume Nodet_

Two `HashMap` instances are allocated per `DependencyManagement`, even when empty. In large dependency graphs this adds up since `manageDependency()` is called for every dependency at every level.

Consider:
1. Lazy initialization (allocate only when a setter is called with non-null)
2. Use `EnumMap<Subject, ...>` instead — it uses a fixed-size array internally, much more compact for small enums

maven-resolver-api/src/main/java/org/eclipse/aether/graph/DependencyNode.java
```diff
@@ -38,40 +39,49 @@
...
     /**
      * A bit flag indicating the dependency scope was subject to dependency management
      *
      * @see #getManagedBits()
+     * @deprecated Use {@link #isManagedSubject(DependencyManagement.Subject)} and {@link #isManagedSubjectEnforced(DependencyManagement.Subject)} instead.
      */
+    @Deprecated
     int MANAGED_SCOPE = 0x02;
 
     /**
```
gnodet, line 60: _Claude Code on behalf of Guillaume Nodet_

These new methods have no `default` implementation. While `@noimplement` signals intent, providing defaults would be more resilient:

```java
default boolean isManagedSubject(DependencyManagement.Subject subject) {
    // Backwards-compatible default based on deprecated getManagedBits()
    return switch (subject) {
        case VERSION -> (getManagedBits() & MANAGED_VERSION) != 0;
        case SCOPE -> (getManagedBits() & MANAGED_SCOPE) != 0;
        case OPTIONAL -> (getManagedBits() & MANAGED_OPTIONAL) != 0;
        case PROPERTIES -> (getManagedBits() & MANAGED_PROPERTIES) != 0;
        case EXCLUSIONS -> (getManagedBits() & MANAGED_EXCLUSIONS) != 0;
    };
}

default boolean isManagedSubjectEnforced(DependencyManagement.Subject subject) {
    return isManagedSubject(subject); // conservative default: all managed = enforced
}
```

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/manager/AbstractDependencyManager.java
```diff
@@ -394,58 +393,59 @@ public DependencyManagement manageDependency(Dependency dependency) {
...
         if (isApplied()) {
-            String version = getManagedVersion(key);
+            AbstractDependencyManager versionOwner = getManagedVersion(key);
             // is managed locally by model builder
             // apply only rules coming from "higher" levels
-            if (version != null) {
+            if (versionOwner != null) {
                 management = new DependencyManagement();
-                management.setVersion(version);
+                management.setVersion(versionOwner.managedVersions.get(key), versionOwner.path.isEmpty());
```
gnodet: _Claude Code on behalf of Guillaume Nodet_

The enforcement determination via `versionOwner.path.isEmpty()` is implicit — `path` is empty when the manager represents the root level. This works but is non-obvious. Consider adding an `isRoot()` helper or at least a comment explaining the semantics.

Also, this means the enforcement is determined by *where the management entry was defined* (root vs transitive), not by an explicit flag on the entry. Is that always correct? A parent POM's `<dependencyManagement>` inherited into a child POM would still be at depth 0 and thus enforced — which seems correct. But worth documenting explicitly.

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/manager/AbstractDependencyManager.java
```diff
@@ -461,7 +461,7 @@ public DependencyManagement manageDependency(Dependency dependency) {
             }
             Collection<Exclusion> result = new LinkedHashSet<>(dependency.getExclusions());
```
gnodet, line 471: _Claude Code on behalf of Guillaume Nodet_

Exclusions are always set with `enforced=false`. Is this intentional? Root-level exclusion management (`<dependencyManagement>` in the project POM) arguably should be enforced the same way as scope/version/optional from root level. If a user explicitly declares exclusions in their project's dependency management, the conflict resolver should arguably respect that.

maven-resolver-api/src/main/java/org/eclipse/aether/graph/DefaultDependencyNode.java
```diff
@@ -201,34 +223,67 @@ public void setVersion(Version version) {
...
+            res |= DependencyNode.MANAGED_OPTIONAL;
+        }
+        if (isManagedSubject(DependencyManagement.Subject.PROPERTIES)) {
+            res |= DependencyNode.MANAGED_PROPERTIES;
+        }
+        if (isManagedSubject(DependencyManagement.Subject.EXCLUSIONS)) {
+            res |= DependencyNode.MANAGED_EXCLUSIONS;
+        }
+        return res;
     }
```
gnodet, line 264: _Claude Code on behalf of Guillaume Nodet_

Subtle inconsistency with `DependencyManagement.isManagedSubjectEnforced()` which first checks `isManagedSubject(subject)`:
```java
// DependencyManagement:
return isManagedSubject(subject) && managedEnforced.getOrDefault(subject, false);
// DefaultDependencyNode:
return managedSubjects.getOrDefault(subject, false);
```
Both return `false` for unmanaged subjects today, but the semantic contract differs. Suggest aligning by adding the `containsKey` guard here too, or documenting the contract explicitly.

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/manager/TransitiveDependencyManager.java
```diff
@@ -145,30 +145,4 @@ protected DependencyManager newInstance(
...
-     * </p>
-     *
-     * @return true only at depth 0 (root level) to ensure inheritance-based properties are only
-     *         derived from the root, false otherwise
-     */
-    @Override
-    protected boolean isInheritedDerived() {
-        return depth == 0;
-    }
 }
```
gnodet, line 143: _Claude Code on behalf of Guillaume Nodet_

The removal of `isInheritedDerived()` is the most consequential behavioral change in this PR. Previously, `TransitiveDependencyManager` overrode this to return `depth == 0`, gating scope/optional management to root-only. Now, scope/optional are always collected into `managedScopes`/`managedOptionals` at all levels, but marked as "advised" (not enforced) when coming from non-root.

This changes the semantics: before, transitive scope/optional management entries were **not applied at all** (they were skipped in `deriveChildManager`). Now they **are applied** but marked as non-enforced, so the conflict resolver can still override them. 

This is a meaningful behavioral difference — a transitive dependency management entry for scope would now *initially set* the scope (which it previously wouldn't), and only later be potentially overridden by the conflict resolver. Is this the intended behavior? It seems correct in principle but should have test coverage.

maven-resolver-api/src/main/java/org/eclipse/aether/graph/DefaultDependencyNode.java
cstamas: fixed

maven-resolver-api/src/main/java/org/eclipse/aether/collection/DependencyManagement.java
cstamas: Maven Resolver 2.x is still Java 8 baseline, so no `record` yet. Using EnumMap

maven-resolver-api/src/main/java/org/eclipse/aether/collection/DependencyManagement.java
cstamas: Using EnumMap

maven-resolver-api/src/main/java/org/eclipse/aether/graph/DependencyNode.java
cstamas: Applied

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/manager/AbstractDependencyManager.java
cstamas: This change is here only to adopt new method and avoid usage of old (deprecated one). Also, if you check this PR, only the `scope` and `optional` are _checked_ (in conflict resolver) for "is it managed or not" and contain needed change for "is it enforced or not". Basically _right now_, whether exclusion is advised or enforced, does not matter, as it is taken as granted (again, only scope and optional are being altered by conflict resolver).

Later, after this PR is merged, the platform support will be the one that will use more this feature.

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/manager/TransitiveDependencyManager.java
cstamas: Yes, this is intentional. But, those will be now applied as "advices" and not as "enforced changes". In case of "nearest" winner selection this should not produce any change in case of conflict (same behavior remains), while it may produce proper "transitive dependency management" in other cases. Where it gets interesting is the use of "highest" winner selection strategy.

maven-resolver-api/src/main/java/org/eclipse/aether/graph/DefaultDependencyNode.java
gnodet: _Claude Code on behalf of Guillaume Nodet_

Thanks, the backwards-compatible mapping looks good. Much better than throwing.

maven-resolver-api/src/main/java/org/eclipse/aether/collection/DependencyManagement.java
gnodet: _Claude Code on behalf of Guillaume Nodet_

Good point about the Java 8 baseline. `EnumMap` is the right choice here. The remaining `Map<Subject, Object>` type erasure concern is minor given that all setters are type-safe and getters are the only cast points. Acceptable trade-off.

maven-resolver-api/src/main/java/org/eclipse/aether/collection/DependencyManagement.java
gnodet: _Claude Code on behalf of Guillaume Nodet_

`EnumMap` is perfect here — fixed-size internal array, much lighter than `HashMap`. Thanks.

maven-resolver-api/src/main/java/org/eclipse/aether/graph/DependencyNode.java
gnodet: _Claude Code on behalf of Guillaume Nodet_

The default implementations look correct. The `default` case in the `isManagedSubject` switch throwing `IllegalArgumentException` is good defensive coding for potential future enum additions. LGTM.

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/manager/AbstractDependencyManager.java
gnodet: _Claude Code on behalf of Guillaume Nodet_

Thanks for the explanation and for switching to `enforced=true`. That makes sense — exclusions are additive and should be respected as-is. Good that the conflict resolver does not alter exclusions currently, but marking them as enforced is the correct default for future-proofing.

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/manager/TransitiveDependencyManager.java
gnodet: _Claude Code on behalf of Guillaume Nodet_

Thanks for clarifying. So the key behavioral difference is:
- **Before:** Transitive scope/optional management entries were silently skipped (never applied)
- **After:** They are applied as "advice" — the initial scope/optional is set, but the conflict resolver can still override them during graph transformation

This is a meaningful improvement, especially for the "highest" winner selection strategy where transitive management entries should have influence. For the "nearest" strategy the end result should be the same since the conflict resolver will still derive scope/optional normally.

I think this behavioral change warrants at least one dedicated test to document and protect the expected behavior — see my overall comment about missing test coverage.

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/manager/AbstractDependencyManager.java
gnodet: _Claude Code on behalf of Guillaume Nodet_

The new `isRootManager()` helper method added in the latest commit is exactly what I had in mind. Much clearer than the raw `path.isEmpty()` check. Thanks.

maven-resolver-api/src/main/java/org/eclipse/aether/graph/DefaultDependencyNode.java
gnodet: _Claude Code on behalf of Guillaume Nodet_

Fixed in the "Align semantic" commit — `DefaultDependencyNode.isManagedSubjectEnforced()` now guards with `isManagedSubject(subject) &&` just like `DependencyManagement`. Thanks.

gnodet (commented):

## Follow-up review after updates

Thanks @cstamas for the quick and thorough iteration. The updates address most of the original concerns well:

**Resolved:**
- `setManagedBits()` now maps bits to the new model instead of throwing — backwards compat preserved
- `EnumMap` instead of `HashMap` — better memory footprint (good call on the Java 8 baseline, no records)
- `default` methods on `DependencyNode` interface — proper fallback for any existing implementations
- `isManagedSubjectEnforced` semantics aligned between `DependencyManagement` and `DefaultDependencyNode`
- `isRootManager()` helper method — much clearer than raw `path.isEmpty()`
- Exclusions changed to `enforced=true` — correct default
- `managedSubjects` uses `null` instead of empty map — good optimization

**Still outstanding — test coverage:**

The PR still has no new tests for the core advised/enforced distinction. This is the single remaining concern. Specifically, I'd like to see tests verifying:

1. **Root-level management produces enforced results:** When `DependencyManager` manages scope/optional from root, `isManagedSubjectEnforced()` returns `true` on the resulting `DependencyNode`.

2. **Transitive management produces advised results:** When `TransitiveDependencyManager` manages scope/optional from a non-root level, `isManagedSubjectEnforced()` returns `false`.

3. **Conflict resolver respects enforcement:** `ClassicConflictResolver` (and `PathConflictResolver`) should skip scope derivation for enforced nodes but apply it for advised nodes. This is the critical end-to-end behavior change.

4. **Backwards compat round-trip:** `setManagedBits(bits)` → `isManagedSubject()` / `isManagedSubjectEnforced()` returns correct values (all enforced by default).

Even a single integration test exercising scenario (3) would significantly reduce the risk of regressions. The behavioral change from "skip transitive scope/optional management entirely" to "apply as advice, let conflict resolver override" is subtle enough to warrant explicit test documentation.

Other than the missing tests, LGTM.

gnodet (commented):

## Follow-up review after merge with master

Merge is clean — all key changes are preserved (`isInheritedDerived()` removed from both classes, conflict resolvers use `isManagedSubjectEnforced`, scope/optional collected unconditionally). Tests pass. A few performance suggestions and a minor observation on the test parser.

### Performance

The switch from `int managedBits` to `EnumMap<Subject, Boolean>` introduces allocation overhead on the hot collection path. The suggestions below are all low-risk, mechanically simple optimizations — none affect correctness.

### Behavioral note

The removal of `isInheritedDerived()` changes `TransitiveDependencyManager` semantics: transitive scope/optional management was previously **discarded** (never collected), now it's **collected as advised** (applied during collection, overridable during conflict resolution). For nearest-wins this produces the same end result. For highest-wins it may surface new scope/optional values from transitive management that were previously invisible. This is the intended design — just noting that it's not a pure refactoring but a semantic improvement.

The conflict resolvers are correctly updated — `isManagedSubjectEnforced` is the right gate for scope/optional derivation, and no other components need changes.

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/PremanagedDependency.java
```diff
@@ -92,7 +95,8 @@ public static PremanagedDependency create(
             boolean premanagedState) {
         DependencyManagement depMngt = depManager != null ? depManager.manageDependency(dependency) : null;
 
-        int managedBits = 0;
+        EnumMap<DependencyManagement.Subject, Boolean> managedSubjects =
+                new EnumMap<>(DependencyManagement.Subject.class);
```
gnodet: _Claude Code on behalf of Guillaume Nodet_

**Performance:** This `EnumMap` is allocated for **every** dependency during collection, even when `depMngt` is null (no management applies). Previously this was `int managedBits = 0` — zero allocation. For a typical enterprise project, the majority of dependencies are unmanaged.

```suggestion
        EnumMap<DependencyManagement.Subject, Boolean> managedSubjects = null;
```

Then lazily create inside the `if (depMngt != null)` block, e.g.:
```java
if (managedSubjects == null) {
    managedSubjects = new EnumMap<>(DependencyManagement.Subject.class);
}
managedSubjects.put(...);
```

`setManagedSubjects(null)` already handles null correctly, so no other changes needed.

maven-resolver-api/src/main/java/org/eclipse/aether/graph/DefaultDependencyNode.java
```diff
@@ -99,12 +102,25 @@ public DefaultDependencyNode(Artifact artifact) {
...
+        this.dependency = node.getDependency();
+        this.artifact = node.getArtifact();
+        this.children = new ArrayList<>(0);
         setAliases(node.getAliases());
         setRequestContext(node.getRequestContext());
-        setManagedBits(node.getManagedBits());
+
+        EnumMap<DependencyManagement.Subject, Boolean> managedSubjects =
+                new EnumMap<>(DependencyManagement.Subject.class);
+        for (DependencyManagement.Subject subject : DependencyManagement.Subject.values()) {
```
gnodet: _Claude Code on behalf of Guillaume Nodet_

**Performance — two issues:**

1. `Subject.values()` allocates a new array each call (Java spec requires a fresh clone). Cache it:
```java
private static final DependencyManagement.Subject[] SUBJECTS = DependencyManagement.Subject.values();
```
then use `for (DependencyManagement.Subject subject : SUBJECTS)`.

2. The `EnumMap` is allocated even when the source node has no managed subjects. Guard the allocation:
```java
EnumMap<DependencyManagement.Subject, Boolean> managedSubjects = null;
for (DependencyManagement.Subject subject : SUBJECTS) {
    if (node.isManagedSubject(subject)) {
        if (managedSubjects == null) {
            managedSubjects = new EnumMap<>(DependencyManagement.Subject.class);
        }
        managedSubjects.put(subject, node.isManagedSubjectEnforced(subject));
    }
}
setManagedSubjects(managedSubjects);
```
This eliminates both the `isEmpty()` check and the wasted allocation for unmanaged nodes.

maven-resolver-api/src/main/java/org/eclipse/aether/graph/DefaultDependencyNode.java
```diff
@@ -201,34 +226,83 @@ public void setVersion(Version version) {
...
-    /**
-     * Sets a bit field indicating which attributes of this node were subject to dependency management.
-     *
-     * @param managedBits The bit field indicating the managed attributes or {@code 0} if dependency management wasn't
-     *            applied.
-     */
+    @Deprecated
     public void setManagedBits(int managedBits) {
-        this.managedBits = (byte) (managedBits & 0x1F);
+        EnumMap<DependencyManagement.Subject, Boolean> subjects = new EnumMap<>(DependencyManagement.Subject.class);
```
gnodet, line 272: _Claude Code on behalf of Guillaume Nodet_

**Performance:** Short-circuit for `managedBits == 0` to avoid the wasted `EnumMap` allocation:

```java
@Deprecated
public void setManagedBits(int managedBits) {
    if (managedBits == 0) {
        setManagedSubjects(null);
        return;
    }
    EnumMap<DependencyManagement.Subject, Boolean> subjects = new EnumMap<>(DependencyManagement.Subject.class);
    // ... rest unchanged
```

maven-resolver-api/src/main/java/org/eclipse/aether/collection/DependencyManagement.java
```diff
@@ -29,22 +30,46 @@
...
-
-    private String scope;
-
-    private Boolean optional;
-
-    private Collection<Exclusion> exclusions;
-
-    private Map<String, String> properties;
+    private final EnumMap<Subject, Object> managedValues;
+    private final EnumMap<Subject, Boolean> managedEnforced;
```
gnodet: _Claude Code on behalf of Guillaume Nodet_

**Performance:** `managedEnforced` only ever stores `true` (if a subject isn't enforced, it's either absent or `false`). An `EnumSet<Subject>` would be a better fit — it's backed by a single `long` for enums with ≤64 constants, so essentially zero overhead vs ~80-100 bytes for the `EnumMap<Subject, Boolean>`.

```java
private final EnumSet<Subject> managedEnforced;
```

Then `setVersion(String version, boolean enforced)` becomes:
```java
if (enforced) {
    this.managedEnforced.add(Subject.VERSION);
} else {
    this.managedEnforced.remove(Subject.VERSION);
}
```

And `isManagedSubjectEnforced` becomes:
```java
return isManagedSubject(subject) && managedEnforced.contains(subject);
```

This halves the per-`DependencyManagement` allocation cost.

maven-resolver-test-util/src/main/java/org/eclipse/aether/internal/test/util/DependencyGraphParser.java
```diff
@@ -291,16 +293,17 @@ private DependencyNode build(DependencyNode parent, LineContext ctx, boolean isR
             DefaultArtifact artifact = new DefaultArtifact(def.coords, def.properties);
             Dependency dependency = new Dependency(artifact, def.scope, def.optional);
             node = new DefaultDependencyNode(dependency);
-            int managedBits = 0;
+            EnumMap<DependencyManagement.Subject, Boolean> managedSubjects =
+                    new EnumMap<>(DependencyManagement.Subject.class);
             if (def.premanagedScope != null) {
-                managedBits |= DependencyNode.MANAGED_SCOPE;
+                managedSubjects.put(DependencyManagement.Subject.SCOPE, true);
```
gnodet, line 299: _Claude Code on behalf of Guillaume Nodet_

Managed subjects from the parser are hardcoded as `enforced=true`. This means graph definition files (`.txt` test resources) cannot express "advised" management — only programmatic test construction can.

Not blocking, since the existing programmatic tests (`enforcedScopeNotDerived`, `advisedScopeIsDerived`, etc.) cover the key scenarios. But if more complex graph-based test scenarios are planned, the parser format could be extended with a syntax like `scope:test!` (enforced) vs `scope:test` (advised).

cstamas:
Applied PR comments except the last one (DependencyGraphParser), as comment says, we can fiddle with it later. Also fixed since tags to point to `2.0.19` (was `2.0.17`)

gnodet (commented):
## AI Review — PR #1815: Advised and Enforced Management

Hi @cstamas, this is a well-iterated PR — the advised/enforced management model is a clean improvement over the binary `managedBits` system, and the backwards-compatibility approach (default methods, bidirectional mapping) is exactly right.

### 🔴 Merge conflict (high)
The PR is currently in a CONFLICTING merge state. CI checks all pass (13/13), so a rebase/merge should be straightforward.

### ⚠️ `setManagedSubjects` undocumented (medium → low)
The new `public void setManagedSubjects(Map<DependencyManagement.Subject, Boolean>)` on `DefaultDependencyNode` lacks Javadoc, `@since` tag, and isn't declared on the `DependencyNode` interface. Since `DefaultDependencyNode` is `final` and the interface is `@noimplement`, the impact is limited — but documenting the `Boolean` semantics (true = enforced, false = advised) would help consumers.

### ⚠️ Type safety trade-off (medium → low)
`DependencyManagement` replaced typed fields with `EnumMap<Subject, Object>` with unchecked casts in getters. However, since `managedValues` is `private final` and only mutated through typed setters, the public API retains type safety. This is a reasonable trade-off given the design goals.

### 💡 Embedded bug fix (low)
The `makeDependencyNode(groupId, artifactId, version, scope)` test utility was passing hardcoded `"compile"` instead of the `scope` parameter — a genuine pre-existing bug now fixed. Consider calling this out in the PR description for traceability.

### 💡 Method rename in internal class (low)
`createDependencyNode` → `createDependencyNodeCycle` in `DependencyCollectorDelegate` (also removing `static`). Since the class is in an `internal` package, this is acceptable — but worth noting for any forks.

**What works well:**
- Eliminating the `isInheritedDerived()` hack with proper advised/enforced semantics
- Solid backwards-compat: `setManagedBits()` maps to new model, `getManagedBits()` reconstructs
- Good test coverage added: root enforcement, transitive advice, conflict resolver scope/optional derivation, backwards-compat round-trip
- Clean incorporation of review feedback (EnumMap, EnumSet, lazy allocation, `isRootManager()` helper)

---
*🤖 This review was generated by [ForgeBot](https://github.com/gnodet/forgebot-skills) using a maker/checker pattern (reviewer + independent verifier). All 5 findings confirmed, with mitigating factors noted.*

gnodet (commented):
## Review Summary

Great work on this, @cstamas! The advised/enforced management model is a clean improvement over the binary `managedBits` approach, and the test coverage added in this iteration is excellent — tests for root enforcement, transitive advice, conflict resolver scope/optional derivation, backwards-compat round-trip, and enforced vs advised subjects are all present and well-structured.

After independent verification of all findings, the PR looks solid:

### Positive Notes

- **Pre-existing bug fix**: The `makeDependencyNode(groupId, artifactId, version, scope)` overload was hardcoding `"compile"` instead of passing the `scope` parameter — good catch fixing this, as it enables the new scope derivation tests to actually work correctly.
- The `createPremanagedDependency` protected method added to `DependencyCollectorDelegate` is a nice extensibility improvement.
- CI is fully green (13/13 checks passing).

### Minor Observation

- `setManagedSubjects` stores the caller's map reference directly without a defensive copy. This is currently safe because `PremanagedDependency` is created per-dependency and `applyTo` is called exactly once, but a `new EnumMap<>(managedSubjects)` in the setter would make the contract more robust for future changes. Not a blocker.

---

gnodet (commented):
Well-designed PR — the architectural approach of replacing binary `managedBits` with a three-state model (not managed / advised / enforced) is sound and correctly addresses the long-standing problem where transitive dependency management would mark scope/optional as "managed", preventing proper inheritance during graph transformation.

The backwards-compatibility mapping (`setManagedBits` → new model, `getManagedBits` → reconstructs bit field) and the `default` methods on `DependencyNode` are well-executed. Test coverage is excellent across root enforcement, transitive advice, conflict resolver scope/optional derivation, backwards-compat round-trip, and null/empty edge cases.

**Two items to address before merge:**

1. **Merge conflict** — the PR is in CONFLICTING state. Three releases (2.0.19, 2.0.20, 2.0.21) have landed since the PR's last update. A rebase onto current master is needed.

2. **Stale `@since` tags** — all `@since 2.0.19` annotations across `DependencyManagement.java`, `DependencyNode.java`, and `PremanagedDependency.java` need to be updated to `@since 2.0.22` since 2.0.19 through 2.0.21 shipped without this PR.

**Minor observations (not blockers):**
- `setManagedSubjects(Map)` on `DefaultDependencyNode` stores the caller's map without a defensive copy. Safe in current usage but a `new EnumMap<>(managedSubjects)` would be more robust.
- `setManagedSubjects` is public but lacks Javadoc and `@since` tag (the method it replaces, `setManagedBits`, was documented).
- The fix in `ConflictResolverTest.makeDependencyNode` replacing hardcoded `"compile"` with the `scope` parameter is a genuine pre-existing bug fix — worth calling out in the PR description for traceability.

gnodet (commented):
Well-designed replacement of binary `managedBits` with an advised/enforced management model. The architecture, backwards compatibility handling, and test coverage are solid. A few items need attention before merge:

**Blocking:**
- ⚠️ **Merge conflict**: The PR's merge state is CONFLICTING. Three releases (2.0.19, 2.0.20, 2.0.21) have landed since the last rebase (June 19). A rebase onto current master is needed.
- ⚠️ **Stale `@since` tags**: All `@since 2.0.19` annotations across `DependencyManagement.java`, `DependencyNode.java`, and `PremanagedDependency.java` need updating to `@since 2.0.22` since those versions shipped without this PR.

**Minor observations:**
- `setManagedSubjects` stores the caller's map reference directly without a defensive copy. Currently safe in practice, but a `new EnumMap<>(managedSubjects)` would prevent aliasing bugs from future callers.
- The fix in `ConflictResolverTest` changing hardcoded `"compile"` to the actual `scope` parameter is a genuine pre-existing bug fix — consider noting this in the PR description for traceability.
- `DependencyGraphParser` always allocates an `EnumMap` even when no managed subjects are present. Minor unnecessary allocation (test-util only).

**Strengths:**
- Backwards compatibility is well-handled: `setManagedBits` maps to the new model, `getManagedBits` reconstructs the bit field, and default methods provide fallback.
- Test coverage is thorough: root enforcement, transitive advice, conflict resolver scope/optional derivation, and backwards-compat round-trip are all tested.
- Properties from system path handling are correctly set as non-enforced.

gnodet (commented):
Well-designed API extension adding advised/enforced management semantics with proper backwards compatibility. The core logic is correct and test coverage is solid. Two minor items:

1. **Missing `@since` annotation**: The new public method `setManagedSubjects(Map<DependencyManagement.Subject, Boolean>)` in `DefaultDependencyNode.java` lacks a `@since 2.0.19` annotation and Javadoc, unlike all other new public API members in this PR.

2. **Pre-existing test bug fix**: The `makeDependencyNode(groupId, artifactId, version, scope)` method in `ConflictResolverTest.java` was ignoring the `scope` parameter and always passing `"compile"`. This valid fix is bundled into the PR but worth noting in the PR description since it affects the existing `derivedScopeChange` test.

PR #1830, opened 2026-04-14
Remove unnecessary repository preparation in metadata resolution for prefixes

slawekjaranowski:
Repository should be already configured, it is too late to change configuration. The same configuration should be used for standard operation and resolving metadata.

Example: plugins can provide unconfigured repo,
when we have configuration for mirror in settings - mirror will be used only for metadata.

---

### DIFF
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/filter/PrefixesRemoteRepositoryFilterSource.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/filter/PrefixesRemoteRepositoryFilterSource.java
@@ -387,10 +387,6 @@ public final class PrefixesRemoteRepositoryFilterSource extends RemoteRepository
         MetadataResolver mr = metadataResolver.get();
         RemoteRepositoryManager rm = remoteRepositoryManager.get();
         if (mr != null && rm != null) {
-            // create "prepared" (auth, proxy and mirror equipped repo)
-            RemoteRepository prepared = rm.aggregateRepositories(
-                            session, Collections.emptyList(), Collections.singletonList(remoteRepository), true)
-                    .get(0);
             // retrieve prefix as metadata from repository
             MetadataResult result = mr.resolveMetadata(
                             new DefaultRepositorySystemSession(session)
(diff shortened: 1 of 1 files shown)


cstamas:
> Example: plugins can provide unconfigured repo,
when we have configuration for mirror in settings - mirror will be used only for metadata.

Are you sure about this? This sounds wrong...

slawekjaranowski:
> > Example: plugins can provide unconfigured repo,
> > when we have configuration for mirror in settings - mirror will be used only for metadata.
> 
> Are you sure about this? This sounds wrong...

I discovered it by using `wildfly-maven-plugin`, which manually add repsitories:

https://github.com/wildfly/wildfly-maven-plugin/blob/3c924bc6d07cccd3954dfd28811364044c1f2f7a/core/src/main/java/org/wildfly/plugin/core/MavenRepositoriesEnricher.java

only proxy is configured here (also manually), 

I have configured mirrors in my settings.

I such case in normal resolutions, mirrors are not used ... but for prefixes metadata we add mirror ...

My debug session show:

image

slawekjaranowski:
Why we need to reconfigure repo ... it should be already done

cstamas:
See https://github.com/apache/maven-resolver/blob/master/maven-resolver-api/src/main/java/org/eclipse/aether/RepositorySystem.java#L312

In fact, see whole Resolver API: all methods _references_ this method above. This is resolver expectation. IF user has set up mirror, it has to be obeyed, I cannot today, but pls ping me on Slack tomorrow to discuss this.

slawekjaranowski:
I know that `newResolutionRepositories` should be used .... but what if autor of plugin doesn't use it? it is my case. 

When repo is correctly created with `newResolutionRepositories` it will be configured and we don't want to do it again.

cstamas:
That is clearly a bug then. In 90% of cases, when plugin operates on received repositories, Maven already primed them. But, if plugin _creates_ new `RemoteRepository` instances, that is a bug.

Moreover, the method `newResolutionRepositories` is not and cannot be idempotent, hence, we (resolver) cannot just invoke that method over and over again, as it will create issues with setups like `mirrorOf: *`.

slawekjaranowski:
I agree that it is a bug in plugin .... but I think that next reconfiguration is not required
and introduce some of inconsistency

when use such repo for standard resolving we use it as is
when used for resolving meta-data we change configuration

slawekjaranowski:
So if:

> Moreover, the method newResolutionRepositories is not and cannot be idempotent, hence, we (resolver) cannot just invoke that method over and over again, as it will create issues with setups like mirrorOf: *.

we also should not call `aggregateRepositories`

cstamas (approved):
Right, is correct. I use "normalized" bare reposes only as keys.

PR #1837, opened 2026-04-16
GH-1835: Expose qualifiers for generic version scheme

### LINKED ISSUE #1835: Preview version check should be exposed via API

Right now, while Resolver supports "preview" versions, it is left to client code (ie. Toolbox) to perform the check. This is clearly a duplication and also the two may drift. Best is Resolver is single source of truth.

cstamas:
This allows nice version selection. Also, provide a version filter that leaves out preview (sans snapshot) versions.

Finally, made generic classes public, they were package protected.

Fixes #1835


maven-resolver-util/src/main/java/org/eclipse/aether/util/version/GenericQualifiers.java
```diff
@@ -0,0 +1,117 @@
...
+     * @param token the string to analyze for qualifier, must not be {@code null}
+     */
+    public static Optional<Integer> qualifier(String token) {
+        // most trivial preview version is "a1"
+        if (token.length() > 1) {
+            String v = token.toLowerCase(Locale.ENGLISH);
+            // simple case: full qualifier label is present (assuming once)
+            for (Map.Entry<String, Integer> entry : QUALIFIERS.entrySet()) {
+                // TODO on map we have empty string "" as key
+                if (!entry.getKey().isEmpty() && v.contains(entry.getKey())) {
```
gnodet: The `v.contains(entry.getKey())` substring check can produce false positives for short qualifier labels. For example:

- `"1.0-arced"` → matches `"rc"` → incorrectly detected as RC
- `"1.0-legacy"` → matches `"ga"` → incorrectly detected as GA
- `"1.0-sacred"` → matches `"cr"` → incorrectly detected as RC
- `"1.0-dispatcher"` → matches `"sp"` → incorrectly detected as SP

In practice Maven versions are fairly structured so the false positive rate may be low, but since this is a new public API (`@since 2.0.17`), it might be worth considering a tokenization-based or word-boundary approach (e.g., splitting on `.`, `-`, `_` first and then doing exact matches on each token — similar to what `tokenQualifier` already does internally for the parsed path).

What do you think?

maven-resolver-util/src/main/java/org/eclipse/aether/util/version/GenericQualifiers.java
```diff
@@ -0,0 +1,117 @@
...
+        QUALIFIERS.put(LABEL_BETA, QUALIFIER_BETA);
+        QUALIFIERS.put(LABEL_MILESTONE, QUALIFIER_MILESTONE);
+        QUALIFIERS.put("rc", QUALIFIER_RC);
+        QUALIFIERS.put("cr", QUALIFIER_RC);
+        QUALIFIERS.put("snapshot", QUALIFIER_SNAPSHOT);
+        QUALIFIERS.put("ga", QUALIFIER_ZERO);
+        QUALIFIERS.put("final", QUALIFIER_ZERO);
+        QUALIFIERS.put("release", QUALIFIER_ZERO);
+        QUALIFIERS.put("", QUALIFIER_ZERO); // TODO: is this entry valid?
+        QUALIFIERS.put("sp", QUALIFIER_SP);
```
gnodet: nit: This TODO was carried over from the original internal code. Now that this class is part of the public API, it might be worth either resolving it (documenting why the empty-string entry exists) or removing it if it's no longer relevant.

maven-resolver-util/src/test/java/org/eclipse/aether/util/graph/version/GenericQualifiersVersionFilterTest.java
```diff
@@ -0,0 +1,70 @@
...
+        filter.filterVersions(ctx);
+        assertVersions(ctx, "1", "3.1", "5.0.0");
+    }
+
+    @Test
+    void testDeriveChildFilter() {
+        GenericQualifiersVersionFilter filter = GenericQualifiersVersionFilter.previewVersionFilter();
+        assertSame(filter, derive(filter, "g:a:1"));
+    }
+}
```
gnodet, line 90: The filter-level tests are good and cover the main scenarios. It might be valuable to add direct unit tests for `GenericQualifiers.qualifier(String)` as well — particularly for edge cases like single-character versions, the short-form detection (`"a1"`, `"b2"`, `"m3"`), and version strings that could trigger the substring false positives mentioned in the other comment.

gnodet (commented):
Nice work extracting the qualifier logic into its own public class — the separation of concerns is clean, and the refactoring in `GenericVersion.Tokenizer` preserves semantic equivalence. A few observations below.

gnodet (approved):
Looks good — clean extraction of qualifier logic into a public API with solid test coverage.

A few non-blocking observations:

1. **Qualifier map type change**: `TreeMap(CASE_INSENSITIVE_ORDER)` → `LinkedHashMap` + explicit `toLowerCase()` — behaviorally equivalent for ASCII labels, just noting the mechanism changed.
2. **`CharsetEncoder` in loop** (`GenericVersionScheme.main()`): could be hoisted above the loop, but negligible since it's CLI code.
3. **`releasePreviewVersionFilter()` naming**: might surprise callers who expect snapshots to be filtered — a brief Javadoc note clarifying snapshots are intentionally retained could help.

Nice improvement to `assertVersions()` diagnostics too.

cstamas:
> 3. **`releasePreviewVersionFilter()` naming**: might surprise callers who expect snapshots to be filtered — a brief Javadoc note clarifying snapshots are intentionally retained could help.

Agreed, and dropped 3rd filter that handled snapshots too, in hope to lessen confusion (multiple filters dealing with snapshots), as while snapshots _are qualifier_, they are also well known by users and well documented in Maven.

Basically kept these two:
* preview - "alpha", "beta" and "milestone", these are already documented as "preview" versions in Maven doco, also are widely used (see site M releases).
* pre-release - preview + "rc/cr"; so these qualifiers are all pre-release, and while they do not cover snapshots (another pre-release) for above mentioned reasons, the filter is not dealing with them.

PR #1838, opened 2026-04-16
Add predicate-delegating filter

cstamas:
A conditional filter, that applies delegate only if predicate applies.


cstamas:
Merged into cleanup that is part of #1837

PR #1841, opened 2026-04-19
Lower various component selection debug noise

cstamas:
They all emit stack traces, and just make noise.
Make clear at DEBUG what is happening, while retain ability for stack traces at TRACE level.


hboutemy (approved):
will make normal debug traces much more clear

PR #1852, opened 2026-04-22
[MRESOLVER-686] Backport version comparison fix to 1.9.x

bentatham:
Fixes `...293.225.2.0-dev for version range null: Comparison method violates its general contract!` when there is a modifier on the version (eg when using `mvn versions:update-parent`).

When two versions differ in kind at index 0 (e.g. a string-prefixed branch-style version versus a numeric release whose first segment is zero), comparePadding with a number filter may return 0 even though the versions are clearly different. That makes equality non-transitive and breaks TimSort when DefaultVersionRangeResolver sorts metadata lists mixing both shapes.

This is the same fix as MRESOLVER-336, which was applied to 2.x but never landed on the 1.9.x line that ships with Maven 3.9.x.

Adds two regression tests covering the three-way transitivity violation and a representative sort.

- [X] Your pull request should address just one issue, without pulling in other changes.
- [X] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [X] Each commit in the pull request should have a meaningful subject line and body.
- [X] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied.
- [X] Run `mvn verify` to make sure basic checks pass.
- [X] You have run the integration tests successfully (`mvn -Prun-its verify`).

- [X] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
- [X] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

### DIFF
--- a/maven-resolver-util/src/main/java/org/eclipse/aether/util/version/GenericVersion.java
+++ b/maven-resolver-util/src/main/java/org/eclipse/aether/util/version/GenericVersion.java
@@ -122,6 +122,9 @@ final class GenericVersion implements Version {
             Item thatItem = those.get(index);
 
             if (thisItem.isNumber() != thatItem.isNumber()) {
+                if (index == 0) {
+                    return thisItem.compareTo(thatItem);
+                }
                 if (number == thisItem.isNumber()) {
                     return comparePadding(these, index, number);
                 } else {
--- a/maven-resolver-util/src/test/java/org/eclipse/aether/util/version/GenericVersionTest.java
+++ b/maven-resolver-util/src/test/java/org/eclipse/aether/util/version/GenericVersionTest.java
@@ -19,7 +19,9 @@
 package org.eclipse.aether.util.version;
 
 import java.util.ArrayList;
+import java.util.Arrays;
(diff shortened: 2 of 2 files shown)


cstamas:
Thanks for porting this!
The bad news: Resolver 1.9.x is being EOLd, along with Maven 3.9.x line, once first [Maven 3.10.0](https://github.com/apache/maven/milestone/126) is out (that uses Resolver 2.x).. So I cannot promise will there be new release of Resolver 1.9.x with this change...

bentatham:
Ok, that's great.  So the underlying bug will be fixed in Maven 3.10.0 then, I believe.

cstamas:
You can already try it, CI deploys snapshots of 3.10.x (choose latest by timestamp):
https://repository.apache.org/content/repositories/snapshots/org/apache/maven/apache-maven/3.10.0-SNAPSHOT/

cstamas (approved):

slachiewicz (approved):

Issue #1743, opened 2026-01-12
IPC named locks on Windows has unstable tests


cstamas:
### Affected version

master

### Bug description

It seems that IPC tests are unstable on Windows: quite often on Windows CI nodes this happens:
```
Error:  Tests run: 9, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 16.16 s <<< FAILURE! -- in org.eclipse.aether.named.ipc.IpcNamedLockFactoryIT
Error:  org.eclipse.aether.named.ipc.IpcNamedLockFactoryIT.exclusiveAccess(TestInfo) -- Time elapsed: 15.01 s <<< ERROR!
java.util.concurrent.TimeoutException: exclusiveAccess(org.junit.jupiter.api.TestInfo) timed out after 15 seconds
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1604)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1604)
	Suppressed: java.lang.InterruptedException
		at java.base/java.lang.Object.wait0(Native Method)
		at java.base/java.lang.Object.wait(Object.java:389)
		at java.base/java.lang.Thread.join(Thread.java:1887)
		at java.base/java.lang.Thread.join(Thread.java:1963)
		at org.eclipse.aether.named.ipc.NamedLockFactoryTestSupport.exclusiveAccess(NamedLockFactoryTestSupport.java:147)
		at java.base/java.lang.reflect.Method.invoke(Method.java:565)
		... 2 more
```

cstamas:
And again today on windows w/ Java 21:
```
[INFO] Running org.eclipse.aether.named.ipc.IpcAdapterIT
Warning:  [stderr] OpenJDK 64-Bit Server VM warning: Sharing is only supported for boot loader classes because bootstrap classpath has been appended
Error:  Tests run: 9, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 17.92 s <<< FAILURE! -- in org.eclipse.aether.named.ipc.IpcAdapterIT
Error:  org.eclipse.aether.named.ipc.IpcAdapterIT.exclusiveAccess -- Time elapsed: 16.32 s <<< ERROR!
java.util.concurrent.TimeoutException: exclusiveAccess() timed out after 15 seconds
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
[...]

PR #1859, opened 2026-05-04
Fix: Remove javax.inject dependency

### LINKED ISSUE #1858: Missing `javax.inject.Provider` from `TrackingFileManagerProvider`
### Affected version

2.0.17

### Bug description

After updating to a new version of a dependency `maven-resolver-supplier-mvn4` to `2.0.17`, our build started to fail due to `java.lang.NoClassDefFoundError: javax/inject/Provider`.
I did a quick research using Claude Code, here is a snipped of its output:

> Here's what happened: PR [#1814](https://github.com/apache/maven-resolver/pull/1814) ("Provide modern TrackingFileManager"), merged for 2.0.17, introduced a new TrackingFileManagerProvider class that implements javax.inject.Provider. The RepositorySystemSupplier.createTrackingFileManager() method at line 232 references this class, which triggers class loading of javax.inject.Provider even when using the Service Loader (non-DI) path. This is the same category of bug as MRESOLVER-298, but in different code introduced by a different change.

Our build is indeed fixed by adding:
```
    <dependency>
      <groupId>javax.inject</groupId>
      <artifactId>javax.inject</artifactId>
      <version>1</version>
      <scope>runtime</scope>
    </dependency>
```
Could you verify this?

cstamas:
As it was wrongly introduced in PM #1814. Core of resolver object graph declares javax.inject as optional, but the mentioned PR made it mandatory.

Fixes #1858

### DIFF
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/LegacyTrackingFileManager.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/LegacyTrackingFileManager.java
@@ -50,7 +50,7 @@ import org.slf4j.LoggerFactory;
  *
  * @since 2.0.17
  * @see NamedLocksTrackingFileManager
- * @see TrackingFileManagerProvider
+ * @see TrackingFileManagerSupplier
  */
 public final class LegacyTrackingFileManager implements TrackingFileManager {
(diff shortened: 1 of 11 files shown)


cstamas:
* macOS JDK 25 got from central `HTTP Status: 403`
* Windows JDK 21 got IPC error #1743 (anyone using/having Windows can lend a hand?)

slawekjaranowski (approved):

PR #1878, opened 2026-05-13
Feat: Reuse same RemoteRepository to HTTP URI creation logic

cstamas:
All HTTP remote repositories have pretty much same or similar expectations regardind base URI of a HTTP remote repository.

### DIFF
--- a/maven-resolver-util/src/main/java/org/eclipse/aether/util/connector/transport/http/HttpTransporterUtils.java
+++ b/maven-resolver-util/src/main/java/org/eclipse/aether/util/connector/transport/http/HttpTransporterUtils.java
@@ -304,6 +304,17 @@ public final class HttpTransporterUtils {
 
     /**
      * Shared code to create "base {@link URI}" for most common HTTP remote repositories and all HTTP transports.
+     * Note: this method just applies common validation and adjustments to URI, but it does not enforce protocol
+     * to be HTTP/HTTPS!
+     * <p>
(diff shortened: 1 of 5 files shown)


kwin (approved):

maven-resolver-util/src/main/java/org/eclipse/aether/util/connector/transport/http/HttpTransporterUtils.java
```diff
@@ -299,4 +301,30 @@ public static Optional<InetAddress> getHttpLocalAddress(
...
+            throw new URISyntaxException(repository.getUrl(), "URL must not be opaque");
+        }
+        if (uri.getRawFragment() != null || uri.getRawQuery() != null) {
+            throw new URISyntaxException(repository.getUrl(), "URL must not have fragment or query");
+        }
+        String path = uri.getRawPath();
+        if (path == null) {
+            path = "/";
+        }
+        if (!path.startsWith("/")) {
```
kwin, line 333: I don't think this can ever happen! Path is according to https://www.rfc-editor.org/rfc/rfc3986#section-3.3 either empty or starts with "/".

cstamas: Yes, the "empty" case is what we want, or maybe not (is "empty" == null?)

PR #1891, opened 2026-05-27
fix: Configuration to be able to control sending RFC 9457 related Accept HTTP header

cstamas:
Some servers have issues handling it (esp the Accept HTTP header), so make it possible to disable it. Moreover, the new config accepts REPO_ID suffix, so this can happen per-server basis.

### DIFF
--- a/maven-resolver-api/src/main/java/org/eclipse/aether/ConfigurationProperties.java
+++ b/maven-resolver-api/src/main/java/org/eclipse/aether/ConfigurationProperties.java
@@ -597,24 +597,27 @@ public final class ConfigurationProperties {
     public static final boolean DEFAULT_VERSION_SCHEME_CACHE_DEBUG = false;
 
     /**
-     * Boolean flag should the HTTP transport support RFC 9457 messages. Some servers have issues with handling the
-     * HTTP Accept headers, when support is enabled. Known servers not supporting RFC 9457 is for
-     * example Sonatype Nexus 2.
+     * Boolean flag should the HTTP transport send HTTP <pre>Accept</pre> header to signal that it supports RFC 9457
+     * error messages. Some servers have issues with handling the HTTP Accept headers. Known servers not supporting it
+     * is for example Staging REST endpoint of Sonatype Nexus 2.
+     * This configuration is only about sending (or not sending) HTTP header with content of
+     * <pre>Accept: application/problem+json</pre> that may be required by some servers to trigger error reporting
+     * using RFC 9457.
      *
      * @since 2.0.19
      * @configurationSource {@link RepositorySystemSession#getConfigProperties()}
      * @configurationType {@link java.lang.Boolean}
-     * @configurationDefaultValue {@link #DEFAULT_HTTP_SUPPORT_RFC9457}
+     * @configurationDefaultValue {@link #DEFAULT_HTTP_SEND_RFC9457_ACCEPT}
      * @configurationRepoIdSuffix Yes
      */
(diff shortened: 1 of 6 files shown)


maven-resolver-api/src/main/java/org/eclipse/aether/ConfigurationProperties.java
```diff
@@ -596,6 +596,26 @@ public final class ConfigurationProperties {
...
+     * HTTP Accept headers, when support is enabled. Known servers not supporting RFC 9457 is for
+     * example Sonatype Nexus 2.
+     *
+     * @since 2.0.19
+     * @configurationSource {@link RepositorySystemSession#getConfigProperties()}
+     * @configurationType {@link java.lang.Boolean}
+     * @configurationDefaultValue {@link #DEFAULT_HTTP_SUPPORT_RFC9457}
+     * @configurationRepoIdSuffix Yes
+     */
+    public static final String HTTP_SUPPORT_RFC9457 = PREFIX_TRANSPORT_HTTP + "supportRfc9457";
```
kwin: This is a confusing comment to me, as this is just about the accept header which is not actually specified in the RFC. I would therefore rather refer to it as EXPLICITLY_ACCEPT_PROBLEM_JSON

kwin (changes requested):
In addition we need to add Accept for regular content (either wildcard for all requests or xml/binary depending on whether it is metadata or artifacts) when we add the accept for problem+json

maven-resolver-transport-apache/src/main/java/org/eclipse/aether/transport/apache/ApacheTransporter.java
```diff
@@ -523,10 +530,23 @@ private <T extends HttpUriRequest> void resume(T request, GetTask task) throws I
     private void handleStatus(CloseableHttpResponse response) throws Exception {
         int status = response.getStatusLine().getStatusCode();
         if (status >= 300) {
-            ApacheRFC9457Reporter.INSTANCE.generateException(response, (statusCode, reasonPhrase) -> {
-                throw new HttpResponseException(statusCode, reasonPhrase + " (" + statusCode + ")");
-            });
+            if (supportRfc9457) {
+                ApacheRFC9457Reporter.INSTANCE.generateException(response, (statusCode, reasonPhrase) -> {
+                    throw new HttpResponseException(statusCode, reasonPhrase + " (" + statusCode + ")");
+                });
```
slawekjaranowski: maybe we can always parse response?

maven-resolver-api/src/main/java/org/eclipse/aether/ConfigurationProperties.java
cstamas: This and @slawekjaranowski comment below makes sense, this config should be ONLY about `Accept` header, and then as Slawek says below, just always parse response?

cstamas:
> In addition we need to add Accept for regular content (either wildcard for all requests or xml/binary depending on whether it is metadata or artifacts) when we add the accept for problem+json

I find this problematic: adding xml or json or any other Accept is not right; transport is only about _artifacts_, and as we know, servers knows to be picky. So, ideally I'd just modific the text of this config param (it is ONLY about adding Accept to fully support RFC 9457), because _historically_ Resolver _never sent any Accept header at all_.

cstamas:
Also, in this very case, it just happens that we use a REST endpoint of an older software. Usually, Resolver transport is hitting non REST.

And this happens ONLY as nx2 offered "transparent support" for maven-deploy-plugin, while deployment to staging was always possible via alternate means (native plugin or so).

kwin (approved):
I would still add `*/*` to accept header values (in case they are sent) but can be done in a separate PR.

PR #1894, opened 2026-05-29
Add any media type to RFC9457 content type header

slawekjaranowski:
Tested with ASF nexus 2.x

### DIFF
--- a/maven-resolver-spi/src/main/java/org/eclipse/aether/spi/connector/transport/http/RFC9457/RFC9457Reporter.java
+++ b/maven-resolver-spi/src/main/java/org/eclipse/aether/spi/connector/transport/http/RFC9457/RFC9457Reporter.java
@@ -36,7 +36,7 @@ import java.io.IOException;
  */
 public abstract class RFC9457Reporter<T, E extends Exception, R> {
     public static final String CONTENT_TYPE_PROBLEM_DETAILS_JSON = "application/problem+json";
-    public static final String CONTENT_TYPE_PROBLEM_DETAILS_JSON_WITH_ANY = "application/problem+json, */*";
+    public static final String CONTENT_TYPE_PROBLEM_DETAILS_JSON_AND_ANY = "application/problem+json,*/*";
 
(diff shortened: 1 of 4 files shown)


slawekjaranowski:
https://github.com/apache/maven-resolver/pull/1891#pullrequestreview-4372174348

maven-resolver-spi/src/main/java/org/eclipse/aether/spi/connector/transport/http/RFC9457/RFC9457Reporter.java
```diff
@@ -36,6 +36,7 @@
  */
 public abstract class RFC9457Reporter<T, E extends Exception, R> {
     public static final String CONTENT_TYPE_PROBLEM_DETAILS_JSON = "application/problem+json";
+    public static final String CONTENT_TYPE_PROBLEM_DETAILS_JSON_WITH_ANY = "application/problem+json, */*";
```
kwin:
```suggestion
    public static final String CONTENT_TYPE_PROBLEM_DETAILS_JSON_AND_ANY = "application/problem+json,*/*";
```

Also no space in CSV!

kwin (approved):
Just one suggestion, otherwise looks good.

cstamas (approved):

olamy:
interesting I didn't know `*/*` was not including `application/problem+json` :)  
Did it work with only `*/*`?
Looks more like a Nexus bug? (any reference?)  (or some Jetty issue on Nexus server side?)
Well on the other side this mean I prefer this `application/problem+json` but otherwise no worries I can accept everything

olamy (approved):

cstamas:
This is not Jetty but Restlet (ancient 1.1.x), and is set to "strict" mode, which means "do not assume anything else than client states".

kwin:
Cloudflare needs explicit problem+json (https://blog.cloudflare.com/rfc-9457-agent-error-pages/), other servers probably as well.

PR #1902, opened 2026-06-07
Improve thread safety across modules

gnodet:
## Thread Safety Improvements

This PR addresses 31 thread safety issues identified through a systematic audit of the codebase. Each fix is a separate commit for easy review.

### HIGH severity fixes (6)
- **IpcClient close/reconnect lifecycle** — reset `initialized` flag in `close()` to allow proper reconnection
- **IpcClient field visibility** — make `socket`/`output`/`input` volatile for cross-thread visibility
- **IpcServer clients HashMap** — use `ConcurrentHashMap` instead of plain `HashMap`
- **IpcNamedLock context leak** — send unlock on lock timeout to prevent server-side context leak
- **BasicAuthCache** — use `ConcurrentHashMap`-backed `AuthCache` to prevent corruption with preemptive auth
- **SmartExecutor RejectedExecutionException** — catch REE to prevent `await()` hanging forever

### MEDIUM severity fixes (15)
- **DataPool cache initialization** — use `computeIfAbsent` instead of get-then-put race
- **WeakInternPool.intern()** — use `compute()` for atomic check-and-update
- **Results.addException/addCycle** — synchronize for `parallelStream` safety
- **DeferredCredentialsProvider** — extend synchronized block to cover delegate read
- **WagonTransporter close-while-executing** — track in-flight wagons for proper cleanup
- **DefaultProxySelector** — use `CopyOnWriteArrayList`
- **DefaultMirrorSelector** — use `CopyOnWriteArrayList`
- **DefaultAuthenticationSelector** — use `ConcurrentHashMap`
- **SimpleArtifactTypeRegistry** — use `ConcurrentHashMap`
- **CachingArtifactTypeRegistry** — use `ConcurrentHashMap.computeIfAbsent()`
- **IpcServer.expirationCheck()** — guard against negative `Thread.sleep()` argument
- **IpcServer Lock memory leak** — remove empty locks from map after unlock
- **IpcServer Lock.unlock()** — complete futures outside monitor to avoid I/O under lock
- **NamedLockFactorySupport getLock/shutdown** — use ReadWriteLock to prevent race
- **Redisson trySetPermits()** — log warning when return value is false

### LOW severity fixes (10)
- **GenericVersionScheme** — fix racy cache statistics
- **Results.errorPath** — add volatile
- **BasicRepositoryConnectorFactory.priority** — add volatile
- **TransferResource.contentLength/resumeOffset** — add volatile
- **Lazy hashCode fields** — add volatile in 4 selector/traverser classes
- **ChainedWorkspaceReader.repository** — add volatile
- **CompositeNamedLock** — remove redundant double `unlockAll()`
- **WagonTransporter pollWagon** — release wagon on reconnect failure
- **MinioTransporter implPut** — close InputStream properly
- **GnupgSignatureArtifactGenerator** — use `CopyOnWriteArrayList`

### Not addressed (4)
- **F-05** (IpcServer phantom lock on partial multi-key acquisition) — complex architectural change; deferred
- **F-23** (WebDAV detection redundant OPTIONS) — benign redundancy, accepted
- **F-28** (DefaultDependencyNode children aliasing) — risk of functional regression
- **F-34** (File transport TOCTOU) — known limitation of filesystem operations

### DIFF
--- a/maven-resolver-api/src/main/java/org/eclipse/aether/transfer/TransferResource.java
+++ b/maven-resolver-api/src/main/java/org/eclipse/aether/transfer/TransferResource.java
@@ -46,9 +46,9 @@ public final class TransferResource {
 
     private final RequestTrace trace;
 
-    private long contentLength = -1L;
+    private volatile long contentLength = -1L;
 
-    private long resumeOffset;
+    private volatile long resumeOffset;
 
     public static Clock getClock() {
         return clock;
--- a/maven-resolver-connector-basic/src/main/java/org/eclipse/aether/connector/basic/BasicRepositoryConnectorFactory.java
+++ b/maven-resolver-connector-basic/src/main/java/org/eclipse/aether/connector/basic/BasicRepositoryConnectorFactory.java
@@ -57,7 +57,7 @@ public final class BasicRepositoryConnectorFactory implements RepositoryConnecto
 
     private final Map<String, ProvidedChecksumsSource> providedChecksumsSources;
 
-    private float priority;
+    private volatile float priority;
 
     @Inject
     public BasicRepositoryConnectorFactory(
--- a/maven-resolver-generator-gnupg/src/main/java/org/eclipse/aether/generator/gnupg/GnupgSignatureArtifactGenerator.java
+++ b/maven-resolver-generator-gnupg/src/main/java/org/eclipse/aether/generator/gnupg/GnupgSignatureArtifactGenerator.java
@@ -28,7 +28,6 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
-import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.function.Predicate;
 
 import org.bouncycastle.bcpg.ArmoredOutputStream;
@@ -50,7 +49,7 @@ import org.slf4j.LoggerFactory;
 final class GnupgSignatureArtifactGenerator implements ArtifactGenerator {
     private static final String ARTIFACT_EXTENSION = ".asc";
     private final Logger logger = LoggerFactory.getLogger(getClass());
-    private final ArrayList<Artifact> artifacts;
+    private final List<Artifact> artifacts;
     private final Predicate<Artifact> signableArtifactPredicate;
     private final PGPSecretKey secretKey;
     private final PGPPrivateKey privateKey;
@@ -71,7 +70,7 @@ final class GnupgSignatureArtifactGenerator implements ArtifactGenerator {
         this.privateKey = privateKey;
         this.hashSubPackets = hashSubPackets;
         this.keyInfo = keyInfo;
-        this.signatureTempFiles = new CopyOnWriteArrayList<>();
+        this.signatureTempFiles = new ArrayList<>();
         logger.debug("Created generator using key {}", keyInfo);
     }
 
@@ -81,7 +80,7 @@ final class GnupgSignatureArtifactGenerator implements ArtifactGenerator {
     }
 
     @Override
-    public Collection<? extends Artifact> generate(Collection<? extends Artifact> generatedArtifacts) {
+    public synchronized Collection<? extends Artifact> generate(Collection<? extends Artifact> generatedArtifacts) {
         try {
             artifacts.addAll(generatedArtifacts);
 
@@ -116,7 +115,7 @@ final class GnupgSignatureArtifactGenerator implements ArtifactGenerator {
     }
 
     @Override
-    public void close() {
+    public synchronized void close() {
         signatureTempFiles.forEach(p -> {
             try {
                 Files.deleteIfExists(p);
--- a/maven-resolver-generator-gnupg/src/main/java/org/eclipse/aether/generator/gnupg/GnupgSignatureArtifactGenerator.java
+++ b/maven-resolver-generator-gnupg/src/main/java/org/eclipse/aether/generator/gnupg/GnupgSignatureArtifactGenerator.java
@@ -27,6 +27,8 @@ import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.function.Predicate;
 
 import org.bouncycastle.bcpg.ArmoredOutputStream;
@@ -54,7 +56,7 @@ final class GnupgSignatureArtifactGenerator implements ArtifactGenerator {
     private final PGPPrivateKey privateKey;
     private final PGPSignatureSubpacketVector hashSubPackets;
     private final String keyInfo;
-    private final ArrayList<Path> signatureTempFiles;
+    private final List<Path> signatureTempFiles;
 
     GnupgSignatureArtifactGenerator(
             Collection<Artifact> artifacts,
@@ -69,7 +71,7 @@ final class GnupgSignatureArtifactGenerator implements ArtifactGenerator {
         this.privateKey = privateKey;
         this.hashSubPackets = hashSubPackets;
         this.keyInfo = keyInfo;
-        this.signatureTempFiles = new ArrayList<>();
+        this.signatureTempFiles = new CopyOnWriteArrayList<>();
         logger.debug("Created generator using key {}", keyInfo);
     }
 
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/CachingArtifactTypeRegistry.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/CachingArtifactTypeRegistry.java
@@ -18,8 +18,8 @@
  */
 package org.eclipse.aether.internal.impl.collect;
 
-import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.eclipse.aether.RepositorySystemSession;
 import org.eclipse.aether.artifact.ArtifactType;
@@ -45,17 +45,10 @@ public class CachingArtifactTypeRegistry implements ArtifactTypeRegistry {
 
     private CachingArtifactTypeRegistry(ArtifactTypeRegistry delegate) {
         this.delegate = delegate;
-        types = new HashMap<>();
+        types = new ConcurrentHashMap<>();
     }
 
     public ArtifactType get(String typeId) {
-        ArtifactType type = types.get(typeId);
-
-        if (type == null) {
-            type = delegate.get(typeId);
-            types.put(typeId, type);
-        }
-
-        return type;
+        return types.computeIfAbsent(typeId, delegate::get);
     }
 }
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/DataPool.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/DataPool.java
@@ -180,53 +180,34 @@ public final class DataPool {
         internArtifactDescriptorManagedDependencies = ConfigUtils.getBoolean(
                 session, true, CONFIG_PROP_COLLECTOR_POOL_INTERN_ARTIFACT_DESCRIPTOR_MANAGED_DEPENDENCIES);
 
-        InternPool<Artifact, Artifact> artifactsPool = null;
-        InternPool<Dependency, Dependency> dependenciesPool = null;
-        InternPool<DescriptorKey, Descriptor> descriptorsPool = null;
-        InternPool<List<Dependency>, List<Dependency>> dependencyListsPool = null;
+        InternPool<Artifact, Artifact> artifactsPool;
+        InternPool<Dependency, Dependency> dependenciesPool;
+        InternPool<DescriptorKey, Descriptor> descriptorsPool;
+        InternPool<List<Dependency>, List<Dependency>> dependencyListsPool;
         if (cache != null) {
-            artifactsPool = (InternPool<Artifact, Artifact>) cache.get(session, ARTIFACT_POOL);
-            dependenciesPool = (InternPool<Dependency, Dependency>) cache.get(session, DEPENDENCY_POOL);
-            descriptorsPool = (InternPool<DescriptorKey, Descriptor>) cache.get(session, DESCRIPTORS);
+            artifactsPool = (InternPool<Artifact, Artifact>) cache.computeIfAbsent(
+                    session,
+                    ARTIFACT_POOL,
+                    () -> createPool(ConfigUtils.getString(session, WEAK, CONFIG_PROP_COLLECTOR_POOL_ARTIFACT)));
+            dependenciesPool = (InternPool<Dependency, Dependency>) cache.computeIfAbsent(
+                    session,
+                    DEPENDENCY_POOL,
+                    () -> createPool(ConfigUtils.getString(session, WEAK, CONFIG_PROP_COLLECTOR_POOL_DEPENDENCY)));
+            descriptorsPool = (InternPool<DescriptorKey, Descriptor>) cache.computeIfAbsent(
+                    session,
+                    DESCRIPTORS,
+                    () -> createPool(ConfigUtils.getString(session, HARD, CONFIG_PROP_COLLECTOR_POOL_DESCRIPTOR)));
+            dependencyListsPool = (InternPool<List<Dependency>, List<Dependency>>) cache.computeIfAbsent(
+                    session,
+                    DEPENDENCY_LISTS_POOL,
+                    () -> createPool(
+                            ConfigUtils.getString(session, HARD, CONFIG_PROP_COLLECTOR_POOL_DEPENDENCY_LISTS)));
+        } else {
+            artifactsPool = createPool(ConfigUtils.getString(session, WEAK, CONFIG_PROP_COLLECTOR_POOL_ARTIFACT));
+            dependenciesPool = createPool(ConfigUtils.getString(session, WEAK, CONFIG_PROP_COLLECTOR_POOL_DEPENDENCY));
+            descriptorsPool = createPool(ConfigUtils.getString(session, HARD, CONFIG_PROP_COLLECTOR_POOL_DESCRIPTOR));
             dependencyListsPool =
-                    (InternPool<List<Dependency>, List<Dependency>>) cache.get(session, DEPENDENCY_LISTS_POOL);
-        }
-
-        if (artifactsPool == null) {
-            String artifactPoolType = ConfigUtils.getString(session, WEAK, CONFIG_PROP_COLLECTOR_POOL_ARTIFACT);
-
-            artifactsPool = createPool(artifactPoolType);
-            if (cache != null) {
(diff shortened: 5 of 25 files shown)


gnodet:
## Remaining Findings Analysis (F-05, F-23, F-28, F-34)

The 4 findings not addressed in this PR were analyzed in detail against the actual source code:

| Finding | Severity | Verdict | Rationale |
|---------|----------|---------|-----------|
| F-05 | HIGH | **SKIP — False positive** | `Lock.unlock()` dual-path logic (holders + waiters removal) under a single monitor already prevents phantom locks. The `thenRun` callback only writes to sockets — it cannot re-add a context to any holders list. |
| F-23 | MEDIUM | **SKIP — Benign race** | Redundant idempotent OPTIONS requests at worst. The `volatile Boolean` ensures visibility. Synchronizing around an HTTP round-trip would be worse — serializing concurrent PUTs behind network I/O. |
| F-28 | LOW | **DEFER** | Valid list aliasing through DataPool cache, but intentional for memory optimization. A 2-line defensive copy fix exists if wanted, but no known bug reports and existing transformers already partially defend against it. |
| F-34 | LOW | **SKIP — Not a bug** | Universal filesystem TOCTOU inherent to all Java NIO. Same `createDirectories` + `newOutputStream` pattern used in 4+ places in the codebase. Standard idiomatic Java. |

Full analysis: see `resolver-remaining-findings-analysis.md` in the audit artifacts.

slachiewicz (approved):

maven-resolver-util/src/main/java/org/eclipse/aether/util/concurrency/SmartExecutor.java
```diff
@@ -84,32 +85,44 @@ class Pooled implements SmartExecutor {
...
+                    Thread.currentThread().setContextClassLoader(tccl);
+                    try {
+                        runnable.run();
+                    } finally {
+                        Thread.currentThread().setContextClassLoader(old);
+                    }
+                });
+            } catch (RejectedExecutionException e) {
+                runnable.run();
+            }
```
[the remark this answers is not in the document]
gnodet: Fixed. The fallback `runnable.run()` is now wrapped in a `try-catch(RuntimeException | Error)` that swallows the exception. This preserves the async `submit()` semantics where exceptions are captured by the Future. Callers like `RunnableErrorForwarder` already record the error before re-throwing, so it will still be reported via `await()`.

maven-resolver-util/src/main/java/org/eclipse/aether/util/concurrency/SmartExecutor.java
```diff
@@ -136,13 +149,21 @@ class Limited implements SmartExecutor {
...
+                            runnable.run();
+                        } finally {
+                            semaphore.release();
+                        }
+                    });
+                } catch (RejectedExecutionException e) {
                     try {
                         runnable.run();
                     } finally {
                         semaphore.release();
```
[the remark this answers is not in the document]
gnodet: Fixed. Same approach as the `Pooled` case above — the fallback `runnable.run()` in `Limited.submit(Runnable)` now catches and swallows `RuntimeException | Error` to match async submission semantics. The semaphore is still properly released in the `finally` block.

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/CachingArtifactTypeRegistry.java
```diff
@@ -45,17 +45,18 @@ public static ArtifactTypeRegistry newInstance(ArtifactTypeRegistry delegate) {
...
     }
 
     public ArtifactType get(String typeId) {
         ArtifactType type = types.get(typeId);
-
-        if (type == null) {
-            type = delegate.get(typeId);
+        if (type != null) {
+            return type;
+        }
```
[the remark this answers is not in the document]
gnodet: Fixed. Replaced the get-then-delegate-then-put sequence with `types.computeIfAbsent(typeId, delegate::get)` for atomic cache population. This avoids redundant `delegate.get()` calls under contention.

maven-resolver-generator-gnupg/src/main/java/org/eclipse/aether/generator/gnupg/GnupgSignatureArtifactGenerator.java
```diff
@@ -63,13 +65,13 @@ final class GnupgSignatureArtifactGenerator implements ArtifactGenerator {
...
-        this.artifacts = new ArrayList<>(artifacts);
+        this.artifacts = new CopyOnWriteArrayList<>(artifacts);
         this.signableArtifactPredicate = signableArtifactPredicate;
         this.secretKey = secretKey;
         this.privateKey = privateKey;
         this.hashSubPackets = hashSubPackets;
         this.keyInfo = keyInfo;
-        this.signatureTempFiles = new ArrayList<>();
+        this.signatureTempFiles = new CopyOnWriteArrayList<>();
         logger.debug("Created generator using key {}", keyInfo);
```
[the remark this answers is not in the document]
gnodet: Fixed. Replaced `CopyOnWriteArrayList` with plain `ArrayList` for both `artifacts` and `signatureTempFiles`, and made `generate()` and `close()` `synchronized`. This avoids the array-copy overhead on each write while still providing thread safety and happens-before guarantees between the two methods.

maven-resolver-api/src/main/java/org/eclipse/aether/transfer/TransferResource.java
```diff
@@ -46,9 +46,9 @@ public final class TransferResource {
 
     private final RequestTrace trace;
 
-    private long contentLength = -1L;
+    private volatile long contentLength = -1L;
```
cstamas, line 49: why?

maven-resolver-connector-basic/src/main/java/org/eclipse/aether/connector/basic/BasicRepositoryConnectorFactory.java
```diff
@@ -57,7 +57,7 @@ public final class BasicRepositoryConnectorFactory implements RepositoryConnecto
 
     private final Map<String, ProvidedChecksumsSource> providedChecksumsSources;
 
-    private float priority;
+    private volatile float priority;
```
cstamas, line 60: why?

maven-resolver-generator-gnupg/src/main/java/org/eclipse/aether/generator/gnupg/GnupgSignatureArtifactGenerator.java
```diff
@@ -79,7 +80,7 @@ public String generatorId() {
     }
 
     @Override
-    public Collection<? extends Artifact> generate(Collection<? extends Artifact> generatedArtifacts) {
+    public synchronized Collection<? extends Artifact> generate(Collection<? extends Artifact> generatedArtifacts) {
```
cstamas, line 83: Why?

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/DataPool.java
```diff
@@ -582,16 +563,23 @@ public V get(K key) {
...
+        @SuppressWarnings("unchecked")
         public V intern(K key, V value) {
-            WeakReference<V> pooledRef = map.get(key);
-            if (pooledRef != null) {
-                V pooled = pooledRef.get();
-                if (pooled != null) {
-                    return pooled;
-                }
+            Object[] result = new Object[1];
+            synchronized (map) {
```
cstamas, line 570: Why?

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/DependencyCollectorDelegate.java
```diff
@@ -550,7 +550,7 @@ protected static class Results {
 
         final int maxCycles;
 
-        String errorPath;
+        volatile String errorPath;
```
cstamas, line 553: Why?

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/selector/AndDependencySelector.java
```diff
@@ -39,7 +39,7 @@ public final class AndDependencySelector implements DependencySelector {
 
     private final Set<? extends DependencySelector> selectors;
 
-    private int hashCode;
+    private volatile int hashCode;
```
cstamas: The `hashCode()` is not concurrency safe here, so I'd make this `final` instead and modify `hashCode()` to just return the `final int hashCode` instead.

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/selector/ExclusionDependencySelector.java
```diff
@@ -41,7 +41,7 @@ public final class ExclusionDependencySelector implements DependencySelector {
     // sorted and dupe-free array, faster to iterate than LinkedHashSet
     private final Exclusion[] exclusions;
 
-    private int hashCode;
+    private volatile int hashCode;
```
cstamas: The `hashCode()` is not concurrency safe here, so I'd make this `final` instead and modify `hashCode()` to just return the `final int hashCode` instead.

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/traverser/AndDependencyTraverser.java
```diff
@@ -38,7 +38,7 @@ public final class AndDependencyTraverser implements DependencyTraverser {
 
     private final Set<? extends DependencyTraverser> traversers;
 
-    private int hashCode;
+    private volatile int hashCode;
```
cstamas: The `hashCode()` is not concurrency safe here, so I'd make this `final` instead and modify `hashCode()` to just return the `final int hashCode` instead.

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/version/ChainedVersionFilter.java
```diff
@@ -33,7 +33,7 @@ public class ChainedVersionFilter implements VersionFilter {
 
     private final VersionFilter[] filters;
 
-    private int hashCode;
+    private volatile int hashCode;
```
cstamas: The `hashCode()` is not concurrency safe here, so I'd make this `final` instead and modify `hashCode()` to just return the `final int hashCode` instead.

maven-resolver-util/src/main/java/org/eclipse/aether/util/repository/ChainedWorkspaceReader.java
```diff
@@ -38,7 +38,7 @@ public final class ChainedWorkspaceReader implements WorkspaceReader {
 
     private final List<WorkspaceReader> readers = new ArrayList<>();
 
-    private WorkspaceRepository repository;
+    private volatile WorkspaceRepository repository;
```
cstamas: Isn't this rather `final`?

cstamas:
Overall looks good, but there are IMHO some unnecessary usage of `volatile` (not biggie, but anyone seeing this keyword may frantically look for "where is this concurrently used?" and maybe find nothing), and similarly for `synchronized`. Latter even slows down things in case of true single threaded run, no? (ie obtaining object monitor...)

maven-resolver-util/src/main/java/org/eclipse/aether/util/repository/ChainedWorkspaceReader.java
cstamas: See https://github.com/apache/maven-resolver/pull/1909

cstamas:
@gnodet pls review #1909 and then rebase and possibly apply comments?

cstamas:
Pls rebase this @gnodet again?

maven-resolver-api/src/main/java/org/eclipse/aether/transfer/TransferResource.java
gnodet: `TransferResource` is created by one thread but its `contentLength` and `resumeOffset` are set later via setters and read from other threads (e.g. transfer listeners). Without `volatile`, writes from the producing thread may not be visible to listener threads due to the Java Memory Model — there is no happens-before relationship between the set and get calls across threads.

maven-resolver-connector-basic/src/main/java/org/eclipse/aether/connector/basic/BasicRepositoryConnectorFactory.java
gnodet: `BasicRepositoryConnectorFactory` is a singleton component whose `priority` can be set via `setPriority()` and read via `getPriority()` from different threads. Without `volatile`, a priority change made by one thread may never be seen by another thread reading it.

maven-resolver-generator-gnupg/src/main/java/org/eclipse/aether/generator/gnupg/GnupgSignatureArtifactGenerator.java
gnodet: `generate()` mutates `this.artifacts` (via `addAll`) and `this.signatureTempFiles`. It can be called multiple times during an upload (once per batch of generated artifacts). If two threads invoke it concurrently, the unsynchronized `ArrayList` operations would cause data races. The `synchronized` ensures mutual exclusion between `generate()` and `close()` which also accesses `signatureTempFiles`.

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/DataPool.java
gnodet: The original `intern()` had a classic check-then-act race: thread A reads `map.get(key)` and finds null/expired, thread B does the same, then both put a new entry — defeating the purpose of interning (callers get different instances for the same key). Wrapping the check-and-put in `synchronized` + `map.compute()` makes the operation atomic. The `synchronized` block is needed because `WeakHashMap` is not thread-safe even for single operations.

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/DependencyCollectorDelegate.java
gnodet: `Results` is shared across threads during parallel dependency collection. `errorPath` is written by `addException()` (now synchronized) but was also read directly by field access. The `volatile` ensures visibility of writes to threads that read `errorPath` outside of the synchronized methods.

maven-resolver-util/src/main/java/org/eclipse/aether/util/repository/ChainedWorkspaceReader.java
gnodet: You are right — this change has been dropped from this PR. Your PR #1909 handles it properly with `AtomicReference` and truly final `readers`.

gnodet: Thanks — the `ChainedWorkspaceReader` change has been dropped from this PR since #1909 (now merged) addresses it properly.

maven-resolver-generator-gnupg/src/main/java/org/eclipse/aether/generator/gnupg/GnupgSignatureArtifactGenerator.java
cstamas: While this generic assessment is right, artifact generator instances are _created_ in installer/deployer (on method entry, so if install/deploy happens on two or more thread -- each thread will have own generator), and after use generated artifacts are passed along with existing artifacts to connector, which again, will deploy in parallel. 

So, the "If two threads invoke it concurrently" observation is true in general, but I don't see where can it happen.

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/DataPool.java
cstamas: Agreed, thanks.

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/DependencyCollectorDelegate.java
cstamas: Agreed, thanks

maven-resolver-api/src/main/java/org/eclipse/aether/transfer/TransferResource.java
cstamas: Agreed, thanks

maven-resolver-connector-basic/src/main/java/org/eclipse/aether/connector/basic/BasicRepositoryConnectorFactory.java
cstamas: Agreed. Perso, I would scrutinize `PrioritizedComponents` that single user of this field across whole resolver.

cstamas (approved):

maven-resolver-generator-gnupg/src/main/java/org/eclipse/aether/generator/gnupg/GnupgSignatureArtifactGenerator.java
cstamas: Change is okay despite my objection. Thanks

cstamas:
@gnodet pls sort out conflicts and lets merge

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/selector/AndDependencySelector.java
gnodet: Done — made the field `final` and computed the hashCode eagerly in each constructor. `hashCode()` now just returns the precomputed value.

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/selector/ExclusionDependencySelector.java
gnodet: Done — made the field `final` and computed the hashCode eagerly in each constructor. `hashCode()` now just returns the precomputed value.

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/traverser/AndDependencyTraverser.java
gnodet: Done — made the field `final` and computed the hashCode eagerly in each constructor. `hashCode()` now just returns the precomputed value.

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/version/ChainedVersionFilter.java
gnodet: Done — made the field `final` and computed the hashCode eagerly in each constructor. `hashCode()` now just returns the precomputed value.

PR #1904, opened 2026-06-07
Bugfix: Fix NPE in PathConflictResolver

### LINKED ISSUE #1903: PathConflictResolver bug: dom4j
### Affected version

master

### Bug description

The `dom4j:dom4j:1.6.1` root transitive resolution causes NPE in new PathConflictResolver.

But there is something funky happening, as even Maven 3.9 and 3.8 create _bad classpath_ for it: contains duplicate of `dom4j:dom4j` versions 1.6.1 -- at root -- and 1.5.2 -- at second level; and it ends up with classpath that has two distinct dom4j versions in it.

To simulate "resolve as root" with` m-dependency-p` (due lack of mojo to point it at remote GAV), I downloaded dom4j [POM](https://repo.maven.apache.org/maven2/dom4j/dom4j/1.6.1/dom4j-1.6.1.pom) locally, renamed it to `pom.xml` and run `dependency:tree` goal: https://gist.github.com/cstamas/59a7eaf571004f75b6179d667233f72e

Reproducer: maven-resolver-demo-snippets class `ResolveTransitiveDependencies` should be modified to resolve transitively `dom4j:dom4j:1.6.1`, and it results in NPE.

This is **truly an edge case** and affects only two things as I am aware currently: this demo above, and Quarkus build Mojo. The thing is that nothing else (aside of these two things) tries to resolve `dom4j` as _root_ (it would happen if you in fact just are building it -- which again, is impossible to happen, given it was Maven 1 project!). So, this may point at Quarkus build Mojo issue as well, like "why is that mojo -- same as altered demo -- trying to resolve a _dependency_ as a _root_?". 

For ref, see:
https://gist.github.com/cstamas/e24577162083f1e100d78a3e5d398cd3

Here, `gav-tree` does same thing (using Maven 3.9.16 + Resolver 1, so classic conflict resolver is used) as Quarkus build Mojo, resolves dom4j as "root". And it does it wrong (after conflict resolution, duplicate is left in: `dom4j:dom4j` is present with two versions!). But, the `gav-tree-dd` (show tree of given dependency, as if it is used as direct dependency; and this happens when you enlist `dom4j:dom4j:1.6.1` in your POM as a dependency), the "real outcome", and is correct.

The (notable) difference between `tree` and `tree-dd` is that latter creates a "fake root", and adds subject as first level (direct) dependency. The dom4j contains some optional dependencies (causing cycles), that in first mode are NOT excluded and in latter are (collection _is contextual_, the "depth" of node matters). In short, in first case, optionals are not ignored (like optional deps in your POM; are still resolved and added to classpath), while in latter they are ignored.

While this _may be a bug of Quarkus build mojo_, the NPE is something that should not happen in resolver.

cstamas:
Fixes #1903 bug in `PathConflictResolver` (PCR).

Also, the fix in 5e97fd15f8dbc562c481252bc8ea1037992e6407 is wrong, and this PR redoes now all.

Problem was cycle detection. Original (2.0.18) PCR did left "uncovered" nodes, still causing cycles, that should have been removed by PCR (causing SO in client code). The commit above remedied it, but caused another issue how cycles were detected (it opted to reuse existing `CYCLIC_CONFLICT_IDS`), but was again wrong.

Then I realized: PCR by its design is "rebuilding" tree with `conflictIds` on Path instances, so it can easily spot properly when a cycle is about to happen (when a next node is partitioned into same `CONFLICT_ID` that is already contained in the path to root.

Proof: the demo code (added comment) is now working as expected, plus, Quarkus build is buildable with Maven 3.10-SNAPSHOT using Resolver with this patch.

As an interesting side-effect, this PR also fixes an issue applicable to _all released maven versions so far_:
https://gist.github.com/cstamas/e24577162083f1e100d78a3e5d398cd3

As the demo now produces "proper" tree as:
```
dom4j:dom4j:jar:1.6.1 [compile]
+- jaxme:jaxme-api:jar:0.3 [compile, optional]
+- jaxen:jaxen:jar:1.1-beta-6 [compile, optional]
|  +- jdom:jdom:jar:1.0 [compile, optional]
|  +- xerces:xmlParserAPIs:jar:2.6.2 [compile, optional]
|  +- xerces:xercesImpl:jar:2.6.2 [compile, optional]
|  \- xom:xom:jar:1.0b3 [compile, optional]
|     +- com.ibm.icu:icu4j:jar:2.6.1 [compile, optional]
|     +- xalan:xalan:jar:2.6.0 [compile, optional]
|     \- org.ccil.cowan.tagsoup:tagsoup:jar:0.9.7 [compile, optional]
+- msv:xsdlib:jar:20030807 [compile, optional]
+- msv:relaxngDatatype:jar:20030807 [compile, optional]
+- pull-parser:pull-parser:jar:2 [compile, optional]
+- xpp3:xpp3:jar:1.1.3.3 [compile, optional]
+- stax:stax-api:jar:1.0 [compile, optional]
\- xml-apis:xml-apis:jar:1.0.b2 [compile]
```

The duplicate `dom4j:dom4j:jar:1.5.2` is gone from the tree (but is present in Maven 3.9, Maven 3.8 and all other versions).

### DIFF
--- a/maven-resolver-demos/maven-resolver-demo-snippets/src/main/java/org/apache/maven/resolver/examples/ResolveTransitiveDependencies.java
+++ b/maven-resolver-demos/maven-resolver-demo-snippets/src/main/java/org/apache/maven/resolver/examples/ResolveTransitiveDependencies.java
@@ -51,7 +51,7 @@ public class ResolveTransitiveDependencies {
                 CloseableSession session = Booter.newRepositorySystemSession(system, Booter.selectFs(args))
                         .build()) {
             // An "interesting" artifact that depends on itself (different version). All Maven versions
-            // released so far contained dom4j:dom4j:1.5.2 on classpath for dom4j:dom4j:1.5.2.
+            // released so far contained dom4j:dom4j:1.5.2 on classpath for dom4j:dom4j:1.6.1
             // In Resolver 2.0.19 this is fixed.
             Artifact artifact = new DefaultArtifact("dom4j:dom4j:1.6.1");
 
--- a/maven-resolver-demos/maven-resolver-demo-snippets/src/main/java/org/apache/maven/resolver/examples/ResolveTransitiveDependencies.java
+++ b/maven-resolver-demos/maven-resolver-demo-snippets/src/main/java/org/apache/maven/resolver/examples/ResolveTransitiveDependencies.java
@@ -50,6 +50,7 @@ public class ResolveTransitiveDependencies {
         try (RepositorySystem system = Booter.newRepositorySystem(Booter.selectFactory(args));
                 CloseableSession session = Booter.newRepositorySystemSession(system, Booter.selectFs(args))
                         .build()) {
+            // Artifact artifact = new DefaultArtifact("dom4j:dom4j:1.6.1");
             Artifact artifact = new DefaultArtifact("org.apache.maven.resolver:maven-resolver-impl:1.3.3");
 
             DependencyFilter classpathFilter = DependencyFilterUtils.classpathFilter(JavaScopes.COMPILE);
--- a/maven-resolver-demos/maven-resolver-demo-snippets/src/main/java/org/apache/maven/resolver/examples/ResolveTransitiveDependencies.java
+++ b/maven-resolver-demos/maven-resolver-demo-snippets/src/main/java/org/apache/maven/resolver/examples/ResolveTransitiveDependencies.java
@@ -52,7 +52,7 @@ public class ResolveTransitiveDependencies {
                         .build()) {
             // An "interesting" artifact that depends on itself (different version). All Maven versions
             // released so far contained dom4j:dom4j:1.5.2 on classpath for dom4j:dom4j:1.6.1
-            // In Resolver 2.0.19 this is fixed.
+            // In Resolver 2.0.19+ this is fixed.
             Artifact artifact = new DefaultArtifact("dom4j:dom4j:1.6.1");
 
             DependencyFilter classpathFilter = DependencyFilterUtils.classpathFilter(JavaScopes.COMPILE);
--- a/maven-resolver-demos/maven-resolver-demo-snippets/src/main/java/org/apache/maven/resolver/examples/ResolveTransitiveDependencies.java
+++ b/maven-resolver-demos/maven-resolver-demo-snippets/src/main/java/org/apache/maven/resolver/examples/ResolveTransitiveDependencies.java
@@ -50,8 +50,10 @@ public class ResolveTransitiveDependencies {
         try (RepositorySystem system = Booter.newRepositorySystem(Booter.selectFactory(args));
                 CloseableSession session = Booter.newRepositorySystemSession(system, Booter.selectFs(args))
                         .build()) {
-            // Artifact artifact = new DefaultArtifact("dom4j:dom4j:1.6.1");
-            Artifact artifact = new DefaultArtifact("org.apache.maven.resolver:maven-resolver-impl:1.3.3");
+            // An "interesting" artifact that depends on itself (different version). All Maven versions
+            // released so far contained dom4j:dom4j:1.5.2 on classpath for dom4j:dom4j:1.5.2.
+            // In Resolver 2.0.19 this is fixed.
+            Artifact artifact = new DefaultArtifact("dom4j:dom4j:1.6.1");
 
             DependencyFilter classpathFilter = DependencyFilterUtils.classpathFilter(JavaScopes.COMPILE);
 
--- a/maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/transformer/PathConflictResolver.java
+++ b/maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/transformer/PathConflictResolver.java
@@ -614,9 +614,9 @@ public final class PathConflictResolver extends ConflictResolver {
         /**
          * Adds node children: this method should be "batch" used, as all (potential) children should be added at once.
          * Method will return really added ones, as this class avoids cycles. Those forming a cycle are added to {@link Path}
-         * structure but are bot recursed, keeping {@link Path} cycle free.
+         * structure but are not recursed (returned in list), keeping {@link Path} cycle free.
          * This method also maintains "conflictIdPath", that is a list of "conflict partition IDs" leading from root
-         * toward given instance. This implies that this conflict resolver, by its nature "recalculates" the
+         * toward given instance. This implies that this conflict resolver, by its nature "redoes" the
          * {@link TransformationContextKeys#CYCLIC_CONFLICT_IDS} calculated by {@link ConflictIdSorter}.
          */
         private List<Path> addChildren(List<DependencyNode> children) throws RepositoryException {
--- a/maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/transformer/PathConflictResolver.java
+++ b/maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/transformer/PathConflictResolver.java
@@ -624,10 +624,11 @@ public final class PathConflictResolver extends ConflictResolver {
 
         /**
          * Adds node children: this method should be "batch" used, as all (potential) children should be added at once.
-         * Method will return really added ones, as this class avoids cycles. Those forming a cycle are added to {@link Path}
-         * structure but are not recursed (returned in list), keeping {@link Path} cycle free.
-         * This method also maintains "conflictIdPath", that is a list of "conflict partition IDs" leading from root
-         * toward given instance. This implies that this conflict resolver, by its nature "redoes" the
+         * Method will return really added {@link Path} instances, as this class avoids cycles. Those forming a cycle
+         * are added to {@link Path} structure but are not recursed (not returned in list), keeping {@link Path} cycle
+         * free. This method also maintains "conflictIdsSinceRoot", that is a set of "conflict IDs" that were touched
+         * while going from tree root to current node instance.
+         * This implies that this conflict resolver, by its nature "redoes" the
          * {@link TransformationContextKeys#CYCLIC_CONFLICT_IDS} calculated by {@link ConflictIdSorter}.
          */
         private List<Path> addChildren(List<DependencyNode> children) throws RepositoryException {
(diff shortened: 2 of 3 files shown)


maven-resolver-demos/maven-resolver-demo-snippets/src/main/java/org/apache/maven/resolver/examples/ResolveTransitiveDependencies.java
```diff
@@ -50,6 +50,7 @@ public static void main(String[] args) throws Exception {
         try (RepositorySystem system = Booter.newRepositorySystem(Booter.selectFactory(args));
                 CloseableSession session = Booter.newRepositorySystemSession(system, Booter.selectFs(args))
                         .build()) {
+            // Artifact artifact = new DefaultArtifact("dom4j:dom4j:1.6.1");
```
slawekjaranowski: do we need such comment?

slawekjaranowski (approved):

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/transformer/PathConflictResolver.java
```diff
@@ -343,7 +319,8 @@ private State(
          * sorted conflictIds can serve as a starting to point to walk the graph.
          */
         private Path build(DependencyNode node) throws RepositoryException {
-            Path root = new Path(this, node, null, false);
+            String nodeConflictId = this.conflictIds.get(node);
+            Path root = new Path(this, node, nodeConflictId, Collections.singletonList(nodeConflictId), null);
```
gnodet: Worth noting: for a virtual root node (one with `getDependency() == null`), `conflictIds.get(node)` returns `null` because `ConflictMarker` only puts nodes with non-null dependencies into the map. This means `conflictIdPath` starts as `[null]`.

This is safe in practice — `newPath.contains(childConflictId)` in `addChildren()` won't match any real (non-null) child conflict ID against `null`, so no false cycle detection. And adding `null` to `partitions` is the same behavior as the old constructor (`this.conflictId = state.conflictIds.get(dn)`). So no regression.

Still, if you wanted to be defensive, you could guard against it:

```suggestion
            String nodeConflictId = this.conflictIds.get(node);
            Path root = new Path(this, node, nodeConflictId,
                    nodeConflictId != null ? Collections.singletonList(nodeConflictId) : Collections.emptyList(),
                    null);
```

This avoids polluting `conflictIdPath` with `null` for virtual roots. Non-blocking — the current code works correctly either way.

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/transformer/PathConflictResolver.java
```diff
@@ -633,19 +613,23 @@ private void moveOutOfScope() {
...
             ArrayList<Path> added = new ArrayList<>(children.size());
             for (DependencyNode child : children) {
-                boolean cycle = this.state
-                        .cyclicPredecessors
-                        .getOrDefault(this.state.conflictIds.get(child), Collections.emptySet())
-                        .contains(this.conflictId);
-                Path c = new Path(this.state, child, this, cycle);
+                String childConflictId = this.state.conflictIds.get(child);
+                List<String> newPath = new ArrayList<>(this.conflictIdPath);
+                boolean cycle = newPath.contains(childConflictId);
```
gnodet: This is the key improvement over 5e97fd15. `contains()` on an `ArrayList<String>` is O(depth) per child, making the total cost O(N·D) where N is node count and D is average depth. For typical Maven dependency trees (depth < 20), this is negligible.

If you ever wanted to optimize for pathological deep trees, you could pass a `Set<String>` alongside the list (or instead of it, since order doesn't matter for cycle detection — only membership). But that's premature optimization for Maven's use case.

One thought: since `newPath` is a new `ArrayList` copy per child, it becomes the child's `conflictIdPath` and is never modified after construction. You could wrap it with `Collections.unmodifiableList(newPath)` in the `Path` constructor to enforce immutability, but again — not blocking, just a style preference.

maven-resolver-demos/maven-resolver-demo-snippets/src/main/java/org/apache/maven/resolver/examples/ResolveTransitiveDependencies.java
```diff
@@ -50,7 +50,10 @@ public static void main(String[] args) throws Exception {
         try (RepositorySystem system = Booter.newRepositorySystem(Booter.selectFactory(args));
                 CloseableSession session = Booter.newRepositorySystemSession(system, Booter.selectFs(args))
                         .build()) {
-            Artifact artifact = new DefaultArtifact("org.apache.maven.resolver:maven-resolver-impl:1.3.3");
+            // An "interesting" artifact that depends on itself (different version). All Maven versions
+            // released so far contained dom4j:dom4j:1.5.2 on classpath for dom4j:dom4j:1.6.1
+            // In Resolver 2.0.19 this is fixed.
```
gnodet: Minor: the comment says "In Resolver 2.0.19 this is fixed" — should this reference the version that will actually include this fix? If 2.0.19 hasn't been released yet, it might be worth saying "In Resolver 2.0.19+" or just leaving it as a general statement. Also, is this the right version number, or should it track whatever the next release cut will be?

gnodet (commented):
Nice fix, Tamas. I reviewed the full history of cycle detection in PCR — from the original `partitions.stream().anyMatch()` approach, through the `cyclicPredecessors` rework in 5e97fd15, to this PR — and the path-based approach is clearly the right design.

## Root cause analysis

The bug in 5e97fd15 was that `cyclicPredecessors` is a **global** property of the conflict-ID graph: it tells you which conflict IDs participate in *some* cycle *somewhere*, but not whether a specific path from root actually traverses a cycle. The old `addChildren()` condition `if (c.parent != null && !c.parent.cycle)` then propagated this to exclude **all** children of any cycle-flagged node from recursion — even children that aren't part of any cycle on that path.

For `dom4j:dom4j:1.6.1`, the cycle is `dom4j_id → jaxen_id → dom4j_id`. With the old approach, jaxen gets `cycle=true` because `cyclicPredecessors.getOrDefault(jaxen_id, ∅).contains(dom4j_id)` is true. Then ALL of jaxen's children (xom, jdom, xerces, etc.) are excluded from exploration because `!c.parent.cycle` is false. Nodes reachable only through those subtrees never get `Path` entries in `partitions`, but their conflict IDs *are* in `sortedConflictIds` (computed by `ConflictIdSorter` on the full graph). When the main loop hits `state.partitions.get(conflictId)` for one of those orphaned IDs, it returns `null`, and `null.isEmpty()` → NPE.

## Why the new approach is correct

Tracking `conflictIdPath` per-`Path` instance gives **path-local** cycle detection: a child is flagged as a cycle only when its own conflictId already appears on the *specific path from root to the current node*. This means:
- `dom4j(1.6.1) → jaxen`: path is `[dom4j_id]`, jaxen_id ∉ path → **not a cycle** ✓
- `dom4j(1.6.1) → jaxen → dom4j(1.5.2)`: path is `[dom4j_id, jaxen_id]`, dom4j_id ∈ path → **cycle** ✓

This is strictly more precise than the global approach and produces correct results for the dom4j scenario.

## Removal of `CYCLIC_CONFLICT_IDS` dependency

Safe. `ClassicConflictResolver` still consumes it, and `ConflictIdSorter` still computes it. PCR now being self-sufficient for cycle detection is actually a simplification of the contract.

## Minor observations

See inline comments for details — all non-blocking.

Overall, the algorithm change is sound, the code is clean, and the existing test suite (`winnerCycleRemoved`, `subtreeRepeatedSameInstanceWithRealCycle`, `nettyBoringSslExample`) should exercise the new cycle detection path. The dom4j demo is a good real-world proof. I'd suggest one small hardening and a unit test addition (see inline), but nothing blocking.

— gnodet

gnodet:
One additional thought on **test coverage**: the existing `ConflictResolverTest` covers several cycle scenarios (`winnerCycleRemoved`, `subtreeRepeatedSameInstanceWithRealCycle`, `nettyBoringSslExample`), but none of them reproduce the specific topology that triggered the NPE — where a cycle-flagged node has children that are *not* in the cycle, and those children's subtrees contain unique conflictIds not reachable by any other path.

A test like the following would nail down the exact regression from 5e97fd15 and prevent it from recurring:

```java
// dom4j-like topology:
// root -> A -> B (in cycle with A via cyclicPredecessors) -> C (unique, only reachable through B)
//                                                         -> A' (same conflictId as A, actual cycle)
// Old code: B flagged as cycle → C never explored → partitions.get(C_id) == null → NPE
// New code: B not a cycle (B_id ∉ [root_id, A_id]) → C explored normally → cycle broken at A'
```

Not blocking the merge, but would be a nice safety net. The existing tests exercise the *correction* path (cycle properly broken) but not the *non-regression* path (non-cycle siblings properly explored).

— gnodet

cstamas:
Applied all PR comments in last commit, using set, using null-protection in set, added test, fixed comment (added + to have some ref to release timeline).

gnodet (approved):
Re-reviewed after the two follow-up commits (`19f9f5cf` "Applied PR comments" and `6af86706` "Fix javadoc too").

**Review feedback properly addressed:**

1. **Null root conflictId** (@gnodet) — handled with the ternary at `build()`: `nodeConflictId != null ? Collections.singleton(nodeConflictId) : Collections.emptySet()`. The `Collections.singleton()`/`emptySet()` is safe because `addChildren()` always copies into a new `HashSet` before mutating.

2. **Version comment** (@gnodet, @slawekjaranowski) — updated to "In Resolver 2.0.19+ this is fixed."

3. **Javadoc** — updated in `addChildren()` to reference `conflictIdsSinceRoot` (set) instead of `conflictIdPath` (list), accurately describing the current implementation.

**Correctness of the fix:**

- **Set propagation is correct.** Each child gets its own `new HashSet<>(this.conflictIdsSinceRoot)` copy, so sibling paths are isolated — one sibling's cycle detection can't pollute another's.
- **Cycle detection via `Set.add()` is sound.** Returns `false` exactly when `childConflictId` is already on the root-to-here path, which is the correct definition of a path-level cycle.
- **Performance is good.** `HashSet.add()` gives O(1) cycle detection per child vs. O(depth) for the old `ArrayList.contains()`. Memory allocation is comparable (same number of collection copies, slightly more per-element overhead for HashSet, but negligible in practice).
- **Edge cases hold.** Empty/single-node graphs terminate immediately in `gatherCRNodes()`. Deeply nested trees create a chain of HashSets — same asymptotic memory as the old ArrayList approach.

**Test `issue1903`:**

The test faithfully reproduces the dom4j scenario (root → a:1.0 → b:1.0 → [c:1.0, a:1.1 → b:1.0]). The assertion that `a` appears exactly once in the transformed tree confirms the cycle is pruned without incorrectly removing the subtree rooted at `b`. Test passes for both ConflictResolver variants.

LGTM — the path-based `conflictIdsSinceRoot` approach is a clean and correct replacement for the global `cyclicPredecessors`.

_Review by Claude Code (automated)_

PR #1910, opened 2026-06-08
Bug: fix IPC named lock client-side robustness for Windows

gnodet:
## Summary

Fixes Windows CI flakiness in `IpcNamedLockFactoryIT.exclusiveAccess()` (seen in PR #1909, CI run 27139141318).

**Root cause**: When the loser thread's `client.lock()` times out in `doLockExclusively()`, the cleanup call `client.unlock(contextId)` could either:
- **Hang forever** — it used `Long.MAX_VALUE` timeout, so if the server/receiver had any issue, the thread blocked indefinitely
- **Throw RuntimeException** — which propagated out of `Access.run()` (which only catches `InterruptedException`), causing the thread to terminate without calling `loser.countDown()`. The winner thread then waited on `loser.await()` forever, and the test hung until Surefire killed it at 25 seconds.

## Changes

1. **`IpcNamedLock`**: Wrap cleanup unlock in `tryUnlock()` — a try-catch that swallows exceptions so the calling thread always returns `false` instead of crashing. This is the primary fix.

2. **`IpcClient.unlock()`**: Replace `Long.MAX_VALUE` timeout with 10 seconds. Normal unlock round-trips complete in microseconds; 10 seconds is generous enough while preventing infinite hangs.

3. **`IpcClient.stopServer()`**: Replace `Long.MAX_VALUE` timeout with 30 seconds.

4. **`IpcClient.receive()`**: On empty response, complete the future exceptionally instead of throwing `IllegalStateException` — the old behavior killed the receiver thread, breaking all subsequent operations on the shared client.

5. **`IpcClient.close()`**: Call `close()` on EOF so pending futures are properly cleaned up when the server disconnects, and reset the `initialized` flag to prevent stale state.

## Test plan

- [x] All 27 IPC tests pass locally (unit + integration: `IpcNamedLockFactoryIT`, `IpcAdapterNoForkIT`, `IpcAdapterIT`)
- [ ] Windows CI should no longer hang on `exclusiveAccess`

### DIFF
--- a/maven-resolver-named-locks-ipc/src/main/java/org/eclipse/aether/named/ipc/IpcClient.java
+++ b/maven-resolver-named-locks-ipc/src/main/java/org/eclipse/aether/named/ipc/IpcClient.java
@@ -299,12 +299,13 @@ public class IpcClient {
                     continue;
                 }
                 if (s.isEmpty()) {
-                    throw new IllegalStateException("Protocol error");
+                    f.completeExceptionally(new IOException("Protocol error: empty response"));
+                    continue;
                 }
                 f.complete(s);
             }
         } catch (EOFException e) {
-            // server is stopped; just quit
+            close(new IOException("Server disconnected", e));
         } catch (Exception e) {
             close(e);
         }
@@ -326,6 +327,7 @@ public class IpcClient {
         try {
             return response.get(time, unit);
         } catch (InterruptedException e) {
+            responses.remove(id);
             throw (IOException) new InterruptedIOException("Interrupted").initCause(e);
         } catch (ExecutionException e) {
             throw new IOException("Execution error", e);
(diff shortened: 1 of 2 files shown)


gnodet:
retest

gnodet:
Pushed a fix for the remaining NPE race condition in `IpcClient.send()`.

**Root cause:** The `send()` method read the `output` field multiple times without holding the same lock as `close(Throwable)`, which nulls the field under `synchronized(this)`. Between `ensureInitialized()` returning and reaching `synchronized(output)`, the receiver thread could call `close(Throwable)` — setting `output = null` — causing a `NullPointerException` at `output.flush()`.

**Fix:** Capture `output` in a local variable immediately after `ensureInitialized()`, check it for null (throwing `IOException("Connection closed")` if the connection was already torn down), and use the local variable for both the `synchronized` block and all writes within it. This eliminates the TOCTOU race between the field read and its use.

gnodet:
Rerunning CI to gather additional Windows stability data.

gnodet:
retest this please

cstamas (approved):

Issue #1920, opened 2026-06-13
WarnChecksumPolicy.onNoMoreChecksums() throws like fail policy on missing checksums


ascheman:
## Behavior

When a remote repository transfer yields no checksum data (the artifact is
fetched but neither `.sha1` nor `.md5` companion exists for the resource),
`AbstractChecksumPolicy.onNoMoreChecksums()` raises a
`ChecksumFailureException` unconditionally:

```java
@Override
public void onNoMoreChecksums() throws ChecksumFailureException {
    throw new ChecksumFailureException("Checksum validation failed, no checksums available");
}
```

`WarnChecksumPolicy` and `FailChecksumPolicy` inherit this method without
overriding it, so `warn` raises the same hard exception as `fail` in this
code path. The current behavior is pinned by
`WarnChecksumPolicyTest.testOnNoMoreChecksums`, so it is intentional /
documented state — but the consequence for users is that only
`<checksumPolicy>ignore</checksumPolicy>` effectively lets transfers proceed
when checksum data is absent, regardless of how the policy is named.

## Where this matters in practice

The `file://` test-fixture pattern hits this every time. Apache plugins have
been working around it by setting `<checksumPolicy>ignore</checksumPolicy>`
in `src/it[/mrm]/settings.xml`. Snapshot from
`gh api search/code` 2026-06-10:

**Added during Maven 4 API migration (June 2024)**

- apache/maven-clean-plugin — [`2da14745`](https://github.com/apache/maven-clean-plugin/commit/2da147458800dfe9ecc24865aef0e718ac56df57) (apache/maven-clean-plugin#20)
- apache/maven-compiler-plugin — [`9dc957dc`](https://github.com/apache/maven-compiler-plugin/commit/9dc957dc55f8a0baeef423084a95769f92951047)
- apache/maven-deploy-plugin — [`88ce1593`](https://github.com/apache/maven-deploy-plugin/commit/88ce1593800015ac41e61f3c78d9b4d0ecb97460)
- apache/maven-project-info-reports-plugin — [`a44cc8ab`](https://github.com/apache/maven-project-info-reports-plugin/commit/a44cc8ab44dda25c08d658666e0b2dadf826f93d) (apache/maven-project-info-reports-plugin#70)
[...]

Issue #1782, opened 2026-01-31
Better checksum control


cstamas:
### New feature, improvement proposal

Right now, we have one single [configuration key](https://maven.apache.org/resolver/configuration.html) that controls both, generation of theirs on deploy, and retrieval on consumption. Make it possible to:
* have one key as today
* but, if user wants, split checksums generated on deploy (ie. user may want sha512 deployed) and checksums looked for on consumption.

PR #1922, opened 2026-06-13
WarnChecksumPolicy: proceed on no checksums

ascheman:
Draft / one shape for discussion — fixes the minimal manifestation of
apache/maven-resolver#1920 while leaving the larger design question open
for `dev@maven.apache.org`.

## Change

Override `AbstractChecksumPolicy.onNoMoreChecksums()` in
`WarnChecksumPolicy` so it logs a warning and returns, instead of
inheriting the unconditional `ChecksumFailureException` throw.

`FailChecksumPolicy` is unchanged: it continues to throw via the
inherited base behavior.

`WarnChecksumPolicyTest.testOnNoMoreChecksums()` updated to assert the
new behavior (no exception).

## Scope and framing

This is the smallest possible patch that addresses #1920's reported
behavior. It does **not** touch:

- the default behavior for repos without an explicit `<checksumPolicy>`
- `file://` vs HTTP transport semantics
- the broader checksum-policy-controls space being redesigned in
  #1782 / #1784 / #1917

Other shapes (transport-aware default in `DefaultChecksumPolicyProvider`,
new opt-in property like `aether.checksums.noChecksumsOutcome` analogous
to RRF's `noInputOutcome`, Maven 4 compatibility-mode flag) are
discussed separately on `dev@maven.apache.org`. **Marking as draft so
the maintainer can decide whether this minimal shape is the right one,
or whether a broader change is preferred.**

## Verification

- New `WarnChecksumPolicyTest.testOnNoMoreChecksums()` passes locally
  (JDK 21, Maven 3.9.16, `2.0.19-SNAPSHOT`).
- Full `maven-resolver-impl` test suite passes locally.

cc @cstamas

### DIFF
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/WarnChecksumPolicy.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/WarnChecksumPolicy.java
@@ -39,4 +39,12 @@ final class WarnChecksumPolicy extends AbstractChecksumPolicy {
                 exception);
         return true;
     }
+
+    @Override
+    public void onNoMoreChecksums() {
+        logger.warn(
+                "Could not validate integrity of download from {}{}: no checksums available",
+                resource.getRepositoryUrl(),
+                resource.getResourceName());
+    }
 }
--- a/maven-resolver-impl/src/test/java/org/eclipse/aether/internal/impl/WarnChecksumPolicyTest.java
+++ b/maven-resolver-impl/src/test/java/org/eclipse/aether/internal/impl/WarnChecksumPolicyTest.java
@@ -79,11 +79,6 @@ public class WarnChecksumPolicyTest {
 
     @Test
     void testOnNoMoreChecksums() {
-        try {
-            policy.onNoMoreChecksums();
-            fail("No exception");
-        } catch (ChecksumFailureException e) {
-            assertTrue(e.getMessage().contains("no checksums available"));
-        }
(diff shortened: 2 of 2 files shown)


cstamas:
@ascheman Sorry, was faster :smile:  please take a peek at other PR and let's reconcile both PRs, as I perso like better your message!

cstamas (approved):

ascheman:
> @ascheman Sorry, was faster 😄 please take a peek at other PR and let's reconcile both PRs, as I perso like better your message!

Thanks!

I intentionally started this as draft as I think there are other opportunities (larger ones, check the mail I just sent to dev Mailinglist). Nevertheless, this could be a good first approximation of a solution which we might merge until something better comes up?

cstamas:
This is still draft, on purpose?

ascheman:
> This is still draft, on purpose?

Yes, did you see my mail on the dev list? I see the two opportunities:
- Just merge it and have a quick fix
- Consider a longer solution in the course of your anticipated redesign of resolver?

For the sake of simplicity (and speed), I would suggest, to merge it now and think about a more general solution later.

Having said that, I will now make it ready and leave it to you to merge now or discuss other solutions first.
// @hboutemy who also asked already about the more elaborated solutions proposed on the mailing list (which seem to make it unnecessary complex in this very moment).

cstamas:
I will merge this as "least painful" thing, and later can discuss, want 2.0.19 out ASAP.

PR #1923, opened 2026-06-13
Fix ApacheTransport auth cache CCEx

### LINKED ISSUE #1919: HttpTransporter shares AuthCache in the session cache under a classloader-agnostic key -> ClassCastException across ClassRealms
### Affected version

`1.9.26` / `1.9.27` (`maven-resolver-transport-http`). The same code is present on `master`/2.x in `maven-resolver-transport-apache` (`ApacheTransporter`), so it appears to affect 2.x as well.

### Bug description

**Summary**

Since 1.9.26, `HttpTransporter` stores its Apache HttpClient `AuthCache` in the **session-level `RepositoryCache`** under a key derived from `getClass().getSimpleName()`. That key is **classloader-agnostic**, so when the transport classes are loaded in more than one `ClassRealm` in the same build, both realms compute the **same** cache key while each has its **own** `org.apache.http.*` classes — which happens when an `extensions=true` plugin pulls its own (typically different) version of HttpClient into its realm than the one Maven core provides. `AuthCache`/`BasicAuthCache` then resolve to different `Class` instances per realm, and the read-back cast fails:

```
java.lang.ClassCastException: class org.apache.http.impl.client.BasicAuthCache
cannot be cast to class org.apache.http.client.AuthCache
(... BasicAuthCache is in unnamed module of loader ClassRealm @<a>;
     AuthCache    is in unnamed module of loader ClassRealm @<b>)
    at org.eclipse.aether.transport.http.HttpTransporter.<init>(HttpTransporter.java:372)
    at org.eclipse.aether.transport.http.HttpTransporterFactory.newInstance(HttpTransporterFactory.java:95)
    ...
    at org.eclipse.aether.internal.impl.DefaultMetadataResolver$ResolveTask.run(DefaultMetadataResolver.java:555)
```

**Location & cause**

`maven-resolver-transport-http/.../HttpTransporter.java` (1.9.27), lines 368–378:

```java
if (session.getCache() != null) {
    String authCacheKey = getClass().getSimpleName() + "-" + repository.getId() + "-"
            + StringDigestUtil.sha1(repository.toString());          // classloader-agnostic key
    synchronized (AUTH_CACHE_MUTEX) {
        AuthCache cache = (AuthCache) session.getCache().get(session, authCacheKey);  // line 372 — throws
        if (cache == null) {
            cache = new BasicAuthCache();
            session.getCache().put(session, authCacheKey, cache);    // stored into the per-session shared cache
        }
        this.authCache = cache;
    }
}
```

`session.getCache()` is a single per-session `RepositoryCache` visible to every `ClassRealm`. Because the key only contains `"HttpTransporter"` + repo id + repo hash (no classloader identity), a transporter loaded by realm A and one loaded by realm B share the same entry — but their `AuthCache`/`BasicAuthCache` are different `Class` objects, so the cast at line 372 throws. `AUTH_CACHE_MUTEX` only serializes access; it does not isolate by realm.

This was introduced by #1790 ("Drastically simplify auth caching", GH-1768), which moved the `AuthCache` from per-transporter state into the shared session cache. It is the same family as #1296 / MRESOLVER-628 ("session prioritized-component caching → cross-realm `ClassCastException`", fixed in 2.0.4) — but that fix covered component factories, not this `AuthCache` instance. #1902 edits this block for thread-safety but keeps the cast, so it does not address this.

**When it occurs**

It is a probabilistic, classloader-topology-dependent failure. It requires:
- the transport loaded in **two realms** — classically an `extensions=true` plugin that bundles its own `maven-resolver-transport-http` + `httpclient` alongside the copies Maven core provides, and
- cold resolution that actually builds an HTTP transporter (`DefaultMetadataResolver` thread pool fetching remote metadata).

It typically presents as an **intermittent** "passes on rerun / warm cache" CI failure.

**Reproduction**

A Maven 3.9.x build using `spring-cloud-contract-maven-plugin` >= 5.0.3 (declared `<extensions>true</extensions>` per its docs; it bundles `maven-resolver-transport-http` 1.9.27 + `httpclient` 4.5.14 at compile scope) on a **cold** local repository with a non-trivial dependency graph:

```
mvn -pl <module-using-the-plugin> validate    # cold ~/.m2 -> BasicAuthCache cannot be cast to AuthCache
```

5.0.2 (which bundles resolver 1.9.25) does not reproduce, because 1.9.25's `HttpTransporter` does not use the session cache for the auth cache.

**Proposed fix**

Make the cache key classloader-aware so each realm keeps its own entry (preserving the intended per-session reuse within a realm):

```diff
-            String authCacheKey = getClass().getSimpleName() + "-" + repository.getId() + "-"
-                    + StringDigestUtil.sha1(repository.toString());
+            String authCacheKey = getClass().getName() + "@"
+                    + Integer.toHexString(System.identityHashCode(getClass().getClassLoader())) + "-"
+                    + repository.getId() + "-"
+                    + StringDigestUtil.sha1(repository.toString());
```

An equivalent alternative is to keep auth-cache state per-transporter (as in <=1.9.25) rather than in the session cache. The same change applies to `master`/2.x `ApacheTransporter`.

cstamas:
Started as proposed, to make session-wide auth cache string key classloader aware. Then realized, we have "classloader aware" class, that we make into string, with tricks, to make it "classloader aware", and this sounds awkward. Hence, `Keys` helper class.

This PR also revealed a bug, hidden deep, where (probably a copy pasted) code snippet used wrong key, see `RemoteRepositoryFilterSourceSupport`.

---

Back story: originally Resolver used `String` as keys that as we know, are perfect candidates for keys. But, if we stick to Maven, same component may end up loaded twice in Maven, and if key value contains something that is component specific (like the issue mentioned Apache transport value is Apache transport specific class), this will cause inevitably `ClassCastEx` if one instance puts something into map, and other one from the other classloader discovers it (and tries to cast it). In Maven, the resolver api, spi, util, impl and basic connector modules cannot be replaced (is prevented by Maven; circumvention is relocation or uber jar, changing the GAV), but other resolver modules can be loaded up multiple times, see https://github.com/apache/maven/blob/4106910f94e92a0967c61c62b4dbbcd6294c97a0/maven-core/src/main/resources/META-INF/maven/extension.xml#L185-L189

Hence the `Keys`, where user still has to decide how "scoped" the key should be, if using `String`s as key element, it will be global, and that is fine for keys mapped to values that are usable across class loaders as well (ie. value is consisted of `java.base` classes only). If we want class loader awareness (like in the issue, for Apache transport we do), then all user should do, is to use class loader specific class as one key element, and it will distinguish it from duplicate class loaded by another class loader. Finally, as update check manager does, "instance bound" keys are still possible too.

This change, while may look "radical" or even breaking, is in fact limited and does not cause API breakage in public surface of modules: api, spi and util.

Fixes #1919

### DIFF
--- a/maven-resolver-api/src/main/java/org/eclipse/aether/Keys.java
+++ b/maven-resolver-api/src/main/java/org/eclipse/aether/Keys.java
@@ -28,15 +28,15 @@ import static java.util.Objects.requireNonNull;
  * long as they are used by components loaded once in system. As we saw, some subsystems like transports can be
  * loaded multiple times. For example, in case of Maven, some transport may be present in Maven core, but also loaded up
  * by some extension. In this case, key should distinguish between their {@link ClassLoader}s.
- * <p/>
+ * <p>
  * It is this class caller responsibility to use proper objects as keys, this class merely helps one to create composite
- * keys our of object instances that are expected to be anyway good candidates as key. Examples of these objects are
+ * keys out of object instances that are expected to be anyway good candidates as key. Examples of these objects are
  * {@link String} instances but also {@link Class} instances, and many other types also usable as keys.
- * <p/>
+ * <p>
  * Important: never forget to perform {@code null}-check, when getting cache from {@link RepositorySystemSession#getCache()}
  * method as it may return {@code null}, when cache is disabled session-wise.
- * <p/>
- * Historical note: As mentioned above, use od {@link String} instances for keys is perfect match, and it worked from
+ * <p>
+ * Historical note: As mentioned above, use of {@link String} instances for keys is perfect match, and it worked from
  * very start. But, as use cases got more and more complex and keys used started to be constructed in more and more
  * sophisticated ways, but were still {@link String} instances. Believe, or not, the sole purpose of string keys
  * was easier debugging. By using this helper class, this convenience should remain.
@@ -53,11 +53,13 @@ public final class Keys {
     /**
      * Creates object instance usable as key in session data and cache. Objects passed to this method may or may
      * not implement equals/hashCode, but it is responsibility of caller to understand what is she or he doing.
-     * <p/>
-     * If the first instance of key elements is an object instance that was returned by this same method, creation of
-     * "subkeys" happens, where original key elements are concatenated with new ones. If the arguments contain only
-     * one argument, and it is an object instance that was returned by this same method, passed in instance is returned.
-     * <p/>
+     * <p>
+     * If the first element of key elements is an object instance that was created by this method, creation of
+     * "subkeys" happens, where original key elements expanded from first element and are concatenated with new ones.
+     * This expansion happens <em>only, if the first key element was already a key created by this class</em>.
+     * If the arguments contain only one argument, and it is an object instance that was created by this method,
+     * the passed in instance is returned unmodified.
+     * <p>
      * Based on what kind of elements are used, one can create multiple kind of keys:
      * <ul>
      *     <li>To create <em>globally matched keys</em>, preferred is to use {@link String} key elements</li>
@@ -89,7 +91,7 @@ public final class Keys {
         private final int hashCode;
 
         private Key(Object[] keys) {
-            this.keys = keys;
+            this.keys = keys.clone();
             this.hashCode = Arrays.hashCode(keys);
         }
 
--- a/maven-resolver-api/src/main/java/org/eclipse/aether/Keys.java
+++ b/maven-resolver-api/src/main/java/org/eclipse/aether/Keys.java
@@ -19,58 +19,88 @@
 package org.eclipse.aether;
 
 import java.util.Arrays;
-import java.util.Collection;
 
 import static java.util.Objects.requireNonNull;
 
 /**
  * A helper class to create keys, to be used with {@link SessionData} and {@link RepositoryCache} instances as keys.
- *
- * It is this class user responsibility to use proper objects as keys, this class merely helps one to create composite
- * keys.
- *
+ * Resolver codebase started with {@link String} keys, that are generally perfect for keys in these constructs, as
+ * long as they are used by components loaded once in system. As we saw, some subsystems like transports can be
+ * loaded multiple times. For example, in case of Maven, some transport may be present in Maven core, but also loaded up
+ * by some extension. In this case, key should distinguish between their {@link ClassLoader}s.
+ * <p/>
+ * It is this class caller responsibility to use proper objects as keys, this class merely helps one to create composite
+ * keys our of object instances that are expected to be anyway good candidates as key. Examples of these objects are
+ * {@link String} instances but also {@link Class} instances, and many other types also usable as keys.
+ * <p/>
  * Important: never forget to perform {@code null}-check, when getting cache from {@link RepositorySystemSession#getCache()}
- * as it may be disabled session-wise.
+ * method as it may return {@code null}, when cache is disabled session-wise.
+ * <p/>
+ * Historical note: As mentioned above, use od {@link String} instances for keys is perfect match, and it worked from
+ * very start. But, as use cases got more and more complex and keys used started to be constructed in more and more
+ * sophisticated ways, but were still {@link String} instances. Believe, or not, the sole purpose of string keys
+ * was easier debugging. By using this helper class, this convenience should remain.
  *
  * @see RepositorySystemSession#getData()
+ * @see SessionData
  * @see RepositorySystemSession#getCache()
+ * @see RepositoryCache
  * @since 2.0.19
  */
 public final class Keys {
     private Keys() {}
 
+    /**
+     * Creates object instance usable as key in session data and cache. Objects passed to this method may or may
+     * not implement equals/hashCode, but it is responsibility of caller to understand what is she or he doing.
+     * <p/>
+     * If the first instance of key elements is an object instance that was returned by this same method, creation of
+     * "subkeys" happens, where original key elements are concatenated with new ones. If the arguments contain only
+     * one argument, and it is an object instance that was returned by this same method, passed in instance is returned.
+     * <p/>
+     * Based on what kind of elements are used, one can create multiple kind of keys:
+     * <ul>
+     *     <li>To create <em>globally matched keys</em>, preferred is to use {@link String} key elements</li>
+     *     <li>To create <em>ClassLoader wide matched keys</em>, make sure at least one key element is {@link Class} that needs to be scoped to ClassLoader</li>
+     *     <li>To create <em>private, matched by creator only keys</em>, make sure to have one key element, that requires instance equality matching, and there is no other same instance of element</li>
+     * </ul>
+     *
+     * @param keys The key elements, it may not be {@code null} and may not have zero elements.
+     * @return An object instance usable as key.
+     */
     public static Object of(Object... keys) {
-        return new CompositeKey(keys);
+        requireNonNull(keys, "keys cannot be null");
+        if (keys.length == 0) {
+            throw new IllegalArgumentException("keys must have at least one element");
+        } else if (keys[0] instanceof Key) {
+            Key head = (Key) keys[0];
+            if (keys.length == 1) {
+                return head;
+            } else {
+                return new Key(concat(head.keys, Arrays.copyOfRange(keys, 1, keys.length)));
+            }
+        } else {
+            return new Key(keys);
+        }
     }
 
-    private static final class CompositeKey {
+    private static final class Key {
         private final Object[] keys;
         private final int hashCode;
 
-        private CompositeKey append(Object... appendedKeys) {
-            requireNonNull(appendedKeys, "appended keys cannot be null");
-            Object[] newKeys = new Object[this.keys.length + appendedKeys.length];
-            System.arraycopy(this.keys, 0, newKeys, 0, this.keys.length);
-            System.arraycopy(appendedKeys, 0, newKeys, this.keys.length, appendedKeys.length);
-            return new CompositeKey(newKeys);
-        }
-
-        private CompositeKey(Object... keys) {
-            this.keys = requireNonNull(keys, "keys cannot be null");
+        private Key(Object[] keys) {
+            this.keys = keys;
             this.hashCode = Arrays.hashCode(keys);
-            if (this.keys.length == 0) {
-                throw new IllegalArgumentException("composite key must have at least one element");
-            }
         }
 
         @Override
         public boolean equals(Object obj) {
             if (this == obj) {
                 return true;
-            } else if (!(obj instanceof CompositeKey)) {
+            } else if (!(obj instanceof Key)) {
                 return false;
             }
-            CompositeKey that = (CompositeKey) obj;
+            Key that = (Key) obj;
             return Arrays.equals(keys, that.keys);
         }
 
@@ -78,5 +108,16 @@ public final class Keys {
         public int hashCode() {
             return hashCode;
         }
+
+        @Override
+        public String toString() {
+            return Arrays.toString(keys);
+        }
+    }
+
+    private static Object[] concat(Object[] one, Object[] two) {
+        Object[] result = Arrays.copyOf(one, one.length + two.length);
+        System.arraycopy(two, 0, result, one.length, two.length);
+        return result;
     }
 }
--- a/maven-resolver-api/src/main/java/org/eclipse/aether/Keys.java
+++ b/maven-resolver-api/src/main/java/org/eclipse/aether/Keys.java
@@ -1,0 +1,82 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.eclipse.aether;
+
+import java.util.Arrays;
+import java.util.Collection;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * A helper class to create keys, to be used with {@link SessionData} and {@link RepositoryCache} instances as keys.
+ *
+ * It is this class user responsibility to use proper objects as keys, this class merely helps one to create composite
+ * keys.
+ *
+ * Important: never forget to perform {@code null}-check, when getting cache from {@link RepositorySystemSession#getCache()}
+ * as it may be disabled session-wise.
+ *
+ * @see RepositorySystemSession#getData()
+ * @see RepositorySystemSession#getCache()
+ * @since 2.0.19
+ */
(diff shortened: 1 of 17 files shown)


slawekjaranowski (approved):

maven-resolver-api/src/main/java/org/eclipse/aether/Keys.java
```diff
@@ -0,0 +1,123 @@
...
+            return new Key(keys);
+        }
+    }
+
+    private static final class Key {
+        private final Object[] keys;
+        private final int hashCode;
+
+        private Key(Object[] keys) {
+            this.keys = keys;
```
gnodet: Minor defensive-programming note: the varargs array is stored directly without copying:

```suggestion
    private Key(Object[] keys) {
        this.keys = keys.clone();
        this.hashCode = Arrays.hashCode(this.keys);
    }
```

All current callers use varargs syntax (which creates a fresh array per call), so this is safe in practice today. But since `Keys` is a public API class in `maven-resolver-api`, a future caller could do:

```java
Object[] arr = { "a", "b" };
Object key = Keys.of(arr);
arr[0] = "changed"; // silently corrupts the key
```

A `clone()` in the constructor would close that off. The cost is negligible since key arrays are tiny. Up to you whether you want to be defensive here — it's the kind of thing that burns someone exactly once, years from now.

maven-resolver-api/src/main/java/org/eclipse/aether/Keys.java
```diff
@@ -0,0 +1,123 @@
...
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * A helper class to create keys, to be used with {@link SessionData} and {@link RepositoryCache} instances as keys.
+ * Resolver codebase started with {@link String} keys, that are generally perfect for keys in these constructs, as
+ * long as they are used by components loaded once in system. As we saw, some subsystems like transports can be
+ * loaded multiple times. For example, in case of Maven, some transport may be present in Maven core, but also loaded up
+ * by some extension. In this case, key should distinguish between their {@link ClassLoader}s.
+ * <p/>
```
gnodet: Very minor Javadoc nits:

1. `<p/>` (lines 31, 33, 36, 39) — self-closing `<p/>` is a no-op in HTML. Standard Javadoc convention is `<p>` (block-level paragraph separator). This appears to be specific to this new file — I didn't find `<p/>` elsewhere in the `maven-resolver-api` sources.

2. Line 33: "keys **our** of object instances" → "keys **out** of object instances"

3. Line 39: "use **od** {@link String}" → "use **of** {@link String}"

maven-resolver-api/src/main/java/org/eclipse/aether/Keys.java
```diff
@@ -0,0 +1,123 @@
...
+     * </ul>
+     *
+     * @param keys The key elements, it may not be {@code null} and may not have zero elements.
+     * @return An object instance usable as key.
+     */
+    public static Object of(Object... keys) {
+        requireNonNull(keys, "keys cannot be null");
+        if (keys.length == 0) {
+            throw new IllegalArgumentException("keys must have at least one element");
+        } else if (keys[0] instanceof Key) {
```
gnodet, line 77: The subkey expansion only happens when a `Key` appears as the **first** element. If a `Key` appears at position > 0, it's treated as a regular element (compared by its own `equals`/`hashCode`). This is probably intentional for the "base key + suffix" use case, but worth mentioning in the Javadoc since it could surprise someone doing:

```java
Object base = Keys.of("prefix");
Object k1 = Keys.of("other", base, "suffix");  // base is NOT expanded
Object k2 = Keys.of(base, "other", "suffix");  // base IS expanded → ["prefix", "other", "suffix"]
```

The current Javadoc says "If the first instance of key elements is an object instance..." — I'd suggest making it slightly more explicit that expansion is first-position-only.

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/filter/RemoteRepositoryFilterSourceSupport.java
```diff
@@ -119,7 +118,7 @@ protected RemoteRepository normalizeRemoteRepository(
     protected String repositoryKey(RepositorySystemSession session, RemoteRepository repository) {
         return repositoryKeyFunctionFactory
                 .repositoryKeyFunction(
-                        FileTrustedChecksumsSourceSupport.class,
+                        RemoteRepositoryFilterSourceSupport.class,
```
gnodet, line 121: Good catch. Confirmed via `git blame`: this was introduced in commit `79b6d5f2` (Repository Key Function SPI, #1679) where the `repositoryKey()` method was added. The `FileTrustedChecksumsSourceSupport.class` reference was clearly copy-pasted from that class's own `repositoryKey()` method and never updated to `RemoteRepositoryFilterSourceSupport.class`.

The practical effect was that `RemoteRepositoryFilterSourceSupport` subclasses and `FileTrustedChecksumsSourceSupport` shared the same repository key function cache entry in `DefaultRepositoryKeyFunctionFactory`. In practice this likely didn't cause visible bugs since both use the same default function, but it's semantically wrong and could mask issues if configuration diverged.

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/filter/DefaultRemoteRepositoryFilterManager.java
```diff
@@ -60,21 +59,22 @@ public DefaultRemoteRepositoryFilterManager(Map<String, RemoteRepositoryFilterSo
...
-            }
-            if (!filters.isEmpty()) {
-                return new Participants(filters);
-            } else {
-                return null;
-            }
-        });
+        return (RemoteRepositoryFilter) session.getData()
+                .computeIfAbsent(
+                        Keys.of(DefaultRemoteRepositoryFilterManager.class, "instance", session.hashCode()), () -> {
```
gnodet: Pre-existing issue (not introduced by this PR), but worth noting: using \`session.hashCode()\` as a key element means that two different derived sessions with a hashCode collision would share the same filter instance. \`Integer.hashCode()\` is just the int value itself, so this relies on \`RepositorySystemSession.hashCode()\` being sufficiently unique. This was the same behavior before the PR (\`INSTANCE_KEY + \".\" + session.hashCode()\`), so no regression here.

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/scope/OptionalDependencySelector.java
```diff
@@ -37,8 +38,8 @@
  * @see Dependency#isOptional()
  */
 public final class OptionalDependencySelector implements DependencySelector {
-    public static final String IGNORED_KEYS = OptionalDependencySelector.class.getName() + ".ignored";
-    public static final String UNSELECTED_KEYS = OptionalDependencySelector.class.getName() + ".unselected";
+    public static final Object IGNORED_KEYS = Keys.of(OptionalDependencySelector.class, "ignored");
```
gnodet, line 41: Note: this changes two \`public static final\` fields from type \`String\` to type \`Object\`. While \`OptionalDependencySelector\` is in the \`internal.impl\` package (so the PR description's claim of "no API breakage in api, spi, util" is accurate), these fields are \`public\` and the class is \`public final\`.

If any external code (e.g., a custom \`DependencySelector\` implementation or a Maven plugin) references \`OptionalDependencySelector.IGNORED_KEYS\` and assigns it to a \`String\` variable, that code will fail to compile. Worth considering whether to keep the \`String\` type here (since these are only used as \`SessionData\` keys, and \`SessionData.get()\` accepts \`Object\`).

That said, the fields were only added in commit \`0447a67c\` (very recently), so the blast radius is likely near zero.

gnodet:
## In-Depth Review: PR #1923 — Fix ApacheTransport auth cache CCEx

### Design Assessment

The `Keys` helper class is a clean, well-considered solution to the ClassLoader-scoping problem. Rather than chasing individual key sites with ad-hoc classloader-aware string concatenation (as the initial fix in `ec6bd858` did with `identityHashCode(classLoader)`), this PR correctly identifies that the **same mechanism** — using `Class` objects as key elements — naturally provides ClassLoader scoping, since `Class.equals()` is identity-based and Class instances are ClassLoader-scoped by definition.

Placing `Keys` in `maven-resolver-api` is the right call: both `SessionData` and `RepositoryCache` live there, and the key contract is defined by those interfaces. The unified factory approach (`Keys.of(...)`) replaces a scattered mix of string concatenation, custom inner classes (`LoggedMirror`), and bare `new Object()` sentinels with a single, consistent pattern.

The three key "flavors" (global/string-only, ClassLoader-aware/Class-element, private/Object-identity) emerge naturally from the element types without needing separate factory methods — this is elegant.

### Key Findings

1. **Defensive copy (Keys.java:91)** — The varargs array is stored without cloning. Safe for all current callers but leaves a footgun in a public API class. See inline comment.

2. **Javadoc nits (Keys.java:31–39)** — `<p/>` (should be `<p>`), "our" → "out", "od" → "of". See inline comment.

3. **Subkey expansion asymmetry (Keys.java:75)** — Only first-position `Key` elements are expanded. The Javadoc could be more explicit about this. See inline comment.

4. **Copy-paste bug fix confirmed (RemoteRepositoryFilterSourceSupport.java:121)** — `git blame` confirms `FileTrustedChecksumsSourceSupport.class` was copy-pasted from #1679 and never corrected. Good catch. See inline comment.

5. **Public field type change (OptionalDependencySelector.java:41-42)** — `IGNORED_KEYS` and `UNSELECTED_KEYS` change from `String` to `Object`. Not in the public API surface (api/spi/util), but these are `public` fields. See inline comment.

6. **`DefaultRemoteRepositoryManager`: `LoggedMirror` removal** — The inner class is replaced with `Keys.of(mirror.getId(), mirror.getUrl(), original.getId(), original.getUrl())`. Semantically equivalent (both use `Arrays.equals`/`hashCode` on the same 4 string elements). Clean simplification.

7. **`DefaultUpdateCheckManager`: private key pattern preserved** — `Keys.of(new Object() { toString() })` correctly maintains the identity-based "private key" semantics of the original bare `new Object()`. The `toString()` override is preserved for debuggability. The parameter type tightening (`Object updateKey` → `String updateKey`) is a correct refinement.

### Correctness Analysis

**The core fix is correct.** The root cause of the ClassCastException is:

1. `ApacheTransporter` stored an `AuthCache` (Apache HttpClient class) in `RepositoryCache` under a string key containing `getClass().getSimpleName()` (just `"ApacheTransporter"`)
2. When the same transport is loaded by two ClassRealms, both instances share the same string key
3. Instance A stores an `AuthCache` from ClassLoader 1
4. Instance B retrieves it and casts to `AuthCache` from ClassLoader 2 → `ClassCastException`

The fix uses `getClass()` (a `Class` object, not its name) as a key element, so instances from different ClassLoaders produce different keys via identity-based `Class.equals()`. This is the correct approach.

**All 15 migration sites use the right key scope:**

| Component | Old pattern | New pattern | Scope correct? |
|-----------|------------|-------------|----------------|
| `ApacheTransporter` | `getSimpleName() + repo` | `Keys.of(getClass(), repo)` | Yes — ClassLoader-aware (transport module) |
| `GlobalState` | `class.getName()` | `Keys.of(class)` | Yes — ClassLoader-aware (transport module) |
| `DataPool` (4 keys) | `class.getName() + "$X"` | `Keys.of(class, "x")` | Yes — ClassLoader-aware (impl, safe for single-load) |
| `PrioritizedComponents` | `class.getName() + disc + hash` | `Keys.of(class, disc, hash)` | Yes — ClassLoader-aware with discriminator |
| `DefaultRepositoryKeyFunctionFactory` | `owner.getName() + ".X"` | `Keys.of(owner, "X")` | Yes — `owner` is already a Class param |
| `DefaultRemoteRepositoryFilterManager` | `class.getName() + ".instance." + sessionHash` | `Keys.of(class, "instance", sessionHash)` | Yes |
| `GroupIdRemoteRepositoryFilterSource` (4 keys) | `getClass().getName() + ".X"` | `Keys.of(class, "X")` | Yes |
| `PrefixesRemoteRepositoryFilterSource` (2 keys) | `getClass().getName() + ".X"` | `Keys.of(class, "X")` | Yes |
| `TrustedChecksumsArtifactResolverPostProcessor` | `class.getName() + ".X"` | `Keys.of(class, "X")` | Yes |
| `OptionalDependencySelector` (2 keys) | `class.getName() + ".X"` | `Keys.of(class, "X")` | Yes |
| `NamedLockFactoryAdapterFactoryImpl` | `class.getName() + ".adapter"` | `Keys.of(class, "adapter")` | Yes |
| `SmartExecutorUtils` | `class.getSimpleName() + "-" + prefix` | `Keys.of(class, prefix)` | Yes |
| `DefaultRemoteRepositoryManager` | `LoggedMirror` inner class | `Keys.of(4 strings)` | Yes — global (no CL scoping needed) |
| `DefaultUpdateCheckManager` | `new Object()` sentinel | `Keys.of(new Object())` | Yes — private key |

### API & Compatibility

The PR description's claim of "no API breakage in public surface of api, spi, util modules" is **verified as correct**:

- `Keys` is a **new** class in `maven-resolver-api` — purely additive
- No method signatures in api/spi/util change
- No public classes removed

In `maven-resolver-impl` (internal, not public API):
- `OptionalDependencySelector.IGNORED_KEYS` / `UNSELECTED_KEYS` change from `String` to `Object` — source-incompatible for any code referencing these as `String`, but these fields were added very recently (#1899) and are in `internal.impl`
- `DefaultUpdateCheckManager.isAlreadyUpdated` / `setUpdated` parameter type tightens from `Object` to `String` — package-private methods, no external impact
- `LoggedMirror` inner class removed — was `private`, no external impact

### Test Coverage

`KeysTest.java` covers the three key flavors well:
- `globalKeys()` — string-only keys, equality and inequality
- `classLoaderAwareKeys()` — two ClassLoaders loading the same class, verifies distinct keys
- `privateKeys()` — Object-identity keys
- `subKey()` — key expansion/concatenation
- Error cases: `noElement()`, `nullVararg()`

**Gaps:**
- No test for `null` as a key element (e.g., `Keys.of("a", null, "b")`) — currently this silently creates a key with a null element. Worth either rejecting or documenting.
- No test for mixed-type key elements (e.g., `Keys.of("string", 42, SomeClass.class)`) — would increase confidence in the `Arrays.hashCode`/`Arrays.equals` delegation
- No test for `toString()` output format (useful for debugging scenarios the Javadoc mentions)
- The `classLoaderAwareKeys` test uses `Paths.get("target/classes")` — this works with Maven Surefire's working directory convention but could be fragile in IDE-based test runners
- No integration test that would catch a ClassCastException regression in the actual transport layer. Understandable given the complexity of setting up multi-ClassLoader Maven-like scenarios, but the gap is worth noting.

### Minor Notes

- The `DefaultUpdateCheckManager.SESSION_CHECKS` comment `// instance bound private key` is helpful — consider adding similar comments at other sites where the key scope choice is non-obvious
- `GlobalState.java` still has its own `CompoundKey` inner class (used for `userTokens`/`expectContinues`) — this is correct, as those keys serve a different purpose than the session cache key

### Verdict

The fix is correct and well-designed. The `Keys` abstraction is a clean improvement over the scattered key patterns, and the ClassLoader-scoping mechanism is sound. The migration is comprehensive and each site uses the appropriate key scope.

The findings above are all minor — the most actionable ones are the Javadoc typos and the optional defensive copy. The copy-paste bug fix in `RemoteRepositoryFilterSourceSupport` is a nice bonus find.

Overall: this looks good to merge, pending the minor fixes above.

---
_In-depth review by Claude Code on behalf of @gnodet_

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/scope/OptionalDependencySelector.java
cstamas: Yes, these two keys are new, and are not yet present in any released code.

PR #1933, opened 2026-06-24
Get rid of BND warnings

cstamas:
By properly formatting bnd parameters (follow emitted advice, add trailing `:`).

---

FTR, warning was:
```
[INFO] --- bnd:7.3.0:bnd-process (bnd-process) @ maven-resolver-spi ---
[WARNING] /home/cstamas/Worx/apache-maven/RELEASE/maven-resolver/maven-resolver-spi/pom.xml [18:74]: No value specified for key: -jpms-module-info. An empty value should be specified as '-jpms-module-info:' or '-jpms-module-info=': <<              -jpms-module-info
      -removeheaders Automatic-Module-Name>>
```

### DIFF
--- a/maven-resolver-api/pom.xml
+++ b/maven-resolver-api/pom.xml
@@ -34,7 +34,7 @@
   <properties>
     <bnd.instructions.additions><![CDATA[
       -jpms-module-info: org.apache.maven.resolver
-      -removeheaders Automatic-Module-Name
+      -removeheaders: Automatic-Module-Name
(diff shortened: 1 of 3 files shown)


gnodet (approved):
Clean and correct fix. The three affected modules are the only ones using BND instruction syntax without the required colon separator — the other modules with `bnd.instructions.additions` use standard OSGi manifest headers that already include colons.

LGTM.

_Fully automatic review from Claude Code_

PR #1947, opened 2026-07-02
Use try-with-resources in AbstractTransporter

### LINKED ISSUE #1521: Use try-with-resources

There's a lot of code like this scattered throughout the project:

```
        FileOutputStream fos = null;
        try {
            fos = new FileOutputStream(file);

            if (data != null) {
                fos.write(data.getBytes(StandardCharsets.UTF_8));
            }

            fos.close();
            fos = null;
        } finally {
            try {
                if (fos != null) {
                    fos.close();
                }
            } catch (final IOException e) {
                // Suppressed due to an exception already thrown in the try block.
            }
        }
```

Use try-with-resources from Java 7

Aayush10016:
## Summary

- Refactor `AbstractTransporter` stream handling to use try-with-resources for the close-on-completion paths.
- Preserve the existing `close=false` behavior by leaving download input streams open and flushing upload output streams.
- Add focused tests for the `utilGet` and `utilPut` close/flush contracts.

Closes apache/maven-resolver#1521.

## Verification

- `mvn -pl maven-resolver-spi -am test` with Temurin JDK 21.0.11

### DIFF
--- a/maven-resolver-spi/src/main/java/org/eclipse/aether/spi/connector/transport/AbstractTransporter.java
+++ b/maven-resolver-spi/src/main/java/org/eclipse/aether/spi/connector/transport/AbstractTransporter.java
@@ -90,12 +90,16 @@ public abstract class AbstractTransporter implements Transporter {
      */
     protected void utilGet(GetTask task, InputStream is, boolean close, long length, boolean resume)
             throws IOException, TransferCancelledException {
-        try (OutputStream os = task.newOutputStream(resume)) {
-            task.getListener().transportStarted(resume ? task.getResumeOffset() : 0L, length);
-            copy(os, is, task.getListener());
-        } finally {
-            if (close) {
-                is.close();
+        if (close) {
+            try (InputStream input = is;
+                    OutputStream os = task.newOutputStream(resume)) {
+                task.getListener().transportStarted(resume ? task.getResumeOffset() : 0L, length);
+                copy(os, input, task.getListener());
+            }
+        } else {
+            try (OutputStream os = task.newOutputStream(resume)) {
+                task.getListener().transportStarted(resume ? task.getResumeOffset() : 0L, length);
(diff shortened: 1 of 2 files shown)


cstamas:
See https://github.com/apache/maven-resolver/blob/master/.github/pull_request_template.md
Have you ICLA?

gnodet (commented):
## Review Summary

Nice cleanup, @Aayush10016! The try-with-resources conversion in `AbstractTransporter` is correct and actually **improves** exception handling over the original code.

### What's good

In the original code, `is.close()` (in `utilGet`) and `os.close()` (in `utilPut`) were called in plain `finally` blocks, meaning their exceptions could **replace** the primary exception from the `try` block. The try-with-resources approach properly adds close exceptions as *suppressed* exceptions instead — this is the correct Java behavior.

The tests are well-focused and verify the close/flush contract for both `close=true` and `close=false` branches.

### Minor observations (non-blocking)

- **Code duplication:** The `transportStarted` + `copy` logic is duplicated in both branches of the `if`/`else` for each method. An alternative is to extract the common body into a private helper method, though the current form with 2-line duplication is clear and defensible.

- **Exception-path tests:** The tests cover the happy path well, but don't test the exception path — e.g., verifying streams are still properly closed when `copy()` throws. Since improved exception handling is a key benefit of this refactor, an exception-path test would strengthen confidence. This is a nice-to-have, not a blocker.

- **Pre-existing:** In the `utilPut` `close=false` branch, `os.flush()` remains in a plain `finally` block, meaning if `copy()` throws and then `flush()` also throws, the flush exception replaces the original. This is unchanged by this PR and could be addressed in a follow-up.

Overall this is a solid improvement. The contributor checklist items noted in the maintainer comment should still be addressed.

---

kwin (approved):
LGTM, thanks

PR #1948, opened 2026-07-05
Bug: Jetty transport used wrong config

cstamas:
Wrong timeout is used.


cstamas:
Windows got 403 from central :disappointed:

kwin (approved):

PR #1953, opened 2026-07-08
Bug: Apache transport deadlock in concurrent use case

cstamas:
Make sure the class SSLConnectionSocketFactory is initialized early, instead lazily in ConcurrentHashMap.computeIfAbsent, as latter may cause deadlock.

Refs:
* https://github.com/quarkusio/quarkus/issues/55317
* https://github.com/quarkusio/quarkus/pull/55345

### DIFF
--- a/maven-resolver-transport-apache/src/main/java/org/eclipse/aether/transport/apache/GlobalState.java
+++ b/maven-resolver-transport-apache/src/main/java/org/eclipse/aether/transport/apache/GlobalState.java
@@ -54,6 +54,13 @@ import org.eclipse.aether.util.ConfigUtils;
  */
 final class GlobalState implements Closeable {
 
+    static {
+        // force initialization of SSLConnectionSocketFactory class:
+        // ensure that the connection socket factory is initialized before we start using it in any multithreaded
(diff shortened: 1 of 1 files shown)


gnodet (approved):
Clean, well-targeted fix for a real deadlock — the pattern (eager class initialization to avoid the documented `ConcurrentHashMap.computeIfAbsent` hazard) is sound and well-established.

**Evaluation:**
- **Correctness:** The JVM guarantees `GlobalState.<clinit>` completes before any static/instance members are used, so `SSLConnectionSocketFactory` is fully initialized before `computeIfAbsent` is ever reached.
- **Sufficiency:** A single call to `getDefaultHostnameVerifier()` triggers `SSLConnectionSocketFactory.<clinit>`, which initializes the class once. All three usage points (lines 153, 166, 190) then access an already-initialized class.
- **Scope:** Minimal, focused change — exactly what's needed.
- **Tests:** No tests added, but acceptable — class-initialization deadlocks are inherently timing-sensitive and non-deterministic.

**Minor observation (non-blocking):** The comment references the Quarkus PR (`quarkus/pull/55345`) which is the downstream workaround. Consider also linking to the Quarkus issue (`quarkus/issues/55317`) which contains the detailed thread-dump and root-cause analysis, for future maintainer context.

PR #1958, opened 2026-07-13
Validate path components in DefaultLocalPathComposer

gnodet:
## Summary

- Add validation in `DefaultLocalPathComposer` to reject coordinate components (groupId, artifactId, version, classifier, extension) that contain path traversal sequences (`..`) or path separator characters (`/`, `\`) before composing the local repository path.
- This is consistent with the existing path validation in `FileTransporter.getPath()` (line 166).
- Add `DefaultLocalPathComposerTest` with tests for normal and malformed coordinates.

## Test plan

- [x] New `DefaultLocalPathComposerTest` passes (7 tests — normal paths, traversal rejection, slash rejection, backslash rejection)
- [x] Existing `SimpleLocalRepositoryManagerTest` passes (3/3)
- [x] Existing `EnhancedLocalRepositoryManagerTest` passes (12/12)

### DIFF
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultLocalPathComposer.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultLocalPathComposer.java
@@ -38,6 +38,13 @@ public final class DefaultLocalPathComposer implements LocalPathComposer {
     public String getPathForArtifact(Artifact artifact, boolean local) {
         requireNonNull(artifact);
 
+        validatePathComponent(artifact.getGroupId(), "groupId");
+        validatePathComponent(artifact.getArtifactId(), "artifactId");
+        validatePathComponent(artifact.getBaseVersion(), "baseVersion");
+        validatePathComponent(artifact.getVersion(), "version");
+        validatePathComponent(artifact.getClassifier(), "classifier");
+        validatePathComponent(artifact.getExtension(), "extension");
+
         StringBuilder path = new StringBuilder(128);
 
         path.append(artifact.getGroupId().replace('.', '/')).append('/');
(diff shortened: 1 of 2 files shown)


gnodet (commented):
## AI-Assisted Review — PR #1958 (Validate path components in DefaultLocalPathComposer)

This is a well-scoped security hardening PR — nice work! The validation is correct, purely additive, and more thorough than the existing precedent in `FileTransporter.getPath()`.

### What works well

- **Correct validation logic**: Checking for `..`, `/`, and `\` independently catches traversal sequences, forward-slash injection, and backslash-based traversal. This is more robust than the existing `FileTransporter.getPath()` check which only catches `"../"`.
- **Null/empty safety**: `validatePathComponent` correctly short-circuits on null and empty values, avoiding false rejections on legitimate empty classifiers.
- **Good test coverage**: The 7 new tests cover the core attack vectors (traversal sequences, slashes, backslashes) and verify that normal version strings with dots and dashes remain unaffected.

### Minor observations (non-blocking)

**1. `metadata.getType()` and `repositoryKey` are not validated**

Both flow into the composed path but aren't passed through `validatePathComponent`. However, these are internally derived rather than attacker-controlled: `repositoryKey` comes from `RepositoryIdHelper` (which uses `PathUtils.stringToPathSegment()` for sanitization) or is the constant `"local"`, and `metadata.getType()` is typically `"maven-metadata.xml"` or internal constants. This is a defense-in-depth gap rather than an exploitable issue — worth noting for a future hardening pass.

**2. Test coverage for non-version fields**

The traversal rejection tests only exercise the `version` field. Since all fields use the same `validatePathComponent` method the logic is covered, but adding at least one test per field (groupId, artifactId, classifier, extension) would strengthen completeness.

Overall this is ready to merge — a clean security improvement at the right layer.

cstamas:
I propose alt solution https://github.com/apache/maven-resolver/pull/1959
And once 2.0.21 is out, Maven implementations of `DefaultVersionResolver` and `DefaultVersionRangeResolver` should reuse it as well.

PR #1959, opened 2026-07-13
Validate path components

cstamas:
But make it reusable as well. When 2.0.21 released, Maven codebase of `DefaultVersionResolver` and `DefaultVersionRangeResolver` should be updated too, and use this util method.

Supersedes (and based on) #1958

### DIFF
--- a/maven-resolver-util/src/main/java/org/eclipse/aether/util/PathUtils.java
+++ b/maven-resolver-util/src/main/java/org/eclipse/aether/util/PathUtils.java
@@ -22,6 +22,9 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 
+import org.eclipse.aether.artifact.Artifact;
+import org.eclipse.aether.metadata.Metadata;
+
 import static java.util.Objects.requireNonNull;
 
 /**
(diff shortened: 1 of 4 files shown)


gnodet (commented):
## AI-Assisted Review — PR #1959 (Validate path components)

The core contribution here — extracting path traversal validation into `PathUtils` for cross-project reuse and applying it at the `DefaultLocalPathComposer` boundary — is well-designed and correct. The `equals("..") + contains("/") + contains("\\")` approach is sound and more precise than PR #1958's `contains("..")`, with the reasoning properly documented in code comments ("version '1..' is valid").

### Verified concern: `GenericVersion` constructor validation

The addition of `validatePathComponent` inside the `GenericVersion` constructor conflates filesystem-safety with version parsing. `GenericVersion` is a pure version-parsing/comparison class used for version ranges, dependency resolution ordering, constraint evaluation, and version comparison — none of which involve filesystem paths.

**The `IllegalArgumentException` escapes through `parseVersion()` uncaught.** The call chain:

1. `GenericVersionScheme.parseVersion(String)` declares `throws InvalidVersionSpecificationException` (checked)
2. It calls `new GenericVersion(version)` → `validatePathComponent()` → throws `IllegalArgumentException` (unchecked)
3. Callers like `AbstractPatternDependencyFilter.isVersionIncludedInRange()` catch only `InvalidVersionSpecificationException` — the unchecked exception propagates unhandled

This is a behavioral change to a widely-used public API. Any downstream code using `parseVersion()` with a version string containing `/` or `\` (unusual but not currently invalid) would get an uncaught `IllegalArgumentException` instead of a successful `Version` object.

### Suggestion

Consider keeping the `PathUtils` extraction and `DefaultLocalPathComposer` integration (which is the right enforcement point — where versions become filesystem path segments), but removing the `GenericVersion` constructor change. If version-level validation is desired as defense-in-depth, it could either:
- Throw `InvalidVersionSpecificationException` instead (if path traversal is considered a version *syntax* violation), or
- Be applied at a different layer (e.g., in Maven's `DefaultVersionResolver`)

### Other notes

- The metadata type exclusion from validation is correctly documented (`.meta/prefixes.txt` contains legitimate path separators)
- This PR completely supersedes PR #1958 — if merged, #1958 should be closed
- Test coverage for the core path validation through `DefaultLocalPathComposerTest` is solid; direct unit tests for `PathUtils.validatePathComponent` and the `GenericVersion` constructor behavior would strengthen coverage

cstamas:
Agreed with assessment: will remove `validatePathComponent` from `GenericVersion` ctor, and it should really go into Maven stuff.

gnodet:
Thanks for the quick update, @cstamas! Removing the `validatePathComponent` from the `GenericVersion` constructor looks right — this keeps the validation at the path-composition layer where it belongs, without affecting version parsing semantics. The remaining changes (PathUtils extraction + DefaultLocalPathComposer validation) look good.

maven-resolver-util/src/main/java/org/eclipse/aether/util/PathUtils.java
```diff
@@ -72,4 +75,50 @@ public static String stringToPathSegment(String string) {
...
+     *
+     * @since 2.0.21
+     */
+    public static void validatePathComponent(String value, String label) {
+        if (value != null && !value.isEmpty()) {
+            // Important: "equals .." and not "contains ..", as if escape attempted, it will contain path separators
+            // OTOH: version "1.." is valid version string!
+            if (value.equals("..") || value.contains("/") || value.contains("\\")) {
+                throw new IllegalArgumentException(
+                        "Invalid " + label + ": must not contain '..', '/' or '\\': " + value);
```
atokle, line 92: Should maybe be:
"Invalid " + label + ": must not be equal to '..' or contain '/' or '\\': " + value);

PR #1965, opened 2026-07-16
Use custom doclet to extract configuration metadata with the help of

kwin:
javadoc


maven-resolver-tools/pom.xml
```diff
@@ -199,8 +204,8 @@
             <phase>package</phase>
             <configuration>
               <mainClass>org.eclipse.aether.tools.CollectConfiguration</mainClass>
+              <classpathScope>test</classpathScope>
               <arguments>
-                <argument>--mode=resolver</argument>
                 <argument>--templates=configuration.properties,configuration.yaml</argument>
```
kwin, line 242: Are those generated artifacts actually used anywhere @cstamas?

gnodet (commented):
## Review Summary

Nice work, @kwin! Replacing the Roaster + ASM dependencies with a standard JDK javadoc doclet is a clean architectural improvement. The intermediate properties file format provides good separation between metadata extraction and rendering, and the `fileContains` pre-filtering to avoid classpath resolution issues is a smart optimization. The behavioral equivalence with the old code looks correct — e.g., the `Long` suffix stripping is correctly omitted since `VariableElement.getConstantValue()` returns a proper `Long` object.

Two minor observations on the new code:

### Low Severity

1. **Redundant `BufferedWriter` wrapping** — In `ConfigurationCollectorDoclet.java`, `Files.newBufferedWriter()` already returns a `BufferedWriter`, so the outer `new BufferedWriter(...)` wrapper is unnecessary. Harmless, but an extra allocation:

   ```suggestion
   try (Writer writer = Files.newBufferedWriter(output, StandardCharsets.UTF_8)) {
   ```

2. **Broad exception catch in `fileContains`** — In `CollectConfiguration.java`, the `fileContains` method catches `Exception` instead of `IOException`. Since `Files.readString()` only throws `IOException` as a checked exception, catching the broader `Exception` could mask programming errors (e.g., `NullPointerException`) by silently returning `false`. Consider narrowing to `IOException`.

### Notes

- The PR title appears truncated: "Use custom doclet to extract configuration metadata with the help of" — consider completing it.
- CI checks are still pending at time of review.

---

cstamas:
The maven mode is used in maven:
https://github.com/apache/maven/blob/7aa3c8a37b091a5a86d3dae3a7d99ce910fd6caa/pom.xml#L1043

maven-resolver-tools/pom.xml
cstamas: The properties and yaml files are deployed: https://repo.maven.apache.org/maven2/org/apache/maven/resolver/maven-resolver-tools/2.0.20/

But are not yet used anywhere AFAIK. The idea was to allow some other app (IDE? CLI?) to pick them up and provide some programmatic help/assistance with config, but we never got there.

@gnodet ?

gnodet (commented):
## Follow-up Review

Thanks for the quick update, @kwin! Both findings from the initial review have been addressed in the new commit:

- ✅ Redundant `BufferedWriter` wrapping removed
- ✅ Broad `Exception` catch narrowed to `IOException`

The new enum resolution logic is well-structured — the fallback chain from compile-time constant → enum constant kind → initializer expression analysis covers the practical cases nicely. The test fixture exercises both indirect enum references (`DEFAULT_ENUM = SampleEnum.VALUE_A`) and direct enum constant references (`SampleEnum#VALUE_B`), which is good coverage.

Three minor observations on the new code (all low severity):

1. **Stale assertion message** — `assertEquals(4, ...)` says "expected three configuration keys" (see inline)
2. **Redundant `docTrees.getTree(field)` call** — called twice in `resolveEnumReference`, could use Java 16+ pattern matching `instanceof` (see inline)
3. **Theoretical null from `docTrees.getPath()`** — `resolveReferencedField` passes the result of `docTrees.getPath(contextField)` directly to `new DocTreePath(...)` without a null check. In practice this is very unlikely since the doclet only processes included source elements, but a null guard would be more defensive.

None of these are blocking — the PR looks good overall.

---

maven-resolver-tools/src/test/java/org/eclipse/aether/tools/ConfigurationCollectorDocletTest.java
```diff
@@ -0,0 +1,142 @@
...
+        try (InputStream in = ConfigurationCollectorDocletTest.class.getResourceAsStream(FIXTURE)) {
+            assertNotNull(in, "fixture source not found on classpath: " + FIXTURE);
+            Files.copy(in, sourceFile);
+        }
+        Path output = tempDir.resolve("configuration-keys.properties");
+
+        runDoclet(sourceFile, output);
+
+        Map<String, Map<String, String>> keys = readKeys(output);
+        assertEquals(4, keys.size(), "expected three configuration keys");
```
gnodet: Nit: the assertion expects 4 keys but the failure message still says "three":

```suggestion
        assertEquals(4, keys.size(), "expected four configuration keys");
```

maven-resolver-tools/src/main/java/org/eclipse/aether/tools/ConfigurationCollectorDoclet.java
```diff
@@ -0,0 +1,478 @@
...
+    /**
+     * Resolves an enum constant that a field is initialized with, including the enum type in the result
+     * (e.g. a field declared as {@code SomeEnum FOO = SomeEnum.VALUE} resolves to {@code SomeEnum.VALUE}).
+     * Returns {@code null} if the field's initializer is not a simple enum reference.
+     */
+    private String resolveEnumReference(VariableElement field) {
+        if (!(docTrees.getTree(field) instanceof VariableTree)) {
+            return null;
+        }
+        ExpressionTree initializer = ((VariableTree) docTrees.getTree(field)).getInitializer();
```
gnodet: Minor: `docTrees.getTree(field)` is called twice — once for the `instanceof` guard and again for the cast. Since the project targets Java 17+, pattern matching `instanceof` eliminates the redundant call:

```suggestion
    private String resolveEnumReference(VariableElement field) {
        if (!(docTrees.getTree(field) instanceof VariableTree variableTree)) {
            return null;
        }
        ExpressionTree initializer = variableTree.getInitializer();
```

gnodet (commented):
## Follow-up Review (commits 3 & 4)

Thanks for applying the suggestions from the last review, @kwin — commit 4 (`727e16bf`) correctly adopts pattern matching `instanceof` and fixes the assertion message. Both look good.

Regarding commit 3 ("Restore maven mode"), the dual-mode architecture is clean — the split into `processResolverField` / `processMavenField` keeps the two code paths well-separated, and the `@Config` annotation attribute reading is correct. The mode is properly threaded from picocli CLI through to the doclet. A few observations:

### Medium

1. **No test coverage for Maven mode** — The existing test only exercises the resolver path (Javadoc block tags). The Maven mode has meaningfully different logic (annotation-based metadata extraction), and without a test fixture that includes an `@Config`-annotated field, regressions in `processMavenField` / `getAnnotation` would go undetected. Consider adding a second fixture and test method similar to the existing one but with `--mode=maven`.

### Low

2. **No mode validation in doclet** — The `mode` field is a raw `String` set by the `--mode` option. An invalid value (e.g. `--mode=typo`) silently falls through to resolver mode since `doRun` checks `"maven".equals(mode)` and treats everything else as resolver. Practical risk is low because `CollectConfiguration` validates via the picocli `Mode` enum, but if the doclet is ever invoked standalone, this could be surprising.

3. **Dead null check on `description`** — In `processMavenField`, the `if (description != null)` guard on the `replace("*", ...)` call is unreachable: when `docComment != null`, `renderContent(getFullBody())` returns a non-null `String`; when `docComment == null`, `description` is `""`. Harmless, but slightly misleading.

### Notes

- The PR description still says "Remove unused 'maven' mode" — consider updating it to reflect that Maven mode is now restored.
- Good use of `Locale.ROOT` in `source.toLowerCase()` — an improvement over the default-locale alternative.

---

kwin:
Long term I would rather move the Maven specific part to https://github.com/apache/maven so that it does not need to use reflection (to prevent dependencies from Resolver to Maven). Therefore I wouldn't spend time here now to add additional tests for Maven. For example https://maven.apache.org/ref/4.0.0-rc-5/maven-configuration.html still emits `system_properties` as source, as there is no mapping defined for all source enums (https://github.com/apache/maven-resolver/blob/79d102b66235f33ad1e6134e18451ac3ee91b44a/maven-resolver-tools/src/main/java/org/eclipse/aether/tools/CollectConfiguration.java#L199).

maven-resolver-tools/pom.xml
gnodet: Right, these are not used yet by any consumer that I'm aware of. The original idea was to provide a machine-readable catalog of all configuration keys so that external tools (IDE plugins, CLI helpers, etc.) could offer auto-completion, validation, or contextual help for resolver/maven configuration properties.

I still think it's a good idea to keep deploying them — they're cheap to generate and having a stable, versioned catalog of configuration keys published to Central makes it easy for tooling to pick them up whenever someone builds that integration. The format is straightforward (key/type/default/description/since), so it wouldn't be hard to consume.

That said, if we feel they add noise to the release, we could also consider not attaching them as separate artifacts and instead embedding the properties file inside the jar as a classpath resource (e.g. `META-INF/maven-resolver/configuration.properties`). That would still make them discoverable but without publishing extra artifacts.

kwin: Let's leave here for the time being.

gnodet (commented):
**Follow-up review #4 — commit `154e76df` (mode validation & cleanup)**

Nice work — this commit addresses two of the three findings from the previous round:

✅ **Mode validation** (previously flagged): The explicit `"resolver".equals(mode)` check with `else throw IllegalArgumentException` correctly prevents unknown modes from silently falling through to resolver processing. The exception is caught by the `RuntimeException` handler in `run()` and properly reported.

✅ **Dead null check** (previously flagged): The `if (description != null)` guard was correctly removed — the ternary always produces a non-null string.

✅ **`system_properties` source**: Good addition completing the enum mapping alongside `model` and `user_properties`.

**One low-severity observation** (not blocking): The mode validation happens inside the field iteration loop (line 157). If scanned sources happened to contain no constant fields, the loop body would never execute and an invalid mode would silently succeed. In practice this is unreachable because `CollectConfiguration` constrains the mode to valid enum values and pre-filters sources to those containing configuration markers — but moving the validation to the top of `doRun()` (alongside the existing `output == null` check) would make the invariant self-documenting.

**Still open from prior rounds**: No test coverage for the `maven` mode path or the new `IllegalArgumentException` path. Not blocking for this commit.

cstamas (approved):

kwin:
This requires some more work for error reporting and also for emitting types (should be FQCN for non simple types)

gnodet (commented):
**Follow-up review #5** — commits 6–8 (pattern matching, mode validation, FQCN enum types)

Great progress — commits 6-7 address our prior review findings (pattern matching, dead null check, mode validation), and commit 8 is a substantial improvement with proper FQCN resolution using `DocTreePath`/`DocTrees.getElement()`. The `Reporter` integration replacing `System.err` is also a nice touch.

**Findings:**

**1. [Important] Wrong tag name in `extractClassLink()` error messages**

`extractClassLink()` processes `@configurationType` content:
```java
String configurationType =
    getConfigurationType(extractClassLink(field, docComment, blockTags.get("configurationType")));
```

But all three error messages inside the method reference `@configurationDefaultValue`:
```java
throw new IllegalArgumentException("Missing content for @configurationDefaultValue");
// ...
"Expected a class link in @configurationDefaultValue, but got a member reference: ..."
// ...
"No valid {@link ...} reference found in @configurationDefaultValue"
```

This would send developers chasing the wrong tag when they hit these errors. Should be `@configurationType` in all three messages.

**2. [Minor] Stale javadoc on `resolveEnumReference()`**

The method javadoc still says it returns "including the enum type in the result (e.g. `SomeEnum.VALUE`)", but after commit 8 it now returns just the constant name (e.g., `VALUE`). The example contradicts the behavior.

**Non-blocking observations:**
- The `{@link ...}` prefix/suffix stripping in `getConfigurationType()` is now dead code (since `extractClassLink()` returns a FQCN directly), but the `java.lang.` stripping is still needed. Minor cleanup opportunity.
- In `lookupConstant()`, the else branch uses `field.getConstantValue()` again instead of the already-extracted `value` variable.

**CI:** Still running (all jobs pending at time of review).

**Verdict:** Solid improvements overall. Finding #1 is the only actionable item — a quick fix to swap the tag name in 3 error messages.

---
_Reviewed with AI on behalf of @gnodet. This review was generated by an AI agent and may contain inaccuracies; please verify all suggestions before applying._

PR #1762, opened 2026-01-26
Expose additional transport details to TransportListener

kwin:
For HTTP Transporters this is:
- HTTP Version
- SSL Protocol (only HTTPS)
- SSL Cipher Suite (only HTTPS)
- Compression Algorithm (if used)

This closes #1761

Following this checklist to help us incorporate your
contribution quickly and easily:

- [ ] Your pull request should address just one issue, without pulling in other changes.
- [ ] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [ ] Each commit in the pull request should have a meaningful subject line and body.
  Note that commits might be squashed by a maintainer on merge.
- [ ] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied.
  This may not always be possible but is a best-practice.
- [ ] Run `mvn verify` to make sure basic checks pass.
  A more thorough check will be performed on your pull request automatically.
- [ ] You have run the integration tests successfully (`mvn -Prun-its verify`).

If your pull request is about ~20 lines of code you don't need to sign an
[Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf) if you are unsure
please ask on the developers list.

To make clear that you license your contribution under
the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
you have to acknowledge this by using the following check-box.

- [ ] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
- [ ] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

PR #1966, opened 2026-07-16
Feat: new (limited) transport

cstamas:
Many applications (outside of Maven) integrate Resolver and their use case is almost always "consume" (use Resolver to resolve artifacts). Also, many of those applications are still Java 8 level, hence, the JDK transport for them is no-go. On the other hand, there is no lightweight replacement for them, except to use some "heavyweight" transporter, but in case of CLI applications this is usually undesirable, as they count every byte.

Transitive hull sizes of several existing transports:
* jdk - 1.5MB / Java 11+
* wagon - 1.9MB (without any provider; unusable like this) / Java 8+
* apache - 2.4MB / Java 8+
* jetty - 10.9MB / Java 11+
* minio - 23.5MB / Java 8+

The new transport has no dependencies and small size:
* url - 870 KB


maven-resolver-transport-url/src/main/java/org/eclipse/aether/transport/url/UrlTransporter.java
```diff
@@ -0,0 +1,284 @@
...
+import org.eclipse.aether.spi.connector.transport.http.HttpConstants;
+import org.eclipse.aether.spi.connector.transport.http.HttpTransporter;
+import org.eclipse.aether.spi.connector.transport.http.HttpTransporterException;
+import org.eclipse.aether.spi.io.PathProcessor;
+import org.eclipse.aether.transfer.NoTransporterException;
+import org.eclipse.aether.util.connector.transport.http.HttpTransporterUtils;
+
+/**
+ * A special, "read only" and limited capability transport usable for bootstrapping. It provides HTTP with minimal
+ * support (only basic auth, only GET/HEAD).
```
kwin: Please mention which HTTP library is used here and in the module description

maven-resolver-transport-url/README.md
```diff
@@ -0,0 +1,19 @@
+# Maven Resolver URL Transport
```
kwin, line 18: Shouldn't this rather be added to https://github.com/apache/maven-resolver/blob/master/src/site/markdown/transporter-known-issues.md?

maven-resolver-transport-url/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+This is special Transport with limited HTTP capabilities. The main use case of this transport is outside of
+Maven, but in case of other apps that are Java 8, and integrate Resolver mostly for **consumption purposes**.
```
kwin:
```suggestion
Maven, but in case of other apps that are Java 8, and integrate Resolver only for **consumption purposes**.
```

gnodet (commented):
**Initial review of PR #1966 — new URL transport**

Clean design for a lightweight transport that fills a real gap for Java 8+ consumers. The negative priority (-1.0f) correctly ensures it won't be auto-selected over more capable transports, and the test class properly disables unsupported scenarios. Good reuse of `HttpTransporterUtils` and `HttpTransporterTest`.

However, the manual redirect implementation has several issues compared to what `HttpClient.Redirect.NORMAL` provides for free in the JDK transport:

### Findings

**🔴 Connection resource leak in `perform()` recursion (high)**

When a redirect (301/302) or auth retry (401/407) triggers the recursive `return perform(...)` call, the current `HttpURLConnection` is abandoned without calling `con.disconnect()`. Since the response body is not consumed on these status codes, the connection cannot be returned to the JDK connection pool. Up to `MAX_REDIRECTS` (5) plus 2 auth-retry connections can leak sockets per request. Each recursive branch should call `con.disconnect()` before the recursive return.

The same issue applies in `implPeek` which never disconnects.

**🟡 Credentials forwarded on cross-origin redirects (medium, security)**

When following a redirect to a different host, `currAuth` and `currProxyAuth` are passed unchanged to the recursive call, causing the `Authorization` header to be sent to the redirect target. Per RFC 7235, credentials should be stripped when the redirect target has a different authority (host:port). The JDK transport uses `HttpClient.Redirect.NORMAL` which handles this automatically.

**🟡 HTTPS-to-HTTP redirect downgrade allowed (medium, security)**

The scheme validation checks that the redirect uses HTTP or HTTPS, but does not reject a redirect from HTTPS to HTTP. The JDK transport's `Redirect.NORMAL` explicitly prevents this. Combined with the credential forwarding issue above, this could cause credentials to be transmitted in plaintext.

**🟡 Missing 307/308 redirect codes (medium)**

Only 301 (`HTTP_MOVED_PERM`) and 302 (`HTTP_MOVED_TEMP`) are handled. 307 (Temporary Redirect) and 308 (Permanent Redirect) are commonly used in practice. Both preserve the HTTP method, which is safe for this read-only (GET/HEAD) transport.

### Suggested fix for redirect handling

All four redirect/security issues can be addressed together in `perform()`:

```java
} else if (responseCode == HttpURLConnection.HTTP_MOVED_PERM
        || responseCode == HttpURLConnection.HTTP_MOVED_TEMP
        || responseCode == 307
        || responseCode == 308) {
    String location = con.getHeaderField(HEADER_LOCATION);
    if (location == null) {
        throw new IOException("Redirect response missing Location header");
    }
    URI redirectUri = URI.create(con.getURL().toString()).resolve(location);
    String scheme = redirectUri.getScheme();
    if (!"http".equalsIgnoreCase(scheme) && !"https".equalsIgnoreCase(scheme)) {
        throw new IOException("Unsupported redirect protocol: " + scheme);
    }
    // Strip credentials on cross-origin redirect or protocol downgrade
    String origAuthority = con.getURL().getAuthority();
    boolean crossOrigin = !redirectUri.getAuthority().equals(origAuthority);
    boolean downgrade = "https".equalsIgnoreCase(con.getURL().getProtocol())
            && "http".equalsIgnoreCase(scheme);
    con.disconnect();
    target.add(0, redirectUri);
    return perform(method, target,
            (crossOrigin || downgrade) ? null : currAuth,
            currProxyAuth, task);
}
```

Similarly, add `con.disconnect()` before each auth-retry recursive call.

gnodet (commented):
**Follow-up review #2 — commits `0898a9e5`, `8e716484`, `79cb996a`**

All four findings from the initial review have been addressed:

✅ **Connection resource leak**: `con.disconnect()` is now called before every recursive `perform()` call (redirect, auth retry, proxy auth retry) and before throwing exceptions on error paths. `implPeek` uses try/finally for cleanup, and `implGet` disconnects on non-200. Clean.

✅ **Cross-origin credential forwarding**: Authority comparison (`currentAuthority.equalsIgnoreCase(redirectAuthority)`) correctly strips `currAuth` on cross-origin redirects. Keeping `currProxyAuth` intact is correct — proxy auth is for the proxy, not the target.

✅ **HTTPS→HTTP downgrade prevention**: Explicit check before the general scheme validation. Clear error message.

✅ **307/308 redirects**: Named constants (`HTTP_STATUS_TEMPORARY_REDIRECT`, `HTTP_STATUS_PERMANENT_REDIRECT`) and added to the redirect condition. Good.

The status code reshuffling (response code check first in the `else if` conditions) improves readability.

Looks good to me. 👍

gnodet (commented):
**Follow-up review #3 — commits `b754bdf9`, `744e55af`, `eabbb7af` (configurable redirects & cleanup)**

Excellent evolution — you've taken the security feedback and built a proper configurable redirect system. This is the right approach for a library: default to the secure option while giving users control.

### What's new

✅ **`RedirectMode` enum** (NONE / SAME_AUTHORITY / ANY) — users can choose their security/convenience tradeoff. Default `ANY` with auth stripping on cross-origin is a pragmatic default.

✅ **`redirectAllowDowngrade`** — HTTPS→HTTP downgrade is now configurable (default: `false` = secure). Users who need it can opt in per repository.

✅ **`maxRedirects`** — configurable (default: 5). The `target.size() - 1` accounting is correct.

✅ **Per-repository configuration** via `.<repoId>` suffix — standard Maven Resolver pattern, nicely documented.

✅ **`implGet` refactored** — entire body wrapped in try/finally with `con.disconnect()`. Much cleaner than the previous per-path disconnect calls. `implPeek` similarly improved.

✅ **303 See Other** added to redirect codes — correct for a read-only (GET/HEAD) transport.

✅ **`UrlTransporterConfigurationKeys`** — clean separation of config constants with proper Javadoc annotations.

### Minor observation (not blocking)

`RedirectMode.valueOf(...)` will throw `IllegalArgumentException` if a user misconfigures the value (e.g. typo like `"ANYY"`). This is reasonable fail-fast behavior, consistent with how Maven Resolver handles other configs. Could optionally log a warning and fall back to the default, but not necessary.

Looks good to me. 👍

gnodet (commented):
**Follow-up review #4** — reviewing commit `4c5d834` ("Applied PR reviews")

Clean polish pass. All changes look correct:

- ✅ `UnsupportedOperationException` in `implPut` — matches the `ClasspathTransporter` precedent for read-only transports. Removing `throws Exception` from the override is valid (narrowing exception signature).
- ✅ Defensive `IllegalArgumentException` catch around `URI.create()` / `resolve()` in redirect handling — addresses a real gap where a malformed `Location` header from a server could cause an unhandled crash. Connection disconnect and cause chain are properly preserved.
- ✅ `final` declarations on redirect variables — idiomatic and enforce single-assignment.
- ✅ Better error message including `redirectAuthority` — helpful for debugging, no security concern.
- ✅ Documentation expansion accurately reflects the transport's capabilities.

One very minor nit: `transporter-known-issues.md` is missing a trailing newline (the `\ No newline at end of file` marker in the diff), while most other markdown files in `src/site/markdown/` end with one. Not a blocker at all.

gnodet (commented):
**Follow-up review #5** — commit `1ee10959` ("Add NL to file end")

✅ Trailing newline added to `transporter-known-issues.md` — addresses the nit from review #4. Thanks for the quick fix!

gnodet (commented):
**Follow-up review #6** — commits `812ad23f` (merge from master) + `a771559c` ("Adopt to latest test changes")

✅ Test adaptation after merging master. The `UrlTransporterTest` now overrides `supportsHttp3()` and `supportsHttp2()` to return `false` — correct, since URL transport uses `HttpURLConnection` which only supports HTTP/1.1. Clean merge, no conflicts.

gnodet (commented):
**Follow-up review #7** — commits 18–19 (test adaptations + merge from upstream/master)

**Changes reviewed:**
- **Commit 18** ("Fixes"): Guards SSL property assertions (`SSL_PROTOCOL`, `SSL_CIPHER_SUITE`) in `testPeek_SSL()` and `testGet_SSL()` behind `exposeContentCodingInTransportProperties()`. `UrlTransporterTest` overrides this to return `false` — correct, since `HttpURLConnection` doesn't expose SSL/TLS metadata.
- **Commit 19** (merge): Cleanly incorporates the merged PR #1967 (HTTP/3 name-based comparison in `JdkTransporter.toHttpVersion()`).

**Observations:**
- The `exposeContentCodingInTransportProperties()` method was originally introduced (in PR #1762) for content-coding assertions, so reusing it for SSL property guards is a slight semantic stretch. If a future transporter exposes SSL properties without content-coding, a separate guard would be needed. This is non-blocking — the method lives on master, not in this PR's scope.

**Verdict:** ✅ Looks good — test-only changes, consistent with the existing override pattern, all 18 CI matrix jobs pass.

---
_Reviewed with AI on behalf of @gnodet. This review was generated by an AI agent and may contain inaccuracies; please verify all suggestions before applying._

PR #1949, opened 2026-07-06
Support HTTP/3 in Jetty and JRE HTTP Client

kwin:
Consolidate http version configuration among HTTP transporters Default to HTTP/2 in all transporters except Apache HTTP Client 4.x

Following this checklist to help us incorporate your
contribution quickly and easily:

- [ ] Your pull request should address just one issue, without pulling in other changes.
- [ ] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [ ] Each commit in the pull request should have a meaningful subject line and body.
  Note that commits might be squashed by a maintainer on merge.
- [ ] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied.
  This may not always be possible but is a best-practice.
- [ ] Run `mvn verify` to make sure basic checks pass.
  A more thorough check will be performed on your pull request automatically.
- [ ] You have run the integration tests successfully (`mvn -Prun-its verify`).

If your pull request is about ~20 lines of code you don't need to sign an
[Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf) if you are unsure
please ask on the developers list.

To make clear that you license your contribution under
the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
you have to acknowledge this by using the following check-box.

- [ ] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
- [ ] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

PR #1967, opened 2026-07-16
Fix reporting of HTTP/3 in JdkTransporter

kwin:
Use enum name, as HttpVersion.HTTP_3 was only added in Java26+

### DIFF
--- a/maven-resolver-transport-jdk-parent/maven-resolver-transport-jdk11/src/main/java/org/eclipse/aether/transport/jdk/JdkTransporter.java
+++ b/maven-resolver-transport-jdk-parent/maven-resolver-transport-jdk11/src/main/java/org/eclipse/aether/transport/jdk/JdkTransporter.java
@@ -388,7 +388,12 @@ final class JdkTransporter extends AbstractTransporter implements HttpTransporte
             case HTTP_2:
                 return HttpTransportProperty.HttpVersion.HTTP_2;
             default:
-                throw new IllegalArgumentException("Unsupported HTTP version: " + version);
+                // support HTTP_3 via name (as only part of Java 26+ API)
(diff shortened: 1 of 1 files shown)


cstamas (approved):

gnodet (commented):
**Review of PR #1967** — Fix reporting of HTTP/3 in JdkTransporter

Clean, minimal fix for an oversight in the recently merged PR #1949. Looks correct.

**Analysis:**

- ✅ **Correctness**: `toHttpVersion()` converts `java.net.http.HttpClient.Version` → resolver's `HttpTransportProperty.HttpVersion`. On Java 26+, `HTTP_3` exists in `HttpClient.Version` but can't appear in a `switch` compiled against Java 11. The `version.name()` comparison is the standard approach for this cross-compilation scenario, and already used by `resolveHttpVersion()` elsewhere in this same file.

- ✅ **Root cause**: PR #1949 added `getHttpVersion`, `resolveHttpVersion`, and `getMaximumSupportedHttpVersion` to handle HTTP/3 on the *request* side but didn't update `toHttpVersion` on the *response* side. If a JDK 26+ runtime negotiated HTTP/3, `toHttpVersion` would crash with `IllegalArgumentException` when reporting the protocol version in transport properties.

- ✅ **Backward compatibility**: On Java 11-25, `HttpClient.Version` has no `HTTP_3` value, so the name check simply never matches. Any truly unknown future value still throws `IllegalArgumentException`.

- ✅ **No test included**: Understandable — testing requires a Java 26+ runtime where `HttpClient.Version.HTTP_3` exists.

One very minor style observation: the `else` wrapping the `throw` is technically unnecessary since the `if` branch returns, but the code is clear as-is.

PR #1970, opened 2026-07-16
Make sure to always close input streams bound to responses

### LINKED ISSUE #1964: JettyTransporter and JdkTransporter do not consume response body of PUT potentially leading to dangling connections or deadlocks
### Affected version

2.0.20

### Bug description

Not consuming the response body can lead to stale http connections and in the end to deadlocks.
The response body is not consumed in https://github.com/apache/maven-resolver/blob/79d102b66235f33ad1e6134e18451ac3ee91b44a/maven-resolver-transport-jetty/src/main/java/org/eclipse/aether/transport/jetty/JettyTransporter.java#L309 and not in https://github.com/apache/maven-resolver/blob/79d102b66235f33ad1e6134e18451ac3ee91b44a/maven-resolver-transport-jdk-parent/maven-resolver-transport-jdk11/src/main/java/org/eclipse/aether/transport/jdk/JdkTransporter.java#L409 (when there is a status code < 300).

Although Maven does not expect a response from a repository PUT request, we should make sure if a response body is sent this is consumed to not lead to dangling connections.

Just closing the underlying input stream is enough for both transporters.

kwin:
Ensure all connections are closed when transporter is closed Add PUT IT where response contains body

This closes #1964

### DIFF
--- a/maven-resolver-test-http/pom.xml
+++ b/maven-resolver-test-http/pom.xml
@@ -134,7 +134,6 @@
     <dependency>
       <groupId>org.awaitility</groupId>
       <artifactId>awaitility</artifactId>
-      <version>4.3.0</version>
     </dependency>
   </dependencies>
 
--- a/maven-resolver-test-http/pom.xml
+++ b/maven-resolver-test-http/pom.xml
@@ -131,6 +131,11 @@
       <groupId>com.google.code.gson</groupId>
       <artifactId>gson</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.awaitility</groupId>
+      <artifactId>awaitility</artifactId>
+      <version>4.3.0</version>
+    </dependency>
   </dependencies>
 
 </project>
--- a/maven-resolver-test-http/src/main/java/org/eclipse/aether/internal/test/util/http/HttpServer.java
+++ b/maven-resolver-test-http/src/main/java/org/eclipse/aether/internal/test/util/http/HttpServer.java
@@ -27,6 +27,7 @@ import java.nio.file.Files;
 import java.nio.file.StandardOpenOption;
 import java.util.ArrayList;
 import java.util.Base64;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
@@ -60,6 +61,8 @@ import org.eclipse.jetty.http3.server.HTTP3ServerConnectionFactory;
 import org.eclipse.jetty.http3.server.HTTP3ServerQuicConfiguration;
 import org.eclipse.jetty.io.ByteBufferPool;
 import org.eclipse.jetty.io.Content;
+import org.eclipse.jetty.io.DatagramChannelEndPoint;
+import org.eclipse.jetty.io.EndPoint;
 import org.eclipse.jetty.quic.quiche.server.QuicheServerConnector;
 import org.eclipse.jetty.quic.quiche.server.QuicheServerQuicConfiguration;
 import org.eclipse.jetty.server.Handler;
@@ -188,6 +191,8 @@ public class HttpServer {
     private int serverErrorStatusCode;
 
     private final List<LogEntry> logEntries = Collections.synchronizedList(new ArrayList<>());
+
+    private String responseBodyForPut;
 
     public String getHost() {
         return "localhost";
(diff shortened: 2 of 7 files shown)


kwin:
Some investigations ongoing on how to properly test for dangling connections: https://github.com/jetty/jetty.project/issues/15436.

gnodet (commented):
**Verdict: ✅ Approve** — excellent fix for a real connection leak that could cause deadlocks in production.

## What's great

The production fixes are correct and well-targeted:
- **JdkTransporter `implPut()`**: adds the missing `finally { closeBody(response); }` — this was the core bug. `closeBody` is already null-safe (checks `streamHttpResponse != null`), so the `response = null` initialization is safe.
- **JdkTransporter `implGet()`**: clean simplification — removes redundant inner `closeBody()` calls since the outer `finally` already handles all exit paths.
- **JettyTransporter `implPut()`**: wrapping `InputStreamResponseListener` in try-with-resources and moving RFC 9457 handling inside ensures the listener is always closed. The `HttpTransporterException` still propagates correctly (it's not caught by the `ExecutionException` catch block).
- **Connection leak detection in `tearDown()`** is a force multiplier — every existing HTTP transporter test now doubles as a leak detector.

## Suggestions

**1. Null guard before Awaitility check** (`HttpTransporterTest.java`, line ~309)

The Awaitility assertion runs *before* the `if (httpServer != null)` check. If `setUp()` fails before creating the server, `@AfterEach` still runs and the lambda will NPE, masking the real failure:

```java
// Current:
Awaitility.await().atMost(5, TimeUnit.SECONDS).until(() -> httpServer.getNumConnectedEndPoints() == 0);
if (httpServer != null) {

// Suggested — move inside the null guard:
if (httpServer != null) {
    Awaitility.await().atMost(5, TimeUnit.SECONDS).until(() -> httpServer.getNumConnectedEndPoints() == 0);
    httpServer.stop();
    httpServer = null;
}
```

**2. HTTP 204 with response body — may weaken the test** (`HttpServer.java`, PUT handler)

The response body is written, then status is set to `SC_NO_CONTENT` (204). Per RFC 9110 §6.4.1, 204 responses should have no content. While Jetty will still put the bytes on the wire, some HTTP client implementations may skip draining the body for 204, which could make the test pass even *without* the fix. Consider using `SC_OK` (200) when `responseBodyForPut` is set so the test is unambiguous:

```java
if (responseBodyForPut != null) {
    writeResponseBodyMessage(req, response, responseBodyForPut);
    response.setStatus(HttpServletResponse.SC_OK);
} else {
    response.setStatus(HttpServletResponse.SC_NO_CONTENT);
}
```

## Minor observations (no action needed)

- **Double closeBody in `implPut`**: The inner `try { ... } finally { closeBody(response); }` around error handling is now redundant with the new outer `finally`. Harmless (idempotent close), but in `implGet` the equivalent was cleaned up — consistency could be improved.
- **Hardcoded Awaitility version**: `4.3.0` is hardcoded in `maven-resolver-test-http/pom.xml` rather than managed in the parent POM's `<dependencyManagement>`.
- **Pre-existing**: `JettyTransporter.implGet()` also creates `InputStreamResponseListener` without try-with-resources — the same pattern this PR fixes in `implPut()`. Might be worth a follow-up.

CI: ⏳ All pending (just un-drafted). Production code changes are straightforward enough to review without CI, but will want to confirm all tests pass.

---
_Reviewed with AI assistance on behalf of @gnodet. This review was generated by an AI agent and may contain inaccuracies; please verify all suggestions before applying._

cstamas (approved):

gnodet (commented):
**Follow-up review (commit 2/2):** ✅ All suggestions addressed — looks great.

- **Null guard**: Awaitility check correctly moved inside `if (httpServer != null)` ✓
- **PUT response status**: Using `SC_CREATED` (201) instead of `SC_NO_CONTENT` (204) when a body is present — even better than the 200 I suggested, since 201 is the semantically correct success status for PUT-creates-resource ✓
- **Awaitility version**: Centralized in parent POM `<dependencyManagement>`, with a helpful note about the 4.3.1 release issue ✓
- **Double closeBody in `implPut`**: Inner try/finally removed, now consistent with `implGet` simplification ✓

No further comments.

---
_Reviewed with AI assistance on behalf of @gnodet. This review was generated by an AI agent and may contain inaccuracies; please verify all suggestions before applying._

PR #1971, opened 2026-07-16
Fix: goal is to fix master

cstamas:
Do not use Enum in default values, as it broke the build on master. Instead, convert it to enum name (as all enum were handled so far). Once the enum fix lands on master, this can be undone.

Configuration is never "hot", so the back-and-forth conversion, while kinda ugly, is really not a penalty here.

### DIFF
--- a/maven-resolver-api/src/main/java/org/eclipse/aether/ConfigurationProperties.java
+++ b/maven-resolver-api/src/main/java/org/eclipse/aether/ConfigurationProperties.java
@@ -546,7 +546,7 @@ public final class ConfigurationProperties {
      *
      * @since 2.0.21
      */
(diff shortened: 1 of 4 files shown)


kwin:
This is not about the conversion but about the documentation of allowed values. With enum this is much cleaner. String enforces to list all allowed values explicitly.

kwin:
I just committed https://github.com/apache/maven-resolver/commit/46da42c5f285f9561395b9bf38e077e95b9a9766 which should bring back master to green.

cstamas:
This PR was not about documentation, but to make SNAPSHOT deployed by CI a reality, to become able to pick it up by Maven builds.

BTW, now, when Maven ingests Resolver 2.0.21-SNAPSHOT, how will the maven documentation generation work?

PR #1973, opened 2026-07-17
Fix: be more defensive regarding request traces

cstamas:
As they NPE in UTs but would also NPE in calls made by some non-maven callers that usually do not set trace at all.

### DIFF
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultRepositorySystem.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultRepositorySystem.java
@@ -277,8 +277,11 @@ public class DefaultRepositorySystem implements RepositorySystem {
             throws ArtifactResolutionException {
         requireNonNull(requests, "requests cannot be null");
         // All requests in a batch share the same trace context, so checking any one is sufficient.
-        RequestTrace firstTrace =
-                requests.stream().map(ArtifactRequest::getTrace).findFirst().orElse(null);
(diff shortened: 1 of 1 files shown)


gnodet (commented):
**Verdict: ✅ Approve** — clean, correct fix for a real NPE.

**What this fixes:** `Stream.findFirst()` throws `NullPointerException` when the first mapped element is `null` (it internally calls `Optional.of(value)` which rejects null). When a non-Maven caller doesn't set a trace on its requests, `getTrace()` returns `null` and the stream crashes before `orElse(null)` is ever reached.

**What I checked:**
- `isReentrant(RequestTrace)` is already null-safe — the for-loop condition `t != null` exits immediately, returning `false` (i.e., "not re-entrant" → validation runs). So the downstream behavior for null traces was always correct; only the stream extraction was broken.
- The `Objects` import was already present — no structural changes needed.
- Both `resolveArtifacts()` and `resolveMetadata()` had the same pattern and both are fixed symmetrically.
- Single-request overloads (`resolveArtifact`, `collectDependencies`, etc.) call `isReentrant(request.getTrace())` directly — no stream, no issue.

**Minor observation (no action needed):** The comment "All requests in a batch share the same trace context, so checking any one is sufficient" is slightly imprecise now that the code specifically seeks the first *non-null* trace, but this is purely cosmetic.

CI: ✅ 18/18 jobs green.

---
_Reviewed with AI assistance on behalf of @gnodet. This review was generated by an AI agent and may contain inaccuracies; please verify all suggestions before applying._

PR #1975, opened 2026-07-17
Bug: in certain cases Resolver caused build failure

cstamas:
While it should not.

The result repository MAY be `null` in certain cases, and that resulted in omission of recording ArtifactNotFoundEx.

While not biggie, later on, if this resolution was in fact happening for a POM, despite descriptor policy would be "ignore missing" (default in Maven 3 and 4), due not recorded ANFex, this error would not be recognized as "missing POM", and would fail the build or artifact collection or anything involving reading descriptor.

### DIFF
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultArtifactResolver.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultArtifactResolver.java
@@ -414,7 +414,14 @@ public class DefaultArtifactResolver implements ArtifactResolver {
                         if (result.getExceptions().isEmpty()) {
                             Exception exception =
                                     new ArtifactNotFoundException(request.getArtifact(), (RemoteRepository) null);
-                            result.addException(result.getRepository(), exception);
+                            // Note: result.getRepository() MAY BE null; in cases when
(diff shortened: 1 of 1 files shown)


gnodet (commented):
**Review: Bug fix for silently dropped `ArtifactNotFoundException`**

Clean, minimal, and correct. The root cause is well-identified: `addException(ArtifactRepository, Exception)` has a `repository != null` guard that silently discards the exception when `result.getRepository()` is null. This happens when no remote repository even attempted the artifact (e.g. snapshot vs release policy mismatch) and the local repository doesn't have it either.

Using the ternary with `ArtifactResult.NO_REPOSITORY` as fallback is better than the deprecated one-arg `addException(Exception)` because it preserves the actual repository when available. The sentinel `NO_REPOSITORY` exists precisely for this case and is already used at line 249 in the same file.

**One nit:** the inline comment has a double negative — "no artifact was **not** even tried" reads as the opposite of the intended meaning. Consider:
```java
// Note: result.getRepository() MAY BE null; in cases when
// the artifact was not even tried by any remote repository (snapshot vs repo policy)
// and local repository does not have it either
```

No blocking concerns. CI still running (10/20 pass so far, all green).

---
_Reviewed with `oss-code-reviewer` on behalf of @gnodet. This review was generated by an AI agent and may contain inaccuracies; please verify all suggestions before applying._

PR #1977, opened 2026-07-18
Do not check UrlTransporter for open connections after close

kwin:
Connection pooling for HttpUrlTransporter is managed by JRE with no way to explicitly close idle connections
(compare with
https://docs.oracle.com/javase/6/docs/technotes/guides/net/http-keepalive.html)

### DIFF
--- a/maven-resolver-test-http/src/main/java/org/eclipse/aether/internal/test/util/http/HttpTransporterTest.java
+++ b/maven-resolver-test-http/src/main/java/org/eclipse/aether/internal/test/util/http/HttpTransporterTest.java
@@ -307,13 +307,23 @@ public abstract class HttpTransporterTest {
             closer = null;
         }
         if (httpServer != null) {
-            // check for leaked connections (e.g., due to not closing response body streams)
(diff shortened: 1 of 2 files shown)


cstamas (approved):

gnodet (commented):
Clean fix. The template method pattern (`closesAllConnectionsOnTransporterClose()`) follows the existing hooks in this base class (`supportsPreemptiveAuth()`, `supportsHttp3()`, etc.) — immediately familiar and extensible for any future transport that also can't control its connection pool.

The `testPut_WithResponseBody` disable is correct — URL transport doesn't support PUT, consistent with all other PUT test disables in `UrlTransporterTest`.

**One nit**: the Awaitility timeout was reduced from 5s to 3s alongside the skip-logic change. CI is green across all 18 matrix cells, so the practical risk is low, but the reduction is undocumented. If a slow CI agent causes flakiness later, this is the line to investigate. The 5s original was presumably a generous bound.

ascheman:
FWIW: #1978 addresses the same failure while keeping the leak assertion active for the URL transporter (by actually closing connections in the test via `Connection: close`), retaining the coverage #1970 introduced. Might be worth consolidating on that approach — the `testPut_WithResponseBody` `@Disabled` part is common to both anyway.

PR #1978, opened 2026-07-18
Feat: config to close connection at end of tx

cstamas:
New config option to make URL transport close connection at end of each transaction.

### DIFF
--- a/maven-resolver-transport-url/src/main/java/org/eclipse/aether/transport/url/UrlTransporterConfigurationKeys.java
+++ b/maven-resolver-transport-url/src/main/java/org/eclipse/aether/transport/url/UrlTransporterConfigurationKeys.java
@@ -71,13 +71,19 @@ public final class UrlTransporterConfigurationKeys {
     public static final boolean DEFAULT_REDIRECT_ALLOW_DOWNGRADE = false;
 
     /**
-     * Whether persistent connections should be used by this transport. This configuration allows controlling only this transport,
-     * while the default value (that is JVM-wide) comes from {@code http.keepAlive} Java system property, that is {@code true}
-     * by default.
+     * Whether persistent connections should be used by this transport. This configuration affects only this transport,
+     * whether it should close connection at the end of each transaction or not. Important to note, that globally, on
+     * JVM-wide level this is controlled by
+     * <a href="https://docs.oracle.com/javase/8/docs/technotes/guides/net/http-keepalive.html">{@code http.keepAlive}</a>
+     * Java system property, that is {@code true} by default. If this system property is set to {@code false}, this
+     * transport cannot use persistent connections either.
      *
      * @configurationSource {@link RepositorySystemSession#getConfigProperties()}
      * @configurationType {@link java.lang.Boolean}
+     * @configurationDefaultValue {@link #DEFAULT_CLOSE_CONNECTION}
      * @configurationRepoIdSuffix Yes
      */
-    public static final String CONFIG_PROP_KEEP_ALIVE = CONFIG_PROPS_PREFIX + "keepAlive";
+    public static final String CONFIG_PROP_CLOSE_CONNECTION = CONFIG_PROPS_PREFIX + "closeConnection";
+
+    public static final boolean DEFAULT_CLOSE_CONNECTION = false;
 }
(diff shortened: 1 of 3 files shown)


maven-resolver-transport-url/src/main/java/org/eclipse/aether/transport/url/UrlTransporterConfigurationKeys.java
```diff
@@ -69,4 +69,15 @@ private UrlTransporterConfigurationKeys() {}
...
+    /**
+     * Whether persistent connections should be used by this transport. This configuration allows controlling only this transport,
+     * while the default value (that is JVM-wide) comes from {@code http.keepAlive} Java system property, that is {@code true}
+     * by default.
+     *
+     * @configurationSource {@link RepositorySystemSession#getConfigProperties()}
+     * @configurationType {@link java.lang.Boolean}
+     * @configurationRepoIdSuffix Yes
+     */
+    public static final String CONFIG_PROP_KEEP_ALIVE = CONFIG_PROPS_PREFIX + "keepAlive";
```
kwin: Not sure this option should be exposed. I don't really see a need for ever setting this flag.

cstamas: I would argue about this. I saw all kind of broken MRMs in my life, but this was still an option to users, as we really do not check what kind of HTTP headers are they configuring, so they can easily add this header as well.

IMO, the fact this config would exist, would not mean no issue, and even may be handy in some cases.

cstamas:
Will keep this config, and remove the UT hack so both can be now merged.

gnodet (commented):
Good approach — adding a per-transport `closeConnection` config gives users a targeted escape hatch without the JVM-wide blast radius of `-Dhttp.keepAlive=false`. The implementation follows the existing config pattern precisely (`ConfigUtils.getBoolean`, repo-ID suffix, `CONFIG_PROPS_PREFIX`). Thread safety is airtight (`private final`, constructor-only).

**One suggestion** — the javadoc opening sentence has inverted polarity relative to the property name:

> "Whether persistent connections should be used by this transport."

With `closeConnection` defaulting to `false`, a reader would interpret this as "persistent connections should NOT be used by default" — the opposite of actual behavior. The rest of the javadoc is accurate. Suggested rewording:

```java
/**
 * Whether this transport should close the HTTP connection at the end of each transaction, effectively
 * disabling persistent connections (keep-alive). This configuration affects only this transport.
 * Important to note, that globally, on JVM-wide level persistent connections are controlled by
 * ...
 */
```

This aligns the javadoc polarity with the property name: "Whether ... should close the connection" → `closeConnection = true/false`.

gnodet (commented):
**Follow-up review (commits 3–4)**

The javadoc on `CONFIG_PROP_CLOSE_CONNECTION` now accurately describes the property:

> *"Whether this transport should close the HTTP connection at the end of each transaction, effectively disabling persistent connections (keep-alive)."*

This resolves the polarity issue flagged in the prior review — the description aligns correctly with the property name, default (`false` = keep-alive), and behavior. The added context about the JVM-wide `http.keepAlive` system property is a nice touch.

The UT hack removal (commit 3) is clean — with #1977 merged, the `UrlTransporterTest` connection leak check is properly skipped via the template method, so no test-side workaround is needed here.

CI: 19/20 matrix jobs pass; the one failure (`macos-latest jdk-26-zulu 4.0.0-rc-5`) is a transient Maven Central network timeout downloading `io.grpc:grpc-context:1.78.0` — unrelated to this PR.

No remaining concerns — looks good to merge. 👍

---
_Reviewed with `oss-code-reviewer` on behalf of @gnodet. This review was generated by an AI agent and may contain inaccuracies; please verify all suggestions before applying._

kwin:
@cstamas Shouldn't this rather be a transporter agnostic config then? One can already use this via `aether.connector.http.headers` though.

cstamas:
IMO this config is specific to transports that does not (or cannot) control their pool, and so far it is only the URL transport.

cstamas:
IMO this config is specific to transports that does not (or cannot) control their pool, and so far it is only the URL transport.

PR #1981, opened 2026-07-20
Improve rendering of javadoc content

kwin:
Render all inline tags in the same way as the standard doclet. Some code simplifications.

### DIFF
--- a/maven-resolver-tools/src/main/java/org/eclipse/aether/tools/ConfigurationCollectorDoclet.java
+++ b/maven-resolver-tools/src/main/java/org/eclipse/aether/tools/ConfigurationCollectorDoclet.java
@@ -41,24 +41,30 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Properties;
 import java.util.Set;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
+import com.sun.source.doctree.AttributeTree;
 import com.sun.source.doctree.DocCommentTree;
 import com.sun.source.doctree.DocTree;
+import com.sun.source.doctree.EndElementTree;
+import com.sun.source.doctree.EntityTree;
 import com.sun.source.doctree.LinkTree;
 import com.sun.source.doctree.LiteralTree;
 import com.sun.source.doctree.SinceTree;
+import com.sun.source.doctree.StartElementTree;
+import com.sun.source.doctree.SystemPropertyTree;
 import com.sun.source.doctree.TextTree;
 import com.sun.source.doctree.UnknownBlockTagTree;
+import com.sun.source.doctree.ValueTree;
 import com.sun.source.tree.ExpressionTree;
 import com.sun.source.tree.IdentifierTree;
 import com.sun.source.tree.MemberSelectTree;
 import com.sun.source.tree.VariableTree;
 import com.sun.source.util.DocTreePath;
 import com.sun.source.util.DocTrees;
+import com.sun.source.util.SimpleDocTreeVisitor;
 import jdk.javadoc.doclet.Doclet;
 import jdk.javadoc.doclet.DocletEnvironment;
 import jdk.javadoc.doclet.Reporter;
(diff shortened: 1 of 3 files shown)


gnodet (commented):
Good improvement overall — replacing the fragile regex-based Javadoc tag post-processing with a proper `SimpleDocTreeVisitor` is the right approach. The handling of `{@value}`, `{@systemProperty}`, HTML elements, and entities is well done.

**Bug — `{@code}` / `{@literal}` rendering swapped:**

In `visitLiteral` (line 502), the condition `node.getKind() == DocTree.Kind.LITERAL` routes `{@literal}` through `renderAsCode()` and `{@code}` through `escapeHtml()`. Per the Javadoc spec:
- `{@code text}` (Kind.CODE) should render as `<code>text</code>`
- `{@literal text}` (Kind.LITERAL) should render as plain escaped text

The condition should check `DocTree.Kind.CODE` instead. The sibling `visitLink` method in the same file demonstrates the correct pattern: `LINK_PLAIN` → `escapeHtml()`, `LINK` → `renderAsCode()`.

The test at line 84 validates the swapped behavior — the expected string shows plain text for `{@code}` and `<code>` tags for `{@literal}`, which matches the buggy code but contradicts the spec.

Suggested fix:
```java
if (node.getKind() == DocTree.Kind.CODE) {
    return renderAsCode(node.getBody().getBody());
} else {
    return escapeHtml(node.getBody().getBody());
}
```

**Minor:** `renderContent` Javadoc has `@param` tags without descriptions.

gnodet (commented):
The refactoring from regex-based `cleanseTags`/`cleanseJavadoc` post-processing to a proper `SimpleDocTreeVisitor` is a great improvement — much more robust. The added support for `{@value}`, `{@systemProperty}`, HTML elements, and entities is welcome, and the cleanup of `nvl` to `Objects.toString` is a nice simplification.

However, `visitLiteral` has the `{@code}` and `{@literal}` handling swapped. Details below.

maven-resolver-tools/src/main/java/org/eclipse/aether/tools/ConfigurationCollectorDoclet.java
```diff
@@ -457,121 +467,172 @@ private String resolveReferencedType(
...
+            public String visitLink(LinkTree node, Void p) {
+                String ref = node.getReference() != null ? node.getReference().getSignature() : "";
+                String label = renderContent(node.getLabel(), null, null, false);
+                String text = label == null || label.isEmpty() ? ref : label;
+                return node.getKind() == DocTree.Kind.LINK_PLAIN ? escapeHtml(text) : renderAsCode(text);
+            }
+
+            @Override
+            public String visitLiteral(LiteralTree node, Void p) {
+                if (node.getKind() == DocTree.Kind.LITERAL) {
```
gnodet: The condition here is inverted. `DocTree.Kind.CODE` corresponds to the `{@code}` tag (monospace, wrapped in `<code>`) while `DocTree.Kind.LITERAL` corresponds to the `{@literal}` tag (plain text, HTML-escaped, no monospace). The current code wraps `{@literal}` content in `<code>` and leaves `{@code}` content unwrapped — the opposite of what the standard doclet does.

Notably, `visitLink` (line 497) gets this right: it maps `LINK` (code-font semantics) to `renderAsCode` and `LINK_PLAIN` (plain-text semantics) to `escapeHtml`. The same pattern should apply here.

```suggestion
            public String visitLiteral(LiteralTree node, Void p) {
                if (node.getKind() == DocTree.Kind.CODE) {
                    return renderAsCode(node.getBody().getBody());
                } else {
                    return escapeHtml(node.getBody().getBody());
                }
            }
```

maven-resolver-tools/src/test/java/org/eclipse/aether/tools/ConfigurationCollectorDocletTest.java
```diff
@@ -78,8 +78,11 @@ void extractsBooleanStringAndEnumConfigurations(@TempDir Path tempDir) throws Ex
         assertNotNull(string, "string key missing");
         assertEquals("String", string.get("configurationType"));
         assertEquals("\"hello\"", string.get("defaultValue"));
-        assertEquals("Yes", string.get("supportRepoIdSuffix"));
         assertEquals("", string.get("since"), "no @since expected");
+        assertEquals("Yes", string.get("supportRepoIdSuffix"));
+        assertEquals(
+                "A string value with some inline tags. Value <code>\"hello\"</code> is the default. <code>some.property</code> is used. This text is code. <code>This text is literal.</code> <code>java.lang.String</code> is the type.",
```
gnodet: The expected string here reflects the inverted behavior — it expects `{@code This text is code.}` to render without a `<code>` wrapper and `{@literal This text is literal.}` to render with one. Once the condition in `visitLiteral` is fixed, this expectation should be updated to match:

```suggestion
                "A string value with some inline tags. Value <code>\"hello\"</code> is the default. <code>some.property</code> is used. <code>This text is code.</code> This text is literal. <code>java.lang.String</code> is the type.",
```

gnodet (approved):
Solid refactoring that replaces fragile regex-based Javadoc tag post-processing with a proper `SimpleDocTreeVisitor`. The `{@code}`/`{@literal}` bug flagged by prior reviews was correctly fixed in commit `0c4adcb4`.

**Minor observations (non-blocking):**
- `escapeHtml` escapes `&`, `<`, `>` but not `"` — if an attribute value ever contained the same quote character used as delimiter, the output HTML would be malformed. Extremely unlikely for Javadoc content.
- `visitValue` for bare `{@value}` (no reference) produces `<code></code>` instead of the enclosing field's constant value per spec. Pre-existing limitation, not a regression.
- Whitespace normalization (`replaceAll("\\s+", " ")`) in the `trim=true` path collapses whitespace inside `<code>` spans. Documented limitation.

**Strengths:**
- The old `cleanseTags` regex `\{@\w\w\w\w (.+?)\}` could only match 4-letter tag names, silently leaving `{@literal}`, `{@value}`, `{@systemProperty}` unprocessed. The visitor approach handles all standard inline tags.
- The `cleanseJavadoc` line-stripping was redundant since `getFullBody()` already excludes block tags.
- `nvl` → `Objects.toString` migration uses standard library.
- Test coverage exercises `{@value}`, `{@systemProperty}`, `{@code}`, `{@literal}`, and `{@link}` inline tags.

PR #1983, opened 2026-07-20
Improve error reporting

kwin:
Add unit tests to check error reporting

### DIFF
--- a/maven-resolver-tools/src/main/java/org/eclipse/aether/tools/ConfigurationCollectorDoclet.java
+++ b/maven-resolver-tools/src/main/java/org/eclipse/aether/tools/ConfigurationCollectorDoclet.java
@@ -217,11 +217,6 @@ public class ConfigurationCollectorDoclet implements Doclet {
                 getConfigurationType(extractClassLink(field, docComment, blockTags, "configurationType"));
         String defValue = resolveDefaultValue(type, field, docComment, blockTags.get("configurationDefaultValue"));
 
-        if (defValue == null) {
-            // Error was already reported, skip this field
-            return;
-        }
-
         UnknownBlockTagTree sourceTag = blockTags.get("configurationSource");
         UnknownBlockTagTree repoIdTag = blockTags.get("configurationRepoIdSuffix");
 
--- a/maven-resolver-tools/src/main/java/org/eclipse/aether/tools/ConfigurationCollectorDoclet.java
+++ b/maven-resolver-tools/src/main/java/org/eclipse/aether/tools/ConfigurationCollectorDoclet.java
@@ -30,7 +30,6 @@ import javax.lang.model.util.ElementFilter;
 import javax.tools.Diagnostic;
 
 import java.io.IOException;
-import java.io.UncheckedIOException;
 import java.io.Writer;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
@@ -133,15 +132,15 @@ public class ConfigurationCollectorDoclet implements Doclet {
         try {
             return doRun(environment);
         } catch (RuntimeException e) {
-            reporter.print(Diagnostic.Kind.ERROR, "Error running ConfigurationCollectorDoclet: " + e.getMessage());
-            e.printStackTrace(reporter.getStandardWriter());
+            reportError("Error running ConfigurationCollectorDoclet: " + e.getMessage());
             return false;
         }
     }
 
     private boolean doRun(DocletEnvironment environment) {
         if (output == null) {
-            throw new IllegalStateException("Missing required --output option");
+            reportError("Missing required --output option");
+            return false;
         }
         docTrees = environment.getDocTrees();
         List<Map<String, String>> discoveredKeys = new ArrayList<>();
@@ -153,18 +152,55 @@ public class ConfigurationCollectorDoclet implements Doclet {
                     continue;
                 }
                 DocCommentTree docComment = docTrees.getDocCommentTree(field);
-                if ("maven".equals(mode)) {
-                    processMavenField(type, field, docComment, discoveredKeys);
-                } else if ("resolver".equals(mode)) {
-                    processResolverField(type, field, docComment, discoveredKeys);
-                } else {
-                    throw new IllegalArgumentException("Unknown mode: " + mode);
+                try {
+                    if ("maven".equals(mode)) {
+                        processMavenField(type, field, docComment, discoveredKeys);
+                    } else if ("resolver".equals(mode)) {
+                        processResolverField(type, field, docComment, discoveredKeys);
+                    } else {
+                        // TODO: move to beginning of run() and validate mode before processing any types
+                        reportError("Unknown mode: " + mode);
+                        return false;
+                    }
+                } catch (DocTreePathAwareRuntimeException e) {
+                    reportError(e.getDocTreePath(), e.getMessage());
+                } catch (RuntimeException e) {
+                    DocTreePath rootPath = new DocTreePath(docTrees.getPath(field), docComment);
+                    reportError(rootPath, e.getMessage());
                 }
             }
         }
 
-        writeProperties(discoveredKeys);
+        try {
+            writeProperties(discoveredKeys);
+        } catch (IOException e) {
+            reportError("Failed to write properties file: " + e.getMessage());
+            return false;
+        }
         return true;
+    }
+
+    /**
+     * Reports an error message at a specific DocTreePath location.
+     *
+     * @param path the DocTreePath where the error occurred
+     * @param message the error message
+     */
+    private void reportError(DocTreePath path, String message) {
+        if (path != null) {
+            reporter.print(Diagnostic.Kind.ERROR, path, message);
+        } else {
+            reportError(message);
+        }
+    }
+
+    /**
+     * Reports a global error message without location information.
+     *
+     * @param message the error message
+     */
+    private void reportError(String message) {
+        reporter.print(Diagnostic.Kind.ERROR, message);
     }
 
     private void processResolverField(
@@ -172,14 +208,22 @@ public class ConfigurationCollectorDoclet implements Doclet {
         if (docComment == null) {
             return;
         }
-        Map<String, List<? extends DocTree>> blockTags = collectBlockTags(docComment);
+        Map<String, UnknownBlockTagTree> blockTags = collectBlockTags(docComment);
         if (!blockTags.containsKey("configurationSource")) {
             return;
         }
 
         String configurationType =
-                getConfigurationType(extractClassLink(field, docComment, blockTags.get("configurationType")));
+                getConfigurationType(extractClassLink(field, docComment, blockTags, "configurationType"));
         String defValue = resolveDefaultValue(type, field, docComment, blockTags.get("configurationDefaultValue"));
+
+        if (defValue == null) {
+            // Error was already reported, skip this field
+            return;
+        }
+
+        UnknownBlockTagTree sourceTag = blockTags.get("configurationSource");
+        UnknownBlockTagTree repoIdTag = blockTags.get("configurationRepoIdSuffix");
 
         Map<String, String> entry = new LinkedHashMap<>();
         entry.put("key", String.valueOf(field.getConstantValue()));
@@ -187,9 +231,11 @@ public class ConfigurationCollectorDoclet implements Doclet {
         entry.put("fqName", type.getQualifiedName() + "." + field.getSimpleName());
         entry.put("description", cleanseJavadoc(renderContent(docComment.getFullBody())));
         entry.put("since", nvl(getSince(type, docComment), ""));
-        entry.put("configurationSource", getConfigurationSource(renderContent(blockTags.get("configurationSource"))));
+        entry.put(
+                "configurationSource",
+                getConfigurationSource(renderContent(sourceTag != null ? sourceTag.getContent() : null)));
         entry.put("configurationType", configurationType);
-        entry.put("supportRepoIdSuffix", toYesNo(renderContent(blockTags.get("configurationRepoIdSuffix"))));
+        entry.put("supportRepoIdSuffix", toYesNo(renderContent(repoIdTag != null ? repoIdTag.getContent() : null)));
         discovered.add(entry);
     }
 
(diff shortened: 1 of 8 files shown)


gnodet (commented):
Nice work on the error-reporting improvements, @kwin — the continue-on-error pattern with precise `DocTreePath` locations is a big UX improvement for developers debugging their javadoc tags. The `DocTreePathAwareRuntimeException` is a clean solution, and I especially like that you fixed the pre-existing bug where `extractClassLink` hardcoded `"@configurationDefaultValue"` in error messages even when called for the `@configurationType` tag.

One thing caught my attention:

**Behavioral regression: fields without `@configurationDefaultValue` are now silently dropped**

In `processResolverField`, the new null guard:

```java
String defValue = resolveDefaultValue(type, field, docComment, blockTags.get("configurationDefaultValue"));
if (defValue == null) {
    // Error was already reported, skip this field
    return;
}
```

changes behavior for fields that have `@configurationSource` but no `@configurationDefaultValue` tag. On master, `resolveDefaultValue` returned `null` for a missing tag, but the field was still included via `nvl(defValue, "")` producing an empty default value string. After this PR, the field is silently skipped.

This affects real configuration keys in the codebase — a quick check shows at least:

| File | `@configurationSource` | `@configurationDefaultValue` |
|------|:---:|:---:|
| `ClasspathTransporterFactory.java` | 1 | 0 |
| `DefaultOfflineController.java` | 3 | 0 |
| `GnupgConfigurationKeys.java` | 5 | 4 |

The comment "Error was already reported" is also not accurate — when `contentTag` is `null` (tag absent), `resolveDefaultValue` returns `null` silently without reporting anything.

Suggested fix: remove the `if (defValue == null) return;` guard and keep the original `nvl(defValue, "")` handling when building the entry. The `null` return from `resolveDefaultValue` for a missing tag is a valid "no default" case, not an error. The actual error case (unresolvable link) already throws `DocTreePathAwareRuntimeException` and is caught by the per-field catch blocks.

Minor nits:
- There's a stray blank line in `extractClassLink` between the `if` condition and the `throw` statement
- The top-level `run()` catch lost the stack trace (`e.printStackTrace(reporter.getStandardWriter())` was removed) — for truly unexpected failures (NPEs, etc.), the stack trace is valuable for diagnosis

cstamas:
And as can be seen on 2.0.21 release: https://maven.apache.org/resolver-archives/resolver-LATEST/configuration.html

Not all "Source" (last column) values are handled, there are some uninterpolated values left.

image

Unsure is this help or renderer issue, but it should be replaced with "Java System Properties", as currently Resolver config is sourced from two places:
* session config properties (main source)
* Java system properties (early ones, usually during object graph construction, when there is no session yet)

gnodet:
Thanks for the follow-up commit, @kwin — the "Don't skip field without default values" fix directly addresses the behavioral regression I flagged in my review.

Verified: `processResolverField` no longer has the early return on null `defValue`, and instead flows through `nvl(defValue, "")` — matching the master branch behavior. Fields like `ClasspathTransporterFactory.CONFIG_PROP_CLASSPATH_MANAGER`, `DefaultOfflineController`'s three keys, and `GnupgConfigurationKeys.GPGCONF_*` will now be correctly included in the generated properties with empty default values, as before.

The rest of the changes (continue-on-error pattern with `DocTreePathAwareRuntimeException`, precise `DocTreePath` locations, updated tests) look solid. CI is running on the new commit — will check back when it completes.

_This comment was generated by an AI agent and may contain inaccuracies. Please verify all suggestions before applying._

gnodet (approved):
Solid error-reporting improvement. The continue-on-error pattern with precise `DocTreePath` locations is a well-designed improvement over the fail-fast approach. Good test coverage with `CapturingDiagnosticsListener`.

The second commit correctly fixed the behavioral regression where fields without `@configurationDefaultValue` were being silently dropped. CI passes across all matrix combinations.

**Minor observations (not blockers):**
- The generic `RuntimeException` catch block (line 169) constructs a `DocTreePath` with `docComment`, but `docComment` can be null when processing maven-mode fields — a guard would prevent masking the real error with an NPE.
- The `run()` catch block no longer prints stack traces for unexpected failures — worth preserving for diagnostic value.
- Stray blank line in `extractClassLink` (line 480) and minor javadoc typo in the test ("the a fixture" → "a fixture").

maven-resolver-tools/src/main/java/org/eclipse/aether/tools/ConfigurationCollectorDoclet.java
```diff
@@ -153,43 +152,85 @@ private boolean doRun(DocletEnvironment environment) {
...
+                    } else {
+                        // TODO: move to beginning of run() and validate mode before processing any types
+                        reportError("Unknown mode: " + mode);
+                        return false;
+                    }
+                } catch (DocTreePathAwareRuntimeException e) {
+                    reportError(e.getDocTreePath(), e.getMessage());
+                } catch (RuntimeException e) {
+                    DocTreePath rootPath = new DocTreePath(docTrees.getPath(field), docComment);
+                    reportError(rootPath, e.getMessage());
```
gnodet, line 236: Low-severity: The `DocTreePath` constructor calls `Objects.requireNonNull` on the `DocCommentTree` parameter, so if `docComment` is null when a `RuntimeException` is caught here, the catch block itself would NPE and mask the original error.

In resolver mode this is unreachable (early return on null `docComment`), but `processMavenField` does not guard against null `docComment` in the same way. Interestingly, `buildTagPath` (line 355–361 in this PR) already applies exactly this defensive pattern.

```suggestion
            } catch (RuntimeException e) {
                DocTreePath rootPath =
                        docComment != null ? new DocTreePath(docTrees.getPath(field), docComment) : null;
                reportError(rootPath, e.getMessage());
```

gnodet (commented):
**Re-review:** The regression flagged in the previous review (skipping fields without default values) was correctly fixed in the follow-up commit — nice work! The continue-on-error pattern with precise `DocTreePath` locations is well-designed and tested. One minor defensive-coding suggestion below.

gnodet (commented):
Well-designed continue-on-error pattern with precise `DocTreePath` error locations. The regression from commit 1 was correctly fixed in commit 2. CI passes across all matrix combinations. A few observations:

**Medium severity:**
- 🐛 **NPE risk in error handler** (line 168): The generic `RuntimeException` catch constructs `new DocTreePath(docTrees.getPath(field), docComment)` without null-checking `docComment`. The `DocTreePath` constructor calls `Objects.requireNonNull` on its `DocCommentTree` parameter, so when `docComment` is null (possible in maven mode), this will throw NPE and mask the original exception. The `buildTagPath` helper already has this null guard — the same pattern should be applied here.

**Low severity:**
- The top-level `run()` catch block no longer prints stack traces (`e.printStackTrace` was removed). For truly unexpected failures (NPEs, ClassCastExceptions), the stack trace is the primary diagnostic artifact.
- Stray blank line between the `if` condition and the `throw` statement in `extractClassLink`.
- Javadoc typo: "Classpath location of the a fixture" should be "Classpath location of a fixture".

**Strengths:**
- The `DocTreePathAwareRuntimeException` pattern allows reporting all errors at once rather than failing on the first one — significant UX improvement.
- The hardcoded `"@configurationDefaultValue"` fix in `extractClassLink` resolves a pre-existing bug.
- The `collectBlockTags` return type refactor is clean and enables precise error locations.

cstamas (approved):
Can we get this rebased?

gnodet (commented):
Well-structured improvement to error reporting in `ConfigurationCollectorDoclet`. The switch from opaque exception propagation to precise source-location-aware diagnostics via the Javadoc Reporter API is a nice improvement, and the fix for the hardcoded `@configurationDefaultValue` tag name in `extractClassLink` error messages is correct. Good test coverage for error paths.

One minor note: the outer catch in `run()` replaced `e.printStackTrace()` with `reportError()`, which only reports the message. For truly unexpected `RuntimeExceptions` that escape `doRun()`, the stack trace loss could make debugging harder. Consider including the exception class name or cause.

maven-resolver-tools/src/main/java/org/eclipse/aether/tools/ConfigurationCollectorDoclet.java
```diff
@@ -115,18 +151,31 @@ public String getName() {
...
+                            try {
+                                output = Paths.get(arg);
+                            } catch (InvalidPathException e) {
+                                throw new IllegalArgumentException("Invalid output file path: " + arg, e);
+                            }
+                        }),
+                new SingleArgumentOption(
+                        List.of("--mode", "-m"), "The scanning mode, either 'resolver' or 'maven'", "<mode>", arg -> {
+                            try {
+                                Mode.valueOf(arg.toUpperCase(Locale.ROOT));
```
gnodet: Bug: `Mode.valueOf()` result is discarded — the `mode` field stays at `RESOLVER` regardless of the `--mode` argument, silently breaking maven mode scanning.

```suggestion
                                mode = Mode.valueOf(arg.toUpperCase(Locale.ROOT));
```

maven-resolver-tools/src/main/java/org/eclipse/aether/tools/ConfigurationCollectorDoclet.java
```diff
@@ -262,19 +382,15 @@ private void processMavenField(
...
-        entry.put("configurationSource", source);
-        entry.put("configurationType", configurationType);
-        entry.put("supportRepoIdSuffix", "");
-        discovered.add(entry);
+        return new ConfigurationEntry(
+                String.valueOf(field.getConstantValue()),
+                path.getDocComment() != null ? getFullBodyContent(path) : "",
+                Objects.toString(defaultValue, ""),
+                getFullyQualifiedName(field),
+                Objects.toString(getSince(path), ""),
```
gnodet: Bug: `Objects.toString(getSince(path), "")` is incorrect because `getSince()` returns `Optional<String>`. Since `Optional.empty()` is never null, `Objects.toString` calls `Optional.toString()` producing the literal string "Optional.empty" instead of "". The resolver-mode counterpart at line 311 correctly uses `.orElse("")`.

```suggestion
                    getSince(path).orElse(""),
```

gnodet (changes requested):
**Re-review after new commits**

Previously flagged NPE risk (null docComment) is now fixed — nice work on the null-check at lines 213-217.

Two correctness bugs found in the new commits — see inline comments.

Additionally:
- `getFullyQualifiedName(field)` (lines 310/389) returns only the enclosing type's FQN, dropping the field name. Master produced `type.getQualifiedName() + "." + field.getSimpleName()` (e.g. `org.eclipse.aether.ConnectorKeys.CONFIG_PROP_THREADS`). Consider adding a `visitVariable` override or appending `"." + field.getSimpleName()` at the call site.
- Minor: javadoc typo "the a fixture" (test line 57), empty `@return` tag (line 330), run() catch still loses stack trace (line 191, low impact since doRun() now handles most exceptions per-field).

maven-resolver-tools/src/main/java/org/eclipse/aether/tools/ConfigurationCollectorDoclet.java
kwin: I no longer see a code path where rootPath may be null!

maven-resolver-tools/src/main/java/org/eclipse/aether/tools/ConfigurationCollectorDoclet.java
```diff
@@ -584,81 +700,198 @@ protected String defaultAction(DocTree node, Void p) {
...
-        String since = getSinceTag(docComment, fieldContext);
-        if (since == null && type != null) {
+    private Optional<String> getSince(DocTreePath path) {
+        String since = getSinceTag(path);
+        if (since == null) {
             // fall back to the enclosing type's @since
-            since = getSinceTag(docTrees.getDocCommentTree(type), null);
+            DocTreePath parentPath = path.getParentPath();
+            if (parentPath != null) {
+                return getSince(parentPath);
```
gnodet: The `getSince` fallback to the enclosing type's `@since` tag may no longer work correctly. The old code explicitly called `docTrees.getDocCommentTree(type)` to look up the TYPE's doc comment when a field lacked its own `@since` tag. The refactored code recurses via `DocTreePath.getParentPath()`, but `rootPath` (constructed at the top of the processing loop) is a root `DocTreePath` whose `getParentPath()` returns `null` per JDK API contract, so this fallback branch never executes.

This would affect configuration key classes where fields inherit `@since` from the enclosing type — for example, `GnupgConfigurationKeys` has `@since 2.0.0` on the class with 5 config fields that have no field-level `@since`, and `ConfigurationProperties` has many more such fields. These fields would produce empty `since` values in the generated documentation.

Worth verifying that the generated output still includes `since` values for inherited cases.

gnodet (commented):
**Re-review (3rd pass):** The three bugs flagged in the previous review (mode assignment, `Objects.toString` on `Optional`, `getFullyQualifiedName` dropping field name) are all correctly fixed — nice work! One remaining observation below about the `getSince` fallback behavior.

maven-resolver-tools/src/main/java/org/eclipse/aether/tools/ConfigurationCollectorDoclet.java
kwin: Indeed, added a test case and fixed the issue in https://github.com/apache/maven-resolver/pull/1983/commits/39dd159af2caeb582624eff4a9744707e7278f70.

gnodet (approved):
**Re-review (4th pass):** All four previously flagged issues are now resolved:

1. ✅ Mode assignment bug (discarded `valueOf` result) — fixed
2. ✅ `Objects.toString` on `Optional` producing garbage — fixed
3. ✅ `getFullyQualifiedName` dropping field name — fixed
4. ✅ `getSince` fallback to enclosing type's `@since` — fixed with proper Element hierarchy traversal and test coverage via `package-info.java`

The `getSince` inheritance fix is well-designed — using the Element hierarchy (`field → class → package → module`) instead of `DocTreePath.getParentPath()` correctly models Java's documentation inheritance semantics.

Two non-blocking nits:
- The assertion message `"no @since expected"` at line 109 of `ConfigurationCollectorDocletTest.java` is now stale since the expected value changed to `"2.0"` — consider updating to clarify inheritance from `package-info.java`
- `package-info.java` test fixture is missing a final newline

PR #2004, opened 2026-07-21
Track exception/cycle count locally in Results

### LINKED ISSUE #1995: Non-thread-safe access in synchronized Results class in DependencyCollectorDelegate
**Affected version:** HEAD

**File:** `maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/DependencyCollectorDelegate.java:571-598`

`addException()` and `addCycle()` are `synchronized` on `Results`, but they call `result.getExceptions()` and `result.getCycles()` on a `CollectResult` that may not be thread-safe. If `CollectResult`'s internal collections are not synchronized or safely published, concurrent modifications may not be visible.

Originally reported in #1944.

elharo:
Fixes #1995

Instead of querying result.getExceptions().size() and result.getCycles().size() directly (which are non-thread-safe), this tracks the count locally in the Results class.


elharo:
Closing - needs proper base

PR #2006, opened 2026-07-21
Log instead of swallow in IpcNamedLock.tryUnlock

### LINKED ISSUE #1994: Empty catch swallows exceptions in IpcNamedLock.tryUnlock()
**Affected version:** HEAD

**File:** `maven-resolver-named-locks-ipc/src/main/java/org/eclipse/aether/named/ipc/IpcNamedLock.java:108-115`

```java
private void tryUnlock(String contextId) {
    try {
        client.unlock(contextId);
    } catch (Exception e) {
        // Best-effort cleanup
    }
}
```

All exceptions from `unlock()` during timeout cleanup are silently discarded. While documented as intentional, this can mask real IO/connectivity issues.

Originally reported in #1944.

elharo:
Fixes #1994

Changes `tryUnlock()` to log at debug level instead of silently swallowing exceptions during timeout cleanup.


elharo:
Closing - needs proper base

PR #2008, opened 2026-07-22
Enh: Allow validator factory to abstain

### LINKED ISSUE #2007: ValidatorFactory should be allowed to return `null`
### Affected version

all

### Bug description

For simplicity sake, the validator should be allowed to return `null` for validator, like "abstain" from validation.

cstamas:
Changes:
* allow `ValidatorFactory` to "abstain" and return `NOOP`
* introduce alt execution path for managed dependencies (use cases varies)
* cache validators per session

Fixes: #2007

### DIFF
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultRepositorySystemValidator.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultRepositorySystemValidator.java
@@ -73,10 +73,10 @@ public class DefaultRepositorySystemValidator implements RepositorySystemValidat
     private Validator newInstance(RepositorySystemSession session, String name, ValidatorFactory factory) {
         if (session.getCache() != null) {
             return ((ConcurrentHashMap<String, Validator>)
+                            session.getCache().computeIfAbsent(session, SESSION_VALIDATORS, ConcurrentHashMap::new))
+                    .computeIfAbsent(name, k -> requireNonNull(factory.newInstance(session)));
         } else {
             return requireNonNull(factory.newInstance(session));
-        }
-            return factory.newInstance(session);
         }
     }
 
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultRepositorySystemValidator.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultRepositorySystemValidator.java
@@ -229,7 +229,7 @@ public class DefaultRepositorySystemValidator implements RepositorySystemValidat
                 }
                 for (Dependency managedDependency : request.getManagedDependencies()) {
                     try {
-                        validator.validateDependency(managedDependency);
+                        validator.validateManagedDependency(managedDependency);
                     } catch (Exception e) {
                         exceptions.add(e);
                     }
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultRepositorySystemValidator.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultRepositorySystemValidator.java
@@ -25,6 +25,7 @@ import javax.inject.Singleton;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
 import org.eclipse.aether.Keys;
@@ -53,10 +54,10 @@ import static java.util.Objects.requireNonNull;
 @Named
 public class DefaultRepositorySystemValidator implements RepositorySystemValidator {
     private static final Object SESSION_VALIDATORS = Keys.of(DefaultRepositorySystemValidator.class, "validators");
-    private final List<ValidatorFactory> validatorFactories;
+    private final Map<String, ValidatorFactory> validatorFactories;
 
     @Inject
-    public DefaultRepositorySystemValidator(List<ValidatorFactory> validatorFactories) {
+    public DefaultRepositorySystemValidator(Map<String, ValidatorFactory> validatorFactories) {
         this.validatorFactories = requireNonNull(validatorFactories, "validatorFactories cannot be null");
     }
 
@@ -69,11 +70,11 @@ public class DefaultRepositorySystemValidator implements RepositorySystemValidat
     }
 
     @SuppressWarnings("unchecked")
-    private Validator newInstance(RepositorySystemSession session, ValidatorFactory factory) {
+    private Validator newInstance(RepositorySystemSession session, String name, ValidatorFactory factory) {
         if (session.getCache() != null) {
-            return ((ConcurrentHashMap<Integer, Validator>)
+            return ((ConcurrentHashMap<String, Validator>)
                             session.getCache().computeIfAbsent(session, SESSION_VALIDATORS, ConcurrentHashMap::new))
-                    .computeIfAbsent(System.identityHashCode(factory), key -> factory.newInstance(session));
+                    .computeIfAbsent(name, key -> requireNonNull(factory.newInstance(session)));
         } else {
             return factory.newInstance(session);
         }
@@ -82,9 +83,9 @@ public class DefaultRepositorySystemValidator implements RepositorySystemValidat
     @Override
     public void validateVersionRequest(RepositorySystemSession session, VersionRequest request) {
         ArrayList<Exception> exceptions = new ArrayList<>();
-        for (ValidatorFactory factory : validatorFactories) {
-            Validator validator = newInstance(session, factory);
-            if (validator != null) {
+        for (Map.Entry<String, ValidatorFactory> entry : validatorFactories.entrySet()) {
+            Validator validator = newInstance(session, entry.getKey(), entry.getValue());
+            if (validator != ValidatorFactory.NOOP) {
                 try {
                     validator.validateArtifact(request.getArtifact());
                 } catch (Exception e) {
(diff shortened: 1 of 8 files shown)


maven-resolver-spi/src/main/java/org/eclipse/aether/spi/validator/Validator.java
```diff
@@ -54,6 +54,20 @@ default void validateMetadata(Metadata metadata) throws IllegalArgumentException
...
 
+    /**
+     * Validates managed dependency.
+     * <em>Important:</em> They are declarative constraints (version/scope/exclusion overrides) that only take effect
+     * when a matching dependency is encountered during collection. Validating them eagerly may reject valid builds
+     * where a BOM imports managed dependencies with uninterpolated property expressions (e.g. {@code ${osgi.version}})
+     * that are never actually used. If a managed dependency IS matched and its coordinates are invalid, the error
+     * will surface naturally during version resolution or artifact resolution.
+     *
+     * @param managedDependency The managed dependency to validate, never {@code null}.
```
elharo: nit: begin with lower case the
no periods at end

per Oracle guidelines

elharo (changes requested):
Needs tests

Is there a better signal than "null" for abstention that doesn't risk NPEs? Abstention seems equivalent to approval unless something requires some validator to approve?

gnodet (commented):
The overall design of the three features (abstain, managed-dependency differentiation, caching) is sound and addresses real needs. The separate `validateManagedDependency` path correctly avoids false positives from uninterpolated BOM property expressions — well-documented rationale.

Two implementation issues in the caching logic:

1. **Null values not cached**: `ConcurrentHashMap.computeIfAbsent` does not store entries when the mapping function returns null. When a factory abstains (returns null), the factory is re-invoked on every subsequent validation call within the same session — contradicting the Javadoc claim that "even nulls are cached." Fix: use a sentinel object pattern (e.g., `private static final Validator ABSENT = new Validator() {};`) and unwrap on retrieval.

2. **`System.identityHashCode` as map key**: `System.identityHashCode` is not guaranteed unique across distinct object instances. Hash collisions cause the second factory's `newInstance()` to never be called — it silently reuses the first factory's Validator. Since `Integer.equals` compares by value, two different factory instances with the same identityHashCode map to the same key. Fix: use the factory instance itself as the key (`ConcurrentHashMap<ValidatorFactory, Validator>`), which relies on reference equality via `Object.equals`.

**Missing test coverage**: No tests cover the three new behaviors — factory returning null to abstain, per-session caching, or managed dependencies routed to `validateManagedDependency`.

maven-resolver-spi/src/main/java/org/eclipse/aether/spi/validator/ValidatorFactory.java
```diff
@@ -26,11 +26,21 @@
...
+     * The no-op (does not validate anything) instance of validator.
+     *
+     * @since 2.0.22
+     */
+    Validator NOOP = new Validator() {};
+
     /**
-     * Creates a new validator for the session.
+     * Creates a new validator for the session, or {@link #NOOP} to abstain from validation. Factory is consulted
+     * once per session (if cache present in session) and returned instances (even {@code null}s) are cached and reused
```
gnodet: Stale Javadoc: "returned instances (even {@code null}s) are cached and reused" contradicts the `@return` on line 44 which says "never {@code null}", and the implementation wraps the call in `requireNonNull`. The "even nulls" text looks like a leftover from when null was the abstention signal.

```suggestion
     * Creates a new validator for the session, or {@link #NOOP} to abstain from validation. Factory is consulted
     * once per session (if cache present in session) and returned instances are cached and reused
     * across single session.
```

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultRepositorySystemValidator.java
```diff
@@ -65,22 +69,35 @@ private void mayThrow(List<Exception> exceptions, String message) {
         }
     }
 
+    @SuppressWarnings("unchecked")
+    private Validator newInstance(RepositorySystemSession session, String name, ValidatorFactory factory) {
+        if (session.getCache() != null) {
+            return ((ConcurrentHashMap<String, Validator>)
+                            session.getCache().computeIfAbsent(session, SESSION_VALIDATORS, ConcurrentHashMap::new))
+                    .computeIfAbsent(name, key -> requireNonNull(factory.newInstance(session)));
+        } else {
```
gnodet, line 78: Nit: the caching path guards with `requireNonNull` but this non-caching fallback does not. If a factory violates the contract and returns null here, the error surfaces as a confusing NPE downstream rather than a clear fail-fast.

```suggestion
        } else {
            return requireNonNull(factory.newInstance(session));
        }
```

gnodet (commented):
The critical bugs from the previous review are properly fixed — the NOOP sentinel pattern and string-keyed Map are clean solutions. Two minor issues remain.

maven-resolver-spi/src/main/java/org/eclipse/aether/spi/validator/Validator.java
cstamas: While I agree with this sentiment, this javadoc method is not the single one in this PR not following that, and in this codebase this is not the single one either (in fact, whole codebase follows this style). I'd leave this change to some time, and IMO the change should be done consistently across whole codebase.

cstamas: Hm, was wrong: lowercase stands but not the "no period at end"?

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultRepositorySystemValidator.java
```diff
@@ -50,10 +53,11 @@
 @Singleton
 @Named
 public class DefaultRepositorySystemValidator implements RepositorySystemValidator {
-    private final List<ValidatorFactory> validatorFactories;
+    private static final Object SESSION_VALIDATORS = Keys.of(DefaultRepositorySystemValidator.class, "validators");
+    private final Map<String, ValidatorFactory> validatorFactories;
 
     @Inject
-    public DefaultRepositorySystemValidator(List<ValidatorFactory> validatorFactories) {
+    public DefaultRepositorySystemValidator(Map<String, ValidatorFactory> validatorFactories) {
```
elharo, line 60: public API change

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultRepositorySystemValidator.java
```diff
@@ -89,19 +106,21 @@ public void validateVersionRequest(RepositorySystemSession session, VersionReque
...
-            Validator validator = factory.newInstance(session);
-            try {
-                validator.validateArtifact(request.getArtifact());
-            } catch (Exception e) {
-                exceptions.add(e);
-            }
-            for (RemoteRepository repository : request.getRepositories()) {
+        for (Map.Entry<String, ValidatorFactory> entry : validatorFactories.entrySet()) {
+            Validator validator = newInstance(session, entry.getKey(), entry.getValue());
+            if (validator != ValidatorFactory.NOOP) {
```
elharo, line 111: do we really need all these if (validator != ValidatorFactory.NOOP) checks? If it's Validator.NOOP then the validator doesn't throw any exceptions, nothing is added to the exceptions list, and we get exactly the same result. It feels cleaner to just treat Validator.NOOP like every other one.

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultRepositorySystemValidator.java
```diff
@@ -50,10 +53,11 @@
 @Singleton
 @Named
 public class DefaultRepositorySystemValidator implements RepositorySystemValidator {
-    private final List<ValidatorFactory> validatorFactories;
+    private static final Object SESSION_VALIDATORS = Keys.of(DefaultRepositorySystemValidator.class, "validators");
+    private final Map<String, ValidatorFactory> validatorFactories;
```
elharo, line 57: I'm sure there's a reason but I don't immediately see what the String key is here or why we need it. A comment might be helpful.

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultRepositorySystemValidator.java
cstamas: public API defined here https://maven.apache.org/resolver/api-compatibility.html
So is not. Moreover, client code should use DI, and DI protects from changes like these, and they should inject managed beans, not manually (re)construct them.

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultRepositorySystemValidator.java
cstamas: Agreed, this is leftover from `null` being allowed.

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultRepositorySystemValidator.java
cstamas: See "What is Sisu" here https://eclipse.dev/sisu/org.eclipse.sisu.inject/

This is most common technique to have injected "variations" (or qualified beans) for same (component) contract.

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultRepositorySystemValidator.java
cstamas: Well, am leaving it as it is not hurting. OTOH, with today's app frameworks we at least avoid busy-looping (and calling no-op method) over thousands of managed deps...

gnodet (approved):
All previously identified issues are properly fixed. The NOOP sentinel pattern is a clean solution for the abstain mechanism, the Javadoc is now consistent, and null enforcement is uniform across both code paths. Test coverage is solid.

PR #2012, opened 2026-07-23
Fix InterruptedException silently dropping task in SmartExecutor.Limited.submit(Runnable)

### LINKED ISSUE #1993: InterruptedException causes task to be silently dropped in SmartExecutor.Limited.submit(Runnable)
**Affected version:** HEAD

**File:** `maven-resolver-util/src/main/java/org/eclipse/aether/util/concurrency/SmartExecutor.java:177-179`

```java
} catch (InterruptedException e) {
    Thread.currentThread().interrupt();
}
```

In `Limited.submit(Runnable)` (line 155), if `semaphore.acquire()` throws `InterruptedException`, the interrupt flag is restored but the caller's task is never executed. The method simply returns. The `submit(Callable)` variant (line 183-213) correctly handles this by returning a failed `CompletableFuture`. The `Runnable` overload should do something analogous (e.g., run the task inline or throw).

Originally reported in #1944.

elharo:
Fixes #1993

In `Limited.submit(Runnable)` (line 155), if `semaphore.acquire()` throws `InterruptedException`, the interrupt flag was restored but the caller's task was never executed — it was silently dropped. The `submit(Callable)` variant correctly handles this by returning a failed `CompletableFuture`.

This fix throws a `RuntimeException` wrapping the `InterruptedException` after restoring the interrupt flag, so the caller is notified that the task could not be submitted.

### DIFF
--- a/maven-resolver-util/src/main/java/org/eclipse/aether/util/concurrency/SmartExecutor.java
+++ b/maven-resolver-util/src/main/java/org/eclipse/aether/util/concurrency/SmartExecutor.java
@@ -176,6 +176,7 @@ public interface SmartExecutor extends AutoCloseable {
                 }
             } catch (InterruptedException e) {
                 Thread.currentThread().interrupt();
+                throw new RuntimeException(e);
             }
         }
 
--- a/maven-resolver-util/src/main/java/org/eclipse/aether/util/concurrency/SmartExecutor.java
+++ b/maven-resolver-util/src/main/java/org/eclipse/aether/util/concurrency/SmartExecutor.java
@@ -35,8 +35,10 @@ import java.util.concurrent.Semaphore;
 public interface SmartExecutor extends AutoCloseable {
     /**
      * Submits a {@link Runnable} to execution.
+     *
+     * @throws RejectedExecutionException If this executor cannot accept the task.
      */
-    void submit(Runnable runnable);
+    void submit(Runnable runnable) throws RejectedExecutionException;
 
     /**
      * Submits a {@link Callable} to execution, returns a {@link CompletableFuture}.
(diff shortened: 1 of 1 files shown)


maven-resolver-util/src/main/java/org/eclipse/aether/util/concurrency/SmartExecutor.java
```diff
@@ -176,6 +176,7 @@ public void submit(Runnable runnable) {
                 }
             } catch (InterruptedException e) {
                 Thread.currentThread().interrupt();
+                throw new RuntimeException(e);
```
elharo: Throwing a raw RuntimeException is an antipattern. Is there a subclass of RuntimeException we could throw instead?

Whatever exception we throw should be documented in javadoc for this method.

Is the exception handled upstream or is that something we also need to do?

maven-resolver-util/src/main/java/org/eclipse/aether/util/concurrency/SmartExecutor.java
```diff
@@ -35,8 +35,10 @@
 public interface SmartExecutor extends AutoCloseable {
     /**
      * Submits a {@link Runnable} to execution.
+     *
+     * @throws RejectedExecutionException If this executor cannot accept the task.
      */
-    void submit(Runnable runnable);
+    void submit(Runnable runnable) throws RejectedExecutionException;
```
elharo, line 41: don't use throws clause for RuntimeExceptions, javadoc comment only

gnodet (approved):
Correct fix for a real concurrency bug. The previous code silently dropped the submitted task on `InterruptedException`; the fix properly restores the interrupt flag and throws `RejectedExecutionException` (wrapping the original cause), which is semantically precise and avoids violating the concurrency limit that `Limited` exists to enforce.

cstamas (approved):

PR #2014, opened 2026-07-24
Fix #2013: optimize TransitiveDependencyManager performance

### LINKED ISSUE #2013: BfDependencyCollector pool cache hit/miss changes graph structure (cache-transparency violation)
### Summary

The `BfDependencyCollector.doRecurse()` method has a cache-transparency issue: whether a dependency node retains its children depends on whether the **pool cache** hits or misses — a performance optimization is changing the semantic result.

### Root cause

`doRecurse()` uses two independent caching/dedup layers with **different keys**:

1. **Pool cache** — keyed on `(artifact, repos, selector, manager, traverser, filter)`, i.e. it includes the `DependencyManager` (compared via `equals`).
2. **GACE skipper** — keyed on artifact identity only (`G:A:C:E`, no version, no manager).

The interaction at [BfDependencyCollector.java lines ~400-436](https://github.com/apache/maven-resolver/blob/master/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/bf/BfDependencyCollector.java):

```java
List<DependencyNode> children = pool.getChildren(key); // key includes DependencyManager
if (children == null) {        // POOL MISS → skipper is consulted
    if (!skipper.skipResolution(child, parents)) {
        // resolve children, cache in pool
    }
    // else: node gets ZERO children (skipper marked it as duplicate)
} else {                       // POOL HIT → skipper is NEVER consulted
    child.setChildren(children);
}
```

When the pool hits, the skipper is bypassed and the node gets cached children. When the pool misses, the skipper IS consulted and may mark the node as `skippedAsDuplicate`, giving it zero children. The `ConflictResolver` (in default `Verbosity.NONE`) then prunes childless conflict-loser nodes entirely.

This means the graph structure depends on whether the pool cache hits or misses — which is a function of `DependencyManager.equals()`, not of the dependency being resolved.

### How this manifests

Discovered via [apache/maven-shade-plugin#819](https://github.com/apache/maven-shade-plugin/issues/819). Consider a project where:
- Direct dependency `a` depends on `b` (no classifier) and `b:alt` (classifier "alt")
- Both `b` and `b:alt` share a transitive dependency `c`

Under **Maven 3.10** (`ClassicDependencyManager`): at depth ≥ 2, `deriveChildManager()` returns the same instance → pool key for `c` under `b` equals pool key for `c` under `b:alt` → **pool hit** → skipper is never consulted → `c` appears under both variants.

Under **Maven 4** (`TransitiveDependencyManager`): `deriveChildManager()` always creates a new instance with a unique `path` field (added in #1539) → pool key differs → **pool miss** → skipper marks `c` under `b:alt` as `skippedAsDuplicate` → `c` gets zero children → `ConflictResolver` prunes it → `c` disappears from `b:alt`'s subtree.

The same Resolver 2.0.20 produces different graphs under Maven 3.10 vs Maven 4, purely due to the `DependencyManager` choice.

### Reproducer

Standalone reproducer (no plugin build needed): https://github.com/aschemaven/mshade-819-reproducer

```
./run.sh <maven-home>
```

Results:

| Maven | Resolver | `c` under `b` | `c` under `b:alt` |
|---|---|---|---|
| 3.9.16 | 1.x | ✅ present | ✅ present |
| 3.10.0-rc-1 | 2.0.20 | ✅ present | ✅ present |
| 4.0.0-rc-5 | 2.0.13 | ✅ present | ❌ pruned |
| 4.0.0-rc-5 + `-Daether.conflictResolver.verbose=true` | 2.0.13 | ✅ present | ✅ present |

### Suggested fix

Make the pool cache transparent — either:

1. **Always consult the skipper** regardless of pool hit/miss, but let the pool provide cached children if the skipper says "don't skip".
2. **On pool miss + skipper skip**, still try to provide cached children from the winning node's pool entry (since the GACE keys match), so the node retains its children even when the pool key differs.

Either approach would make the graph structure independent of the `DependencyManager.equals()` behavior.

### Workaround

The maven-shade-plugin has a workaround in [PR #820](https://github.com/apache/maven-shade-plugin/pull/820): collecting with `ConflictResolver.CONFIG_PROP_VERBOSE = STANDARD` so pruned nodes are retained as markers. This is a valid fix on the plugin side, but other plugins that walk the dependency graph may be silently affected by the same issue.

gnodet:
## Problem

`TransitiveDependencyManager` (`deriveUntil=Integer.MAX_VALUE`) creates a new child manager at every node in the dependency graph. On large multi-module projects (4,383 modules), JFR profiling showed this consuming **28% of total CPU time** — methods like `deriveChildManager`, `MMap.done()`, `Key.equals()`, and `AbstractMap.equals()` that have zero samples under `ClassicDependencyManager`.

## Changes

Four layered optimizations, each addressing a distinct hotspot:

### 1. Instance reuse in `deriveChildManager()` (ec77f732)

When no new management data is collected (the common case for transitive POMs without `<dependencyManagement>`), return `this` instead of creating a new instance. This restores pool cache transparency — the BF collector's `DataPool.GraphKey` includes the manager, so semantically-equal-but-distinct managers were defeating the cache and preventing subtree reuse.

### 2. `Key` class: cache GACE coordinate strings (69d7e392)

The `Key` inner class stored an `Artifact` reference and called `getGroupId()`/`getArtifactId()` etc. on every `equals()`. Through `RelocatedArtifact` (a delegation wrapper), each getter did a null-check + virtual dispatch — 9% of CPU in the JFR profile. Now caches the four coordinate strings eagerly in the constructor.

### 3. `MMap.DoneMMap.equals()`: identity + hashCode short-circuit (69d7e392)

Added `this == o` identity check, narrowed `instanceof` to `DoneMMap`, and added pre-computed `hashCode` mismatch rejection before the expensive `delegate.equals()` (which triggers `Key.equals()` for every map entry).

### 4. Replace `ArrayList` path with parent pointer — cons-list (52d9d4b8)

The previous design stored all ancestors in an `ArrayList<AbstractDependencyManager>` that was **O(depth) copied** on every `deriveChildManager()`. The `equals()` method compared the entire list element-by-element, each element comparing its own maps — **O(depth²)** worst case.

Replaced with a single `AbstractDependencyManager parent` pointer (cons-list). Siblings share the same parent reference. Key properties:

- **O(1) derive** — no list copy, just set `parent = this`
- **Cascading hashCode** — each level incorporates its parent's pre-computed hash, so a single `int` comparison reflects the entire ancestor chain
- **Hash-guarded recursive equals** — parent comparison is recursive but each level short-circuits on hashCode mismatch, and shared parents (same `==` identity) short-circuit immediately

## Benchmark Results

Test project: [maven-multiproject-generator](https://github.com/maven-turbo-reactor/maven-multiproject-generator) (4,383 modules), `mvn validate -T1`:

| Configuration | Time | vs baseline |
|---|---|---|
| Maven 4 rc-6 unpatched (`TransitiveDependencyManager`) | 793s | 1× |
| Maven 4 + `-Dmaven.resolver.dependencyManagerTransitivity=false` | 19s | 42× |
| Maven 4 + this PR (all optimizations) | **~3s** | **~270×** |

GC pressure also reduced: 10 → 8 GC events, 59ms → 43ms total pause time.

## Notes

- All three manager subclasses updated: `TransitiveDependencyManager`, `ClassicDependencyManager`, `DefaultDependencyManager`
- Lookup semantics preserved: parent chain walked from leaf toward root, last hit = nearest-to-root wins
- `managedLocalPaths` intentionally excluded from `equals()`/`hashCode()` (unchanged)
- Full resolver test suite passes (BF + DF collectors, all three managers)

### DIFF
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/bf/BfDependencyCollector.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/collect/bf/BfDependencyCollector.java
@@ -377,8 +377,6 @@ public class BfDependencyCollector extends DependencyCollectorDelegate {
 
         DependencySelector childSelector =
                 parentContext.depSelector != null ? parentContext.depSelector.deriveChildSelector(context) : null;
-        DependencyManager childManager =
-                parentContext.depManager != null ? parentContext.depManager.deriveChildManager(context) : null;
         DependencyTraverser childTraverser =
                 parentContext.depTraverser != null ? parentContext.depTraverser.deriveChildTraverser(context) : null;
         VersionFilter childFilter =
@@ -389,50 +387,80 @@ public class BfDependencyCollector extends DependencyCollectorDelegate {
                 : remoteRepositoryManager.aggregateRepositories(
                         args.session, parentContext.repositories, descriptorResult.getRepositories(), true);
 
-        Object key = args.pool.toKey(
+        // Optimization: try pool cache with the parent manager as a speculative key BEFORE
+        // calling deriveChildManager. When deriveChildManager returns `this` (the common case —
+        // no new management data at this depth), the parent manager IS the child manager, so
+        // this speculative key matches any previously stored entry. On pool hit, we skip
+        // deriveChildManager entirely, saving even the memoization cache lookup.
+        Object speculativeKey = args.pool.toKey(
                 parentContext.dependency.getArtifact(),
                 childRepos,
                 childSelector,
-                childManager,
+                parentContext.depManager,
                 childTraverser,
                 childFilter);
-
-        List<DependencyNode> children = args.pool.getChildren(key);
-        if (children == null) {
-            boolean skipResolution = args.skipper.skipResolution(child, parentContext.parents);
-            if (!skipResolution) {
-                List<DependencyNode> parents = new ArrayList<>(parentContext.parents.size() + 1);
-                parents.addAll(parentContext.parents);
-                parents.add(child);
-                for (Dependency dependency : descriptorResult.getDependencies()) {
-                    if (childSelector != null && !childSelector.selectDependency(dependency)) {
-                        continue;
-                    }
-                    RequestTrace childTrace = collectStepTrace(
-                            parentContext.trace, args.request.getRequestContext(), parents, dependency);
-                    PremanagedDependency premanagedDependency = PremanagedDependency.create(
-                            childManager, dependency, disableVersionManagement, args.premanagedState);
-                    DependencyProcessingContext processingContext = new DependencyProcessingContext(
-                            childSelector,
-                            childManager,
-                            childTraverser,
-                            childFilter,
-                            childTrace,
-                            childRepos,
-                            descriptorResult.getManagedDependencies(),
-                            parents,
-                            dependency,
-                            premanagedDependency);
-                    // resolve descriptors ahead for managed dependency
-                    processingContext.withDependency(processingContext.premanagedDependency.getManagedDependency());
-                    resolveArtifactDescriptorAsync(args, processingContext, results);
-                    args.dependencyProcessingQueue.add(processingContext);
-                }
-                args.pool.putChildren(key, child.getChildren());
-                args.skipper.cache(child, parents);
-            }
-        } else {
+        List<DependencyNode> children = args.pool.getChildren(speculativeKey);
+        if (children != null) {
             child.setChildren(children);
+            return;
+        }
+
+        // Speculative miss — compute the actual derived manager
+        DependencyManager childManager =
+                parentContext.depManager != null ? parentContext.depManager.deriveChildManager(context) : null;
+
+        Object key;
+        if (childManager == parentContext.depManager) {
+            // Manager unchanged — speculative key was correct, already checked and missed
+            key = speculativeKey;
+        } else {
+            // Manager changed — recompute key and check pool again with the correct manager
+            key = args.pool.toKey(
+                    parentContext.dependency.getArtifact(),
+                    childRepos,
+                    childSelector,
+                    childManager,
+                    childTraverser,
+                    childFilter);
+            children = args.pool.getChildren(key);
+            if (children != null) {
+                child.setChildren(children);
+                return;
+            }
+        }
+
+        // True cache miss — do full resolution
+        boolean skipResolution = args.skipper.skipResolution(child, parentContext.parents);
+        if (!skipResolution) {
+            List<DependencyNode> parents = new ArrayList<>(parentContext.parents.size() + 1);
+            parents.addAll(parentContext.parents);
+            parents.add(child);
+            for (Dependency dependency : descriptorResult.getDependencies()) {
+                if (childSelector != null && !childSelector.selectDependency(dependency)) {
+                    continue;
+                }
+                RequestTrace childTrace =
+                        collectStepTrace(parentContext.trace, args.request.getRequestContext(), parents, dependency);
+                PremanagedDependency premanagedDependency = PremanagedDependency.create(
+                        childManager, dependency, disableVersionManagement, args.premanagedState);
+                DependencyProcessingContext processingContext = new DependencyProcessingContext(
+                        childSelector,
+                        childManager,
+                        childTraverser,
+                        childFilter,
+                        childTrace,
+                        childRepos,
+                        descriptorResult.getManagedDependencies(),
+                        parents,
+                        dependency,
+                        premanagedDependency);
+                // resolve descriptors ahead for managed dependency
+                processingContext.withDependency(processingContext.premanagedDependency.getManagedDependency());
+                resolveArtifactDescriptorAsync(args, processingContext, results);
+                args.dependencyProcessingQueue.add(processingContext);
+            }
+            args.pool.putChildren(key, child.getChildren());
+            args.skipper.cache(child, parents);
         }
     }
 
--- a/maven-resolver-impl/src/test/java/org/eclipse/aether/internal/impl/collect/bf/BfWithSkipperDependencyCollectorTest.java
+++ b/maven-resolver-impl/src/test/java/org/eclipse/aether/internal/impl/collect/bf/BfWithSkipperDependencyCollectorTest.java
@@ -25,6 +25,7 @@ import org.eclipse.aether.collection.CollectRequest;
 import org.eclipse.aether.collection.CollectResult;
 import org.eclipse.aether.collection.DependencyCollectionException;
 import org.eclipse.aether.graph.Dependency;
+import org.eclipse.aether.graph.DependencyNode;
 import org.eclipse.aether.graph.Exclusion;
 import org.eclipse.aether.impl.ArtifactDescriptorReader;
 import org.eclipse.aether.internal.impl.StubRemoteRepositoryManager;
@@ -37,6 +38,8 @@ import org.eclipse.aether.util.graph.selector.ExclusionDependencySelector;
 import org.junit.jupiter.api.Test;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 /**
  * UT for {@link BfDependencyCollector}.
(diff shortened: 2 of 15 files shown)


gnodet (commented):
Clean, well-targeted fix for the cache-transparency bug described in #2013. The optimization to reuse the `DependencyManager` instance when no new management data is collected is semantically correct — it's placed at exactly the right point in `deriveChildManager` (after all management-data collection but before `newInstance()`), and the `isApplied()` guard prevents freezing depth below the `applyFrom` threshold.

A couple of minor, non-blocking observations:

- The unit test only covers `TransitiveDependencyManager`. `DefaultDependencyManager` (`applyFrom=0`) also benefits from the optimization and could be tested with a similar `assertSame`/`assertNotSame` pattern. The existing `testDefault` test indirectly covers management correctness, so this is a minor coverage gap.
- An additional test where depth-1 context contributes management data (making `this.managedVersions` non-null), then the next depth has an empty context (optimization fires), and a subsequent depth has management data for the *same* key would explicitly document the "nearer-to-root wins" invariant — analysis of `getManagedVersion` confirms it iterates the path from root (first match wins), so this is functionally correct already.

Thread safety is fine: the optimization returns `this` without mutation, and `newInstance` creates new `ArrayList` copies of `this.path`, so concurrent calls from different BF collector branches are safe.

gnodet:
Thanks for the review! Both suggestions addressed in 08029607:

1. **`DefaultDependencyManager` coverage** — added `testDeriveChildManagerReusesInstanceWithDefaultManager` which verifies the optimization fires at the very first derivation with `applyFrom=0` (assertSame), and that it correctly creates new instances when management data is present (assertNotSame).

2. **"Nearer-to-root wins" invariant** — added `testNearerToRootWinsAfterOptimizationReusesInstance` which explicitly documents the invariant: root contributes `x:1.0` at depth 1, the optimization reuses the instance at depth 2→3, then a deeper POM tries to override `x` with `2.0` — the test verifies `containsManagedVersion` finds it in ancestors, skips collection (so the optimization fires again), and `getManagedVersion` still returns `1.0`.

gnodet (commented):
**Re-review** after new commit (08029607).

The new commit cleanly addresses both suggestions from the previous review:

1. **DefaultDependencyManager coverage** — new test verifies the optimization fires at the very first derivation (depth 0→1) since `applyFrom=0` makes `isApplied()` true immediately. The `assertSame`/`assertNotSame` assertions are accurate.

2. **Nearer-to-root-wins invariant test** — verifies that when a depth-1 context contributes management data and the next depth has an empty context (optimization reuses `this`), a subsequent depth with management data for the *same* key correctly inherits the nearer-to-root value. The assertion chain through `containsManagedVersion` → skip local collection → all five managed variables stay null is sound.

Test coverage for the optimization is now comprehensive across both manager implementations, with and without management data, and across the `applyFrom` boundary. No issues found.

elharo (approved):

gnodet (commented):
**Re-review** after new commit 69d7e392 ("Optimize Key, MMap, and AbstractDependencyManager equality checks").

Three well-targeted performance optimizations backed by JFR profiling data:

1. **Key class** — caches GACE coordinate strings at construction to avoid repeated virtual dispatch through `RelocatedArtifact` wrappers.
2. **MMap.DoneMMap.equals()** — adds identity short-circuit and hashCode fast-rejection before expensive deep-map equality.
3. **AbstractDependencyManager.equals()** — adds hashCode fast-rejection and reorders comparisons to check cheap `int` (depth) before expensive `ArrayList` (path).

All three are correct: hashCode is pre-computed in each case so fast-rejection is sound, the `DoneMMap` instanceof narrowing is safe (stored maps are always null or `DoneMMap` via `deriveChildManager`/`newInstance`), and the `Key.toString()` format change only affects debug output. The claimed 42× speedup (793s → 19s) on a 4,383-module project is impressive.

Previous commits remain unchanged and were already approved. No issues found.

gnodet (commented):
**Re-review** after new commit 52d9d4b8 ("Replace ArrayList path with parent pointer (cons-list)").

Clean refactor that replaces the O(depth) `ArrayList<AbstractDependencyManager> path` with an O(1) parent-pointer cons-list. Key observations:

- **Lookup semantics preserved:** The old code iterated root-to-leaf with early return (first hit = nearest root wins). The new code walks leaf-to-root with overwrite (last hit = nearest root wins). Both correctly implement "nearest to root wins." Additionally, `containsManagedXxx()` guards during derive prevent duplicate keys in the chain, making walk direction irrelevant in practice.

- **Cascading hash is correct:** Uses `parent.hashCode` (pre-computed field) so a single int comparison reflects the entire ancestor chain.

- **Recursive equals is safe:** `this == obj` identity check short-circuits for shared parents (siblings), and the hash guard rejects most mismatches at the first level. Stack overflow risk is theoretical only — Maven dependency trees don't reach problematic depth, and the instance-reuse optimization prevents chain growth at empty-management levels.

- **Natural fit:** Siblings already share the same parent after the instance-reuse optimization from the earlier commit, making identity-based equality effectively O(1) in the common case.

No issues found.

gnodet (commented):
**Re-review** after new commit 7166e29c.

Two well-layered, correctness-preserving optimizations:

1. **Speculative pool key check** in `BfDependencyCollector.doRecurse()` — tries the pool with `parentContext.depManager` first (betting `deriveChildManager` returns `this`), falling back to recomputing the key when the identity check fails. Null-parent case handled correctly.

2. **IdentityHashMap memoization cache** in `AbstractDependencyManager` for `deriveChildManager` results — sound because the result depends only on the managed dependencies list and current manager state (fixed per-instance). Keying on `getManagedDependencies()` with identity-based lookup is appropriate given DataPool's list interning. A miss is only a performance concern, not a correctness issue.

Thread safety of the IdentityHashMap is correctly guaranteed by the BFS collector's single-threaded traversal loop. The three new tests are well-structured: identity-based cache hits, cache discrimination between different lists, and the common empty-list case. `assertSame` vs `assertEquals` usage is precisely correct.

No issues found.

gnodet (commented):
**Re-review** after new commit 6205d17d.

Two refinements to the optimization stack:

1. **Speculative pool key check** in `BfDependencyCollector` — avoids calling `deriveChildManager` on cache hits. When `deriveChildManager` returns `this` (common case), the speculative key matches. When the manager changes, falls back correctly. Null case handled properly.

2. **Single-entry memoization cache** (`lastManagedDeps` / `lastDeriveResult`) replaces the unbounded `IdentityHashMap` — well-suited to BFS traversal where consecutive siblings share the same interned managed-dependency list. Using identity (`==`) is correct given DataPool's list interning guarantee. Placement after the `isDerived()` guard ensures correctness at all depths.

Choosing a single-entry cache over an unbounded map is a good design decision — avoids retaining every derived DependencyManager and the dependency subtrees they reference, trading a small miss rate for bounded memory.

Three new test methods provide good coverage: identity-based cache hits, cache misses on different list objects, single-entry eviction semantics, and the empty-list case. No issues found.

gnodet (commented):
**Re-review** after commits 071bbe0f and the ancestor-maps refactor.

Two important changes:

1. **Correctness fix** (071bbe0f) — the `containsManagedXxx()` guards now also check `this.managedVersions` (and peers), fixing a bug where the reuse optimization could hide own management data from children. The corresponding test update (flipping `assertSame` to `assertNotSame` for DefaultDependencyManager depth 1→2 with parent management) correctly validates this.

2. **O(1) cumulative ancestor HashMaps** — replaces O(depth) parent-chain walks in `getManagedXxx()` with pre-built cumulative maps constructed incrementally at construction time. Zero-copy sharing when the parent has no per-level data. Since `containsManagedXxx()` prevents duplicate entries during derivation, the HashMap has at most one entry per key, so `putAll` ordering is irrelevant in practice — preserving "nearest to root wins" semantics.

Exclusion additive merge in `mergeAncestorExclusions()` correctly uses `HashMap.merge()` to combine collections from multiple levels, and `getManagedExclusions()` returns shared references safely since the caller wraps them in a new `LinkedHashSet`.

No issues found.

gnodet (commented):
**Re-review** after new commit 3f600516 ("Fix file descriptor leak in FileLockNamedLockFactory").

Clean fix for a real-world problem (FD exhaustion on macOS with its 256 default limit). Adds an LRU-bounded idle channel pool, capping file descriptor consumption during large reactor builds.

Thread safety is correct:
- Base class `NamedLockFactorySupport` guarantees `createLock`/`destroyLock` are serialized per key via `ConcurrentHashMap.compute()`
- All `LinkedHashMap` access is properly `synchronized`
- Eviction I/O (`evicted.close()`) is performed outside the monitor

The configurable `maxCachedChannels` (default 200) is a reasonable tradeoff between syscall avoidance and FD conservation. The single-entry eviction in `destroyLock` converges correctly even when the pool exceeds the cap.

No issues found.

gnodet (commented):
**Re-review** after commit fed4797a.

Three well-designed performance optimizations:

1. **LayeredMap cons-list** — replaces HashMap ancestor maps with zero-copy linked map fragments. Construction becomes O(1) (link on top) instead of O(n) (HashMap copy). Lookups are O(layers), but layers is typically 2-5 in practice. Correctness is maintained because `containsManagedXxx()` during derive ensures at most one value per key across all layers.

2. **4-entry ring buffer memoization** — expands the single-entry cache to handle reactors that alternate between a few distinct BOM patterns. Uses identity comparison (`==`) matching the DataPool interning contract.

3. **Key.hashCode varargs elimination** — replaces `Objects.hash(a,b,c,d)` (allocates `Object[]`) with manual `31*h + field.hashCode()` computation.

One minor observation: comments at lines ~326 and ~335 in `AbstractDependencyManager` still say "O(1) lookup in cumulative ancestor map" but lookups are now O(layers) after the LayeredMap change. The `LayeredMap` class javadoc correctly documents the O(layers) complexity. Non-blocking.

Also, the commit message understates the scope — consider mentioning LayeredMap and ring buffer alongside the varargs fix for git-blame clarity.

gnodet (commented):
**Re-review** after latest commits.

Two focused performance optimizations:

1. **Remaining `Objects.hash` varargs elimination** + memo cache bumped from 4 to 16 entries. `Holder.hashCode` change (`Objects.hash(value)` → `value.hashCode()`) produces different hash values but remains consistent with `Holder.equals`. 16 entries is still constant memory (16 reference pairs), reasonable for reactors with many distinct BOM patterns.

2. **O(1) cycle detection in `PathConflictResolver`** — replaces O(depth) parent-chain walk with a `HashSet<Object>` for `conflictIdsOnPath`, addressing a JFR-identified 3.9% CPU hotspot. HashSet is constructed once per `Path`, children copy via `new HashSet<>(parent)` after parent is fully built. Single-threaded traversal — no data race concern.

No issues found.

cstamas (approved):

gnodet (commented):
Excellent performance optimization backed by JFR profiling data. The four layered optimizations (instance reuse, Key caching, MMap equality short-circuits, cons-list parent pointer with LayeredMap) are correct and well-tested. A couple of minor notes:

- Inline comment at `AbstractDependencyManager.java:468` says "A 4-entry ring buffer" but `MEMO_CACHE_SIZE` was increased to 16 — the Javadoc at line 189 is correct, but this inline comment is stale
- The `transient` modifier on `memoKeys`/`memoValues`/`memoIndex` fields is a no-op since `AbstractDependencyManager` doesn't implement `Serializable` — a comment explaining the intent (excluded from equals/hashCode) would be clearer

The LayeredMap cons-list design, the speculative pool-cache check in BfDependencyCollector, and the PathConflictResolver HashSet optimization are all well-motivated and correctly implemented.

cstamas:
What happens with DF collector?

gnodet:
The DF collector benefits from all four optimizations equally — it uses the same `DataPool` + `GraphKey` subtree caching as BF, and calls `deriveChildManager()` in its `doRecurse()` method identically.

Specifically:

1. **Instance reuse** — DF calls `deriveChildManager(context)` at every node in `doRecurse()`. Returning `this` when nothing changes avoids allocation and restores pool cache transparency (the `GraphKey` includes the manager, so identical-but-distinct instances were defeating cache lookups — same issue as BF).

2. **Key caching + MMap.equals() short-circuits** — `GraphKey.equals()` performance improves for both collectors since they share `DataPool`.

3. **Cons-list parent pointer** — the O(1) derive replaces the O(depth) ArrayList copy in the manager itself, independent of collector. DF's recursive traversal pattern actually makes it slightly *more* sensitive to `deriveChildManager()` cost than BF.

All tests pass with both collectors and all three manager subclasses. The PR body already notes this: "Full resolver test suite passes (BF + DF collectors, all three managers)."

PR #2018, opened 2026-07-29
Fix resource leak pattern in DependencyGraphParser.parseLiteral()

### LINKED ISSUE #1996: Minor resource leak in DependencyGraphParser.parseLiteral()
**Affected version:** HEAD

**File:** `maven-resolver-test-util/src/main/java/org/eclipse/aether/internal/test/util/DependencyGraphParser.java:134-138`

```java
BufferedReader reader = new BufferedReader(new StringReader(dependencyGraph));
DependencyNode node = parse(reader);
reader.close();  // never reached if parse() throws
```

While `StringReader.close()` is a no-op, the pattern is inconsistent with the rest of the codebase and represents a latent bug if the implementation changes.

Originally reported in #1944.

joshinii:
Wrap the BufferedReader in try-with-resources so it is closed even if parse() throws. StringReader.close() is currently a no-op, but this is to keep the method consistent with the rest of the codebase.

Fixes #1996

- [x] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

### DIFF
--- a/maven-resolver-test-util/src/main/java/org/eclipse/aether/internal/test/util/DependencyGraphParser.java
+++ b/maven-resolver-test-util/src/main/java/org/eclipse/aether/internal/test/util/DependencyGraphParser.java
@@ -172,19 +172,9 @@ public class DependencyGraphParser {
      * Parse the graph definition read from the given URL.
      */
     public DependencyNode parse(URL resource) throws IOException {
-        BufferedReader reader = null;
-        try {
-            reader = new BufferedReader(new InputStreamReader(resource.openStream(), StandardCharsets.UTF_8));
+        try (BufferedReader reader =
+                new BufferedReader(new InputStreamReader(resource.openStream(), StandardCharsets.UTF_8))) {
             return parse(reader);
-        } finally {
-            try {
-                if (reader != null) {
-                    reader.close();
-                    reader = null;
-                }
-            } catch (final IOException e) {
-                // Suppressed due to an exception already thrown in the try block.
-            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


gnodet (approved):
Correct and minimal conversion of `parseLiteral()` to try-with-resources, consistent with the same pattern already applied to `parseMultiResource()` in commit 16bbda7e. The change is trivially safe since the underlying `StringReader.close()` is a no-op, but it eliminates a latent leak pattern if the reader type ever changes.

One optional follow-up opportunity: the `parse(URL resource)` method (lines 175-189) still uses the pre-Java-7 try-finally-catch pattern with manual null checks and a swallowed `IOException`. Unlike `parseLiteral()`, that method wraps a real `InputStreamReader` over `resource.openStream()`, so converting it to try-with-resources would be more meaningful and would complete the consistency cleanup within this file.

gnodet (approved):
**Re-review** after new commit (34428516).

The new commit correctly implements the follow-up suggestion from the previous review, converting `parse(URL resource)` from the old try-finally-catch pattern to try-with-resources. This removes 12 lines of boilerplate and matches the identical pattern already used in `parseMultiResource()` (commit 16bbda7e).

Minor behavior improvement: the old code silently swallowed `IOException` from `reader.close()` in the finally block. The try-with-resources version properly adds close exceptions as suppressed exceptions, preserving diagnostic information.

All three reader-creating methods in `DependencyGraphParser` (`parseLiteral`, `parseMultiResource`, `parse(URL)`) now use try-with-resources consistently. Clean work.

cstamas (approved):

PR #2023, opened 2026-08-03
Extract included checksums from x-amz-meta-* headers

### LINKED ISSUE #2019: x-amz-meta-checksum
### Description

Maven Resolver currently supports extracting checksums from HTTP response headers using the following conventions:
```
x-checksum-sha1
x-checksum-md5
x-goog-meta-checksum-sha1
x-goog-meta-checksum-md5
```

AWS S3 and S3-compatible object storage services, such as Cloudflare R2, expose user-defined object metadata using the x-amz-meta-* prefix.

For example, an object uploaded with the metadata:

checksum-sha1 = c74edb60ca2a0b57ef88d9a7da28f591e3d4ce7b
checksum-md5 = 9ad0d8e3482767c122e85f83567b8ce6

is returned with the following response headers:

x-amz-meta-checksum-sha1: c74edb60ca2a0b57ef88d9a7da28f591e3d4ce7b
x-amz-meta-checksum-md5: 9ad0d8e3482767c122e85f83567b8ce6

Maven Resolver does not currently recognize these headers. As a result, when an artifact download is redirected to an S3-compatible presigned URL, Resolver downloads the artifact successfully but subsequently makes additional requests for checksum sidecar files. This happens even though the checksum is already available in the final artifact download response.

### Proposed enhancement

Extend XChecksumExtractor to recognize the following headers:

x-amz-meta-checksum-sha1
x-amz-meta-checksum-md5

The values should use the standard hexadecimal representation:

SHA-1: 40 hexadecimal characters
MD5: 32 hexadecimal characters

This enhancement would improve support for repositories backed by AWS S3 and S3-compatible object stores, including Cloudflare R2.

This proposal is specifically for S3 user-defined metadata headers:

x-amz-meta-checksum-sha1
x-amz-meta-checksum-md5

Relevant code
maven-resolver-impl/src/main/java/
org/eclipse/aether/internal/impl/transport/http/XChecksumExtractor.java

I would be happy to submit a pull request with the implementation, tests, and documentation updates.

iarvind:
AWS S3 exposes user defined object metadata under the x-amz-meta- prefix, so an object uploaded with the checksum-sha1 and checksum-md5 metadata keys is served back with the x-amz-meta-checksum-sha1 and x-amz-meta-checksum-md5 response headers. XChecksumExtractor did not know about them, so Resolver fell through to the Remote External strategy and issued a second request for the checksum sidecar file, even though the checksum was already in hand from the artifact response.

Teach the extractor that prefix. The three supported prefixes are now held in a list and tried in order, keeping the pre-existing precedence: the first prefix yielding any checksum wins, so a partial match on a later prefix cannot leak into an earlier one's result.

Stores that follow the same metadata convention, such as Cloudflare R2, MinIO and Backblaze B2, benefit from this as well.

Fixes #2019

- [x] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

### DIFF
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/transport/http/XChecksumExtractor.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/transport/http/XChecksumExtractor.java
@@ -21,7 +21,9 @@ package org.eclipse.aether.internal.impl.transport.http;
 import javax.inject.Named;
 import javax.inject.Singleton;
 
+import java.util.Arrays;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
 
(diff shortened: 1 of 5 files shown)


gnodet (approved):
Clean, well-structured PR. The refactoring of the duplicated prefix logic into a loop over a list of prefixes improves maintainability while preserving existing precedence semantics (Central-style > Google-style > Amazon-style). Test coverage is thorough — the `XChecksumExtractorTest` correctly uses case-insensitive `TreeMap` to mirror real HTTP transport behavior, and the new transport-level tests for Google-style headers fill a gap.

A couple of minor observations (non-blocking):

- `Arrays.asList` for the prefix list is fine here since the list is `private static final` and never exposed, but `Collections.unmodifiableList(Arrays.asList(...))` would make immutability explicit.
- The extractor currently handles only SHA-1 and MD5. Extending to SHA-256/SHA-512 (which have factories in the project) would be a natural follow-up for all three prefix styles.

Nice that this also incidentally fixes the misleading "Central style" comment on the Google `x-goog-meta-checksum-md5` header.

_On behalf of gnodet_

cstamas (approved):

PR #2027, opened 2026-08-04
Build against Maven 4.0.0-rc-6

slachiewicz:
4.0.0-rc-6 is released. This moves `maven4Version` to it and pins the same version explicitly in the Verify workflow, so CI no longer depends on the `maven4-version` default in maven-gh-actions-shared.

Verified: green on the fork before opening.

### DIFF
--- a/.github/workflows/maven-verify.yml
+++ b/.github/workflows/maven-verify.yml
@@ -29,5 +29,6 @@ jobs:
       ff-run: false
       ff-site-run: false
       maven4-enabled: true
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Straightforward version bump from rc-5 to rc-6, following the established pattern (rc-3→rc-4 in 31cecc91, rc-4→rc-5 in b9d00c85). The CI pin in `maven-verify.yml` is a good addition — keeps CI aligned with the project property. All downstream `${maven4Version}` references pick up the new value automatically.

_On behalf of gnodet_

PR #2032, opened 2026-08-04
Fix deadlock in IPC named lock access tests (#2030)

### LINKED ISSUE #2030: org.eclipse.aether.named.ipc.IpcNamedLockFactoryIT flaky
### Affected version

HEAD on Jenkins

### Bug description

Perhaps a thread-safety issue in the code?

```
exclusiveAccess(TestInfo) org.eclipse.aether.named.ipc.IpcNamedLockFactoryIT
Failing for the past 1 build since #1
Took 25 sec
Error Details

exclusiveAccess(org.junit.jupiter.api.TestInfo) timed out after 25 seconds

Stack Trace

java.util.concurrent.TimeoutException: exclusiveAccess(org.junit.jupiter.api.TestInfo) timed out after 25 seconds
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	Suppressed: java.lang.InterruptedException
		at java.base/java.lang.Object.wait0(Native Method)
		at java.base/java.lang.Object.wait(Object.java:366)
		at java.base/java.lang.Thread.join(Thread.java:2078)
		at java.base/java.lang.Thread.join(Thread.java:2154)
		at org.eclipse.aether.named.ipc.NamedLockFactoryTestSupport.exclusiveAccess(NamedLockFactoryTestSupport.java:150)
		at java.base/java.lang.reflect.Method.invoke(Method.java:580)
```

elharo:
Closes #2030

## Problem

org.eclipse.aether.named.ipc.IpcNamedLockFactoryIT.exclusiveAccess is flaky on Jenkins, timing out after 25 s with the main thread stuck in t1.join().

The root cause is a deadlock in the Access handshake shared by sharedAccess, exclusiveAccess and mixedAccess. When both competing threads fail to acquire the lock within the 100 ms acquisition timeout, both threads count down the loser latch and then block forever on winner.await(), because the winners latch is never counted down. The test then hangs until the JUnit @Timeout fires.

This happens under load when the in-process IPC server thread is starved for more than 100 ms, so neither acquisition completes in time. Earlier fixes only raised the overall test timeout (5 s to 15 s to 25 s), which converts the deadlock into a longer hang without fixing it.

## Fix

- Raise the lock acquisition timeout in Access from 100 ms to 1 s so at least one thread reliably wins even on a loaded machine. Mutual exclusion is still verified: the loser is queued behind the winner on the server and cannot be granted while the winner holds, so its failure is still caused by the winner holding the lock.
- Bound the winner/loser handshake awaits and the test-side joins and latch awaits, and assert the expected outcome. In the (now unlikely) case both threads lose, the test fails fast with a clear message instead of hanging for the full @Timeout.

Verified by running IpcNamedLockFactoryIT 21 consecutive times (15 normal + 6 under simulated CPU load), all passing.

### DIFF
--- a/maven-resolver-named-locks-ipc/src/test/java/org/eclipse/aether/named/ipc/NamedLockFactoryTestSupport.java
+++ b/maven-resolver-named-locks-ipc/src/test/java/org/eclipse/aether/named/ipc/NamedLockFactoryTestSupport.java
@@ -131,10 +131,10 @@ public abstract class NamedLockFactoryTestSupport {
         Thread t2 = new Thread(new Access(namedLockFactory, keys, true, winners, losers));
         t1.start();
         t2.start();
-        t1.join();
-        t2.join();
-        winners.await();
-        losers.await();
+        t1.join(5000);
+        t2.join(5000);
+        assertTrue(winners.await(5, TimeUnit.SECONDS), "expected both threads to win");
+        assertTrue(losers.await(5, TimeUnit.SECONDS), "expected no loser");
     }
 
     @Test
@@ -147,10 +147,10 @@ public abstract class NamedLockFactoryTestSupport {
         Thread t2 = new Thread(new Access(namedLockFactory, keys, false, winners, losers));
         t1.start();
         t2.start();
-        t1.join();
-        t2.join();
-        winners.await();
-        losers.await();
+        t1.join(5000);
+        t2.join(5000);
+        assertTrue(winners.await(5, TimeUnit.SECONDS), "expected a winner");
+        assertTrue(losers.await(5, TimeUnit.SECONDS), "expected a loser");
     }
 
     @Test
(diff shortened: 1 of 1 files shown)


gnodet (approved):
Well-diagnosed fix for the real root cause — both threads failing the 100ms lock acquisition under CPU starvation, leaving both stuck in unbounded `winner.await()`. The two-pronged approach (increasing acquisition timeout to 1s + bounding all internal awaits/joins with timeouts and assertions) is the right fix over the previous band-aid `@Timeout` increases.

The PR description is excellent — clearly documents the root cause, explains why previous fixes were insufficient, and provides verification evidence.

One observation: `NamedLockFactoryAdapterTestSupport.java` has the same unbounded `join()`/`await()` pattern with a 100ms acquisition timeout (`ADAPTER_TIME = 100L`). It may be susceptible to the same deadlock — worth a follow-up pass.

_On behalf of gnodet_

PR #2045, opened 2026-08-08
Fix #1994: log exceptions from IpcNamedLock.tryUnlock()

### LINKED ISSUE #1994: Empty catch swallows exceptions in IpcNamedLock.tryUnlock()

elharo:
Fixes #1994

`IpcNamedLock.tryUnlock()` was silently discarding all exceptions from `client.unlock()` during timeout cleanup, masking real IO/connectivity issues.

**Fix:** log the failure at WARN level instead of swallowing it. The exception is still not propagated, preserving the best-effort cleanup semantics that must not crash the calling thread.

**Test:** added `IpcNamedLockTest.unlockFailureDuringTimeoutCleanupIsLogged`, which mocks `IpcClient` so that `lock()` times out and `unlock()` throws, then asserts the unlock failure is logged. The test fails against the old code (no log output) and passes with the fix.

### DIFF
--- a/maven-resolver-named-locks-ipc/src/test/java/org/eclipse/aether/named/ipc/IpcNamedLockTest.java
+++ b/maven-resolver-named-locks-ipc/src/test/java/org/eclipse/aether/named/ipc/IpcNamedLockTest.java
@@ -1,0 +1,67 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.eclipse.aether.named.ipc;
+
+import java.io.ByteArrayOutputStream;
+import java.io.PrintStream;
+import java.nio.charset.StandardCharsets;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+import org.eclipse.aether.named.NamedLockKey;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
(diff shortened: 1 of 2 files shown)


maven-resolver-named-locks-ipc/src/test/java/org/eclipse/aether/named/ipc/IpcNamedLockTest.java
```diff
@@ -0,0 +1,67 @@
...
+import org.eclipse.aether.named.NamedLockKey;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+class IpcNamedLockTest {
```
gnodet, line 38: Minor: every other test class in this module and project uses `public class`. Consider:
```suggestion
public class IpcNamedLockTest {
```

gnodet (approved):
Clean, minimal fix that adds WARN-level logging for a previously silent exception catch in `IpcNamedLock.tryUnlock()`. The production change is correct — a single `logger.warn()` call using the inherited SLF4J logger, with the exception passed for stack trace logging. WARN level is appropriate for a secondary failure during timeout cleanup.

Two minor style notes (non-blocking):

1. **Test class visibility**: `IpcNamedLockTest` is declared package-private while every other test class in the project uses `public class`. Consider adding `public` for consistency.

2. **stderr capture pattern**: The test captures `System.err` to verify `slf4j-simple` output. This is pragmatic given the project's test dependencies, but worth noting it's not thread-safe for parallel test execution and depends on `slf4j-simple` default configuration.

elharo:
https://github.com/apache/maven-resolver/blob/89f2bc1b284e0024c3a4121f55dd0f4709d35009/maven-resolver-impl/src/test/java/org/eclipse/aether/internal/impl/DefaultRepositorySystemReentrancyTest.java seems to have broken the CI and went in without a pull request. We really need to block this at the repo level.

elharo:
The Verify failures are pre-existing on master, not caused by this change.

Master commit 89f2bc1b has the identical 18 failing Verify jobs (e.g. windows-latest jdk-21-zulu 4.0.0-rc-6 fails on the same lines). The failure is a test-compile error in the *maven-resolver-impl* module:

```
DefaultRepositorySystemReentrancyTest.java:[361,17] incompatible types: no instance(s) of type variable(s) T exist so that java.util.List<T> conforms to java.util.Map<java.lang.String,org.eclipse.aether.spi.validator.ValidatorFactory>
```

introduced by commit f164aa90 `Enh: Allow validator factory to abstain (#2008)`.

This PR only modifies the `maven-resolver-named-locks-ipc` module (IpcNamedLock + a regression test). All 27 unit tests + ITs for that module pass locally (`-Prun-its verify`), and `spotless:check` is green.

kwin:
Master has been fixed meanwhile. Can you rebase?

PR #2056, opened 2026-08-09
Fix #1998: use try-with-resources in test file utilities

### LINKED ISSUE #1998: Leaky stream handling in test utilities (TestFileUtils, TestFileProcessor)
**Affected version:** HEAD

**Files:**
- `maven-resolver-test-util/src/main/java/org/eclipse/aether/internal/test/util/TestFileUtils.java:172-325`
- `maven-resolver-test-util/src/main/java/org/eclipse/aether/internal/test/util/TestFileProcessor.java:63-151`

Methods manually close streams in try-catch-finally blocks instead of using try-with-resources. If an explicit `close()` call throws before the finally block, open resources are leaked. Functionally correct in happy-path but fragile.

Originally reported in #1944.

elharo:
Fixes #1998

Convert manual `try-catch-finally` stream handling to try-with-resources in the test utilities:

- `TestFileUtils`: `copyFile`, `readBytes`, `writeBytes`, `readProps`, `writeProps`
- `TestFileProcessor`: `write(File, String)`, `write(File, InputStream)`, `copy(File, File, ProgressListener)`

Resources created inside each method are now closed reliably (with close exceptions properly suppressed), eliminating the leak when an explicit `close()` throws before the `finally` block. The caller-supplied `InputStream` in `TestFileProcessor.write/copy` remains intentionally unclosed, matching prior behavior.

No new tests required.

### DIFF
--- a/maven-resolver-test-util/src/main/java/org/eclipse/aether/internal/test/util/TestFileProcessor.java
+++ b/maven-resolver-test-util/src/main/java/org/eclipse/aether/internal/test/util/TestFileProcessor.java
@@ -63,23 +63,9 @@ public class TestFileProcessor implements FileProcessor {
     public void write(File file, String data) throws IOException {
         mkdirs(file.getParentFile());
 
-        FileOutputStream fos = null;
-        try {
-            fos = new FileOutputStream(file);
-
+        try (FileOutputStream fos = new FileOutputStream(file)) {
             if (data != null) {
                 fos.write(data.getBytes(StandardCharsets.UTF_8));
-            }
-
-            fos.close();
-            fos = null;
-        } finally {
-            try {
-                if (fos != null) {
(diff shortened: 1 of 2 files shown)


gnodet (approved):
**LGTM** — Clean, correct conversion of 8 methods across two test utility files from manual try-catch-finally to try-with-resources. All conversions are behavior-preserving:

- The `mkdirs` reordering in `copy`/`copyFile` is safe and actually a slight improvement (no resources open when `mkdirs` runs).
- The caller-supplied `InputStream` in `TestFileProcessor.write(File, InputStream)` remains correctly unmanaged.
- Exception suppression via try-with-resources is strictly better for diagnostics than the original swallowed close exceptions.
- Net -130 lines of boilerplate with no functional risk.

CI should pass before merge.

PR #2057, opened 2026-08-09
Reconcile the two download pages and modernise them

slachiewicz:
This repository carries **two** download pages from different generations: the root page in the current form, and `maven-resolver-demos/src/site/xdoc/download.xml.vm` still on the 2014 **mirror-selection form**, with its checksum and KEYS links pointing at `www.apache.org/dist` (a 301 to `downloads.apache.org` for years now).

### The module page advertised a file that does not exist

It offered `maven-resolver-demos-${project.version}-source-release.zip`. **That has never been released.** `maven-resolver-demos` ships inside the `maven-resolver` source release and has no source bundle of its own — `https://downloads.apache.org/maven/resolver/` confirms it.

Rather than making a wrong page prettier, the module page now points at the `maven-resolver` bundle that actually exists, and says so in a sentence, so a reader is not surprised by a page titled after the module offering an archive named after its parent. The URL stays live either way — it is reachable today and deleting it was not on the table.

### Also

- Both pages brought to the form the rest of the Maven estate uses (finishing the sweep started in 2023 that stopped part-way through).
- `www.apache.org/dist` → `downloads.apache.org` for checksums and KEYS.
- **The ASF licence headers are left byte-for-byte as they were.** My first pass spliced in a canonical header and silently downgraded an `https://` LICENSE-2.0 URL to `http://`; the generator now preserves each file's existing preamble and never rewrites it.
- The root page diff is whitespace only.

### Verification

XML well-formedness checked on both files, and the new module page rendered through a site build to confirm the headings, the added sentence and the download row come out right. Local checks only — not claiming CI is green; draft until it is.

### DIFF
--- a/maven-resolver-demos/src/site/xdoc/download.xml.vm
+++ b/maven-resolver-demos/src/site/xdoc/download.xml.vm
@@ -23,104 +23,54 @@ under the License.
   <properties>
     <title>Download ${project.name} Source</title>
   </properties>
+
   <body>
     <section name="Download ${project.name} ${project.version} Source">
 
-      <p>${project.name} ${project.version} is distributed in source format. Use a source archive if you intend to build
-      ${project.name} yourself. Otherwise, simply use the ready-made binary artifacts from central repository.</p>
+      <p><strong>${project.name} ${project.version}</strong> is distributed in source format.</p>
 
-      <p>You will be prompted for a mirror - if the file is not found on yours, please be patient, as it may take 24
-      hours to reach all mirrors.<p/>
+      <p>Use a source archive if you intend to build <strong>${project.name}</strong> yourself.</p>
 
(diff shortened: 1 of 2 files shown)


elharo (approved):
This is a step up from than what we have here now, so let's merge,  but does still need an upgrade to simplified technical English

gnodet (approved):
**LGTM** — Clean documentation improvement that:

- Fixes a factual error on the demos download page (pointed to a non-existent `maven-resolver-demos` source release archive)
- Removes obsolete ASF mirror-selection infrastructure (`[preferred]` redirects)
- Modernizes all URLs to current ASF domains (`dlcdn.apache.org` for downloads, `downloads.apache.org` for checksums/signatures)
- Root page change is whitespace-only
- Both pages now share consistent structure and point to the correct `maven-resolver` source release

PR #2060, opened 2026-08-18
Fix #2058: Add links to configuration page

### LINKED ISSUE #2058: Add javadoc links from configuration page

The configuration page at <https://maven.apache.org/resolver/configuration.html> should emit javadoc links for complex custom types (like for `aether.transport.http.version`).

This can be achieved with the class https://github.com/apache/maven-plugin-tools/blob/master/maven-plugin-tools-api/src/main/java/org/apache/maven/tools/plugin/javadoc/JavadocLinkGenerator.java.
The Maven Plugin Tools also emit javadoc links for goal reports.
References to other fields which are also configuration elements should be emitted as anchor links pointing to the correct fragment within the table.

Jamison929611:
## Summary

- generate local Javadoc links for resolved custom configuration types, including nested types
- add stable anchors for configuration keys and link configuration-field references to those anchors
- cover intermediate metadata and rendered Markdown with focused doclet tests

Fixes #2058

## Verification

- [x] Focused ConfigurationCollectorDocletTest: 4 tests, 0 failures or errors
- [x] Reactor package through maven-resolver-tools with the focused test and Checkstyle, Spotless, and RAT
- [x] spotless:check and checkstyle:check for maven-resolver-tools
- [x] git diff --check
- [ ] Full mvn verify: the broader Windows run is blocked in the unchanged maven-resolver-transport-file tests because the current account cannot create symbolic links
- [ ] Integration tests with the run-its profile were not run

## Contribution checklist

- [x] This pull request addresses one issue without unrelated changes.
- [x] The commit has a meaningful subject and body.
- [x] Unit tests cover the behavioral changes.
- [x] I hereby declare this contribution to be licenced under the Apache License Version 2.0, January 2004.


gnodet (approved):
Clean addition of Javadoc type links and cross-reference anchor links for the configuration page, correctly addressing #2058. The implementation is well-structured:

- `getJavadocUrl` correctly traverses `PackageElement` to handle nested types
- `resolveReferencedField` + `getConfigurationKey` properly resolve `{@link #FIELD}` references to configuration key names
- HTML anchors use dot-separated configuration key identifiers, which are valid `id` values
- Velocity templates correctly treat empty strings as falsy for conditional rendering
- Tests cover both metadata extraction and full-pipeline rendered Markdown output

One minor observation: the `!typeElement.getQualifiedName().toString().startsWith("java.")` filter excludes JDK types from Javadoc links, but `javax.*`/`jakarta.*`/third-party types would still get a relative `apidocs/` link. In practice this is a non-issue since configuration types here are internal enums/classes, but worth noting for future-proofing.

Jamison929611:
Thanks for the review and approval. The PR appears ready from my side. Could you please merge it when you have a chance? If any additional checks or changes are still required, please let me know. Thanks!

maven-resolver-tools/src/main/java/org/eclipse/aether/tools/ConfigurationCollectorDoclet.java
```diff
@@ -757,11 +772,39 @@ private String getConfigurationType(DocTreePath path, Map<String, UnknownBlockTa
...
+            return null;
+        }
+        if (mode == Mode.MAVEN) {
+            return getAnnotation(field, MAVEN_CONFIG_ANNOTATION) != null ? key : null;
+        }
+        DocCommentTree docComment = docTrees.getDocCommentTree(field);
+        return docComment != null && collectBlockTags(docComment).containsKey("configurationSource") ? key : null;
+    }
+
+    private static String getJavadocUrl(TypeElement type) {
```
kwin: Please use https://github.com/apache/maven-plugin-tools/blob/master/maven-plugin-tools-api/src/main/java/org/apache/maven/tools/plugin/javadoc/JavadocLinkGenerator.java. The format of deep links (e.g. method references) differs depending on the javadoc version.

maven-resolver-tools/src/main/java/org/eclipse/aether/tools/ConfigurationCollectorDoclet.java
```diff
@@ -757,11 +772,39 @@ private String getConfigurationType(DocTreePath path, Map<String, UnknownBlockTa
...
+        Element enclosing = type;
+        while (!(enclosing instanceof PackageElement)) {
+            enclosing = enclosing.getEnclosingElement();
+        }
+        String packageName = ((PackageElement) enclosing).getQualifiedName().toString();
+        String qualifiedName = type.getQualifiedName().toString();
+        String className = packageName.isEmpty() ? qualifiedName : qualifiedName.substring(packageName.length() + 1);
+        String packagePath = packageName.replace('.', '/');
+        return "apidocs/" + (packagePath.isEmpty() ? "" : packagePath + "/") + className + ".html";
```
kwin: The prefix should be configurable.

maven-resolver-tools/src/main/java/org/eclipse/aether/tools/ConfigurationCollectorDoclet.java
kwin: Please support all types of links (not only to types, but also to elements (constructors, method, fields)). Also there needs to be some basic validation as this may also refer to types not contained in the Javadoc (e.g. transitive dependencies or JRE built in packages/types/elements). Also the module name potentially needs to be considered.

kwin (changes requested):
Looks good as first draft, however this requires some more work to distinguish between internal and external links (and potentially skip unavailable javadoc links) as well as supporting links to elements and package/modules next to the one for types.

maven-resolver-tools/src/main/java/org/eclipse/aether/tools/ConfigurationCollectorDoclet.java
Jamison929611: Implemented in 0d7e9e95. Link construction now uses JavadocLinkGenerator for type, member, package, and module references. The tests cover both Javadoc 8 and Javadoc 10+ method/constructor fragment formats.

maven-resolver-tools/src/main/java/org/eclipse/aether/tools/ConfigurationCollectorDoclet.java
Jamison929611: Implemented in 0d7e9e95. The internal base URL is configurable with --internal-javadoc-url, the internal Javadoc version with --internal-javadoc-version, and external bases with repeatable --external-javadoc-url. The defaults preserve apidocs/ with Javadoc 21.

maven-resolver-tools/src/main/java/org/eclipse/aether/tools/ConfigurationCollectorDoclet.java
Jamison929611: Implemented in 0d7e9e95. References now support types, fields, methods, constructors, packages, and modules. Internal links require visible declarations backed by the project source tree; external sites are validated through element-list/package-list, including module-aware layouts, and unavailable sites or packages fall back to plain text. Configuration-key anchors still take priority. Tests cover internal and external modules, external member fragments, unavailable packages, and non-public members.

Jamison929611:
@kwin The requested changes are implemented in 0d7e9e95, and each inline review thread has been updated with details. Could you please re-review when convenient? The PR is now mergeable; the Verify workflow is awaiting maintainer approval.

maven-resolver-tools/pom.xml
```diff
@@ -162,6 +162,46 @@
...
+        </exclusion>
+        <exclusion>
+          <groupId>org.codehaus.plexus</groupId>
+          <artifactId>plexus-xml</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.maven</groupId>
+      <artifactId>maven-settings</artifactId>
```
gnodet, line 204: Minor: `maven-settings` isn't directly imported by any source file — it appears to be needed at runtime because `maven-core` (its usual provider) is excluded from `maven-plugin-tools-api` above. A brief comment explaining this transitive dependency rationale would help future maintainers understand why it's explicitly declared.

gnodet (commented):
**Re-review (2nd pass):** Substantial expansion addressing all three points of @kwin's feedback:

1. ✅ Uses `JavadocLinkGenerator` from `maven-plugin-tools-api` for version-aware link formatting (Javadoc 8 vs 10+)
2. ✅ Internal/external link distinction with configurable base URLs (`--internal-javadoc-url`, `--external-javadoc-url`)
3. ✅ Supports types, fields, methods, constructors, packages, and modules with visibility filtering

The priority ordering in `visitLink` is correct — configuration-key cross-references take precedence over Javadoc links, which fall back to plain text. Test coverage is comprehensive (6 test methods).

One minor observation below about dependency documentation.

maven-resolver-tools/pom.xml
Jamison929611: Documented in 79a8ef4d: JavadocLinkGenerator's constructor signature references Settings, so maven-settings is declared directly because maven-core is excluded.

gnodet (approved):
**Re-review (3rd pass):** The maven-settings dependency comment addresses the finding from the previous review. The explanation is clear — `JavadocLinkGenerator`'s constructor references `Settings`, and `maven-core` (its usual transitive provider) is excluded above.

All previously raised items are resolved. The implementation comprehensively addresses #2058 with proper internal/external link distinction, configurable URLs, and support for all element kinds.

maven-resolver-tools/pom.xml
kwin, line 273: This requires additional arguments now to actually consider the javadoc. Also there is a ordering problem as the aggregate javadoc is created after populating the configuration.

kwin (changes requested):

maven-resolver-tools/pom.xml
Jamison929611: Addressed in 21bcb9ac. The render-configuration-page execution now passes the internal Javadoc URL, the Javadoc version, and the Java 21 external API URL. The shared POM properties keep those values aligned with maven-javadoc-plugin. For the ordering issue, configuration collection does not require the aggregate output to exist: internal references are validated against the discovered src/main/java roots while emitting apidocs/ links, so the aggregate Javadocs can still be generated later by the site lifecycle. The machine-readable properties/YAML execution remains unchanged.

Jamison929611: @kwin, could you please take another look when convenient?

gnodet (approved):
**Re-review (4th pass):** Both of @kwin's concerns are properly addressed:

1. ✅ **Additional arguments**: the `render-configuration-page` execution now passes `--internal-javadoc-url`, `--internal-javadoc-version`, and `--external-javadoc-url` with values from shared POM properties (`javadocVersion`, `javadocJavaApiUrl`) that stay in sync with the `maven-javadoc-plugin` configuration.

2. ✅ **Ordering problem**: the doclet discovers `src/main/java` source roots from the processed files (via `findMainJavaSourceRoot`) and validates internal types against those source trees. This means the configuration page can be generated before the aggregate Javadoc exists — the relative `apidocs/` links resolve correctly once the site lifecycle generates the aggregate output.

The shared POM properties eliminate the previously hardcoded `"21"` and Java API URL, keeping both the Javadoc plugin and the configuration doclet aligned.

kwin (changes requested):
I still see no javadoc links in `maven-resolver/target/generated-site/markdown/configuration.md`. Most probably this is related to the fact that the referenced classes are from a foreign module (i.e. not part of `src/main/java` within `maven-resolver-tools`). The logic to check if a type is part of the aggregated javadoc or not need to be reworked.

Jamison929611:
Fixed in 4f21cd46.

The missing link was caused by internalJavadocSourceRoots being derived only from source files that declare configuration keys. That excluded sibling reactor modules whose types are part of the aggregate Javadoc.

The collector now:
- keeps the doclet compilation units limited to marker-bearing configuration sources;
- derives aggregate Javadoc membership from all reactor src/main/java source roots;
- detects those roots using Path operations so the collection also works on Windows.

I added an end-to-end regression test where a configuration source in one module references ConfigurationProperties.HttpVersion from a sibling module. The targeted reactor package run passes with 9 tests, and the real generated target/generated-site/markdown/configuration.md now contains both the ConfigurationProperties.HttpVersion Javadoc link and the #aether.transport.http.version configuration anchor.

@kwin, could you please take another look?

kwin (changes requested):
@Jamison929611 That logic does not work for configuration pages being created for Maven itself: https://github.com/apache/maven/blob/7dc59176c0b0cb09198bef11425b97ece582c606/pom.xml#L1067

gnodet (approved):
**Re-review (5th pass):** The cross-module source-root detection fix correctly addresses @kwin's feedback.

The core change: `javaSourceFiles` now collects ALL `.java` files under `src/main/java` across the reactor (not just configuration-marker-bearing sources), and `internalJavadocSourceRoots` are derived from this broader set. This means types from sibling modules (e.g., `ConfigurationProperties.HttpVersion` from `maven-resolver-api`) are correctly identified as internal and receive proper `apidocs/` links.

The `isInternalJavadocElement()` fallback chain is sound: tries `docTrees.getPath()` first (for types in doclet compilation units), then falls back to `isTypeInInternalSourceTree()` (filesystem check across all discovered source roots).

The new test `linksTypesFromSiblingAggregateJavadocSourceRoots` directly validates the cross-module scenario with a configuration source in one module referencing a type from a sibling module.

Good defensive improvements: `Path.endsWith()` replacing string-based `/src/main/java/` check for Windows compatibility, and `replace('\\', '/')` for path normalization.

Jamison929611:
Addressed in 56e23235.

I verified the Maven execution you linked: it intentionally keeps api/maven-api-core as the positional configuration source root, while its aggregate Javadocs cover the wider reactor.

CollectConfiguration now exposes a repeatable --internal-javadoc-source-tree option. Configuration declarations are still collected only below the positional root; source roots used to validate aggregate Javadoc links are discovered independently from the configured tree(s). If the option is omitted, it defaults to the positional root, preserving the existing Resolver invocation.

For the Maven execution, the existing positional argument can remain unchanged and the page execution can add:

~~~xml
<argument>--internal-javadoc-source-tree=${project.basedir}</argument>
~~~

The regression test now invokes the public CLI with a narrow configuration-module positional root and a separate reactor-wide source tree, rather than placing both modules below the positional root. The full ConfigurationCollectorDocletTest suite passes (9/9), the 26-module package build succeeds, and Resolver's real generated configuration page still contains the HttpVersion Javadoc link and the configuration anchor using the default behavior.

@kwin, could you please take another look?

gnodet (approved):
**Re-review (6th pass):** The repeatable `--internal-javadoc-source-tree` option is the right approach for the Maven cross-reactor case @kwin flagged — it separates the configuration declaration root (positional argument) from the broader aggregate Javadoc source trees, which is exactly what Maven's own configuration page build needs.

maven-resolver-tools/src/main/java/org/eclipse/aether/tools/CollectConfiguration.java
```diff
@@ -138,6 +138,13 @@ public enum Mode {
             description = "External Javadoc base URLs used for validated links")
     protected List<URI> externalJavadocUrls = new ArrayList<>();
 
+    @CommandLine.Option(
+            names = "--internal-javadoc-source-tree",
+            paramLabel = "directory",
+            description =
+                    "Directory trees searched for source roots included in aggregate Javadocs; defaults to the source processing root")
```
kwin: This deserves some more explanation. How exactly are the directories traversed to find source roots?

Jamison929611: Clarified in 9620deeb. Each supplied tree is traversed recursively with Files.walk (without following symbolic links by default). Java files are retained only when their nearest ancestor path ends in src/main/java, and those distinct ancestors become the aggregate Javadoc source roots. The option may be repeated; when omitted, discovery uses the positional source processing tree. The CLI help now documents the recursive traversal, ancestor matching, deduplication, repeatability, and default.

cstamas:
Where are we with this? @kwin you still have "req for change" set

cstamas (approved):

gnodet:
## Review findings

I did a thorough review of this PR, built the website, and inspected the generated configuration page. Here's what I found:

### Generated configuration page ✅
The configuration page generates correctly with:
- Internal `apidocs/` links for resolver types (ConfigurationProperties.HttpVersion, RepositorySystemSession, DependencyNode, etc.)
- External `docs.oracle.com` links for JDK types (160 links)
- Stable anchors for all configuration keys
- Cross-reference links between configuration keys
- All 9 tests pass

### Issues found and fixed in https://github.com/apache/maven-resolver/pull/2060/commits/493d70f1d6814818e74398de541ff3a1480f957a

1. **`visitLink` renders reference instead of label** — The `visitLink` method renders `node.getReference()` (the link target) instead of `node.getLabel()` (the explicit label text). For `{@link Type custom label}`, it shows "Type" instead of "custom label". While current configuration sources don't use explicit labels, this is a correctness bug.

2. **Markdown emphasis around HTML content** — The `configuration.md.vm` template wraps HTML deprecated content in Markdown emphasis markers (`*...*`), mixing Markdown and HTML. Fixed by using `<em>` instead.

3. **Missing trailing newline** in `package-info.java` test fixture.

### Architecture assessment ✅
- The `isInternalJavadocElement` logic correctly handles nested types, visibility filtering, and source-tree-based resolution
- The `JavadocLinkGenerator` integration is well done — version-aware link formatting (JDK 8 hyphen style vs JDK 10+ parenthesis style)
- Error reporting with `DocTreePathAwareRuntimeException` provides precise source-location-aware diagnostics
- Cross-module detection via `--internal-javadoc-source-tree` is clean and avoids coupling to the Maven reactor
- Test coverage is solid across internal/external links, cross-module resolution, module elements, and error cases

cstamas:
Last nit: to be merged, we need checked the "I hereby declare this contribution to be licenced under the Apache License Version 2.0, January 2004." checkbox. Thanks!

Jamison929611:
@cstamas The Apache License declaration is now checked in the PR description. The contribution checklist is complete.

PR #2079, opened 2026-08-30
fix: validate post-transformation path shapes and assert containment at resolve sinks

gnodet:
## Summary

Fixes 4 findings from the maven-resolver security audit (scan-maven-resolver-20260811):

| Finding | Severity | Description |
|---------|----------|-------------|
| f001 | **HIGH** | Leading-dot and colon coordinates bypass `validatePathComponent`: absolute-path writes outside local repo and cross-host fetches |
| f003 | MEDIUM | `FileTransporter` traversal guard misses absolute paths and Windows separators |
| f010 | MEDIUM | Repository id `".."` escapes origin-aware trusted-checksums and split-LRM directories |
| f017 | MEDIUM | Default file-lock name mapper builds lock paths from raw wire-supplied coordinates |

**Root cause:** Strings from remote repositories (artifact coordinates, repository ids) are validated in their pre-transformation shape and then passed to `Path.resolve()`, which returns absolute arguments unchanged. No sink asserts containment.

**Fix:** Validate post-replacement path shape, reject `..`/`.`/`:` segments, and assert `normalize().startsWith(base)` containment at every resolve sink.

## Test plan
- [ ] Existing tests pass
- [ ] New test cases for leading-dot groupId, colon groupId, ".." repository id
- [ ] FileTransporter containment check tested
- [ ] Lock path containment tested

Generated with [Claude Code](https://claude.com/claude-code)

PR #2083, opened 2026-08-30
fix: atomic file publication and authenticated IPC lock daemon

gnodet:
## Summary

Fixes 3 findings from the maven-resolver security audit (scan-maven-resolver-20260811):

| f007 | MEDIUM | Windows final move is truncate-and-copy, not atomic; truncated artifact can survive as trusted |
| f020 | LOW | IPC lock daemon protocol is unauthenticated; bootstrap secret leaks via argv |
| f025 | LOW | IPC lock key digest concatenates names without separators |

**Root cause:** The concurrency layer has spots where its own guarantee is violated: Windows publication is not atomic, and the IPC lock daemon trusts every connection and digests keys ambiguously.

**Fix:** Atomic rename on Windows with retry, authenticated IPC daemon with SecureRandom tokens and peer credentials, unambiguous length-prefixed key digest.

## Test plan
- [ ] Existing tests pass
- [ ] Windows atomic move tested
- [ ] IPC daemon authentication tested
- [ ] Key digest uniqueness tested

### DIFF
--- a/maven-resolver-named-locks-ipc/src/main/java/org/eclipse/aether/named/ipc/IpcClient.java
+++ b/maven-resolver-named-locks-ipc/src/main/java/org/eclipse/aether/named/ipc/IpcClient.java
@@ -26,6 +26,7 @@ import java.io.File;
 import java.io.FileWriter;
 import java.io.IOException;
 import java.io.InterruptedIOException;
+import java.io.OutputStream;
 import java.io.PrintWriter;
 import java.io.RandomAccessFile;
 import java.net.SocketAddress;
@@ -35,9 +36,11 @@ import java.nio.channels.Channels;
 import java.nio.channels.FileLock;
 import java.nio.channels.ServerSocketChannel;
 import java.nio.channels.SocketChannel;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.security.SecureRandom;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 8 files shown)


cstamas (approved):

gnodet (commented):
Excellent security-hardening PR — defense-in-depth across the IPC lock protocol with comprehensive tests.

## Strengths

- **Bootstrap token via stdin** instead of argv prevents exposure through `/proc/<pid>/cmdline` — the right approach
- **`SecureRandom` over `Random`** for the bootstrap token — required for a security credential
- **Per-connection context scoping** is a solid privilege-separation fix: one client can no longer close or acquire on another client's lock contexts
- **Netstring-like lock key digest** prevents boundary-shifted collisions (`["foo","bar"]` vs `["foob","ar"]`) — clean and correct
- **Bounded preallocation** (`Math.min(sz, 1024)`) prevents memory exhaustion from malicious wire-supplied sizes
- **Socket file permissions** (owner-only rwx) on Unix domain sockets add a second layer beyond the token
- **`retryingMove` never-truncate contract** eliminates the torn-read window where a concurrent reader or process kill could observe a truncated file at the final path after checksum validation
- Test coverage is thorough: `IpcServerAccessControlTest` validates context isolation and token enforcement, `IpcNamedLockFactoryTest` covers digest collision resistance, `PathProcessorSupportTest` covers retry, never-truncate guarantee, and staging fallback

## One observation

**`@since 2.0.22` tags** — Same question as the sibling PRs in this security audit batch (#2079–#2086): the `maven-resolver-2.0.22` tag already exists in the repository, so these new API additions (`IpcServer(SocketFamily, String)` constructor at line 232, `restrictToOwner` in `SocketFamily`) should use `@since 2.0.23` (or whatever the next release version will be).

Overall this is high-quality, well-thought-out security work. Well done!

cstamas (approved):

PR #2087, opened 2026-08-30
Validate coordinate components in the Maven 2 layout and local path composer

slachiewicz:
`Maven2RepositoryLayoutFactory` composes a remote URI from artifact and metadata coordinates by
string concatenation, without validating the components first, although `PathUtils` already provides
`validateArtifactComponents` and `validateMetadataComponents` for exactly that and other callers use
them. This calls them in `getLocation(Artifact, boolean)` and `getLocation(Metadata, boolean)`.

`DefaultLocalPathComposer` additionally checks its own result: the composed path is required to be a
relative path with no empty or parent-reference segments, so that resolving it against the local
repository base directory always yields a path inside it. Component validation on entry should
already guarantee that, so this is a cheap check on the composed result rather than a replacement
for it.

No public API changes; both are internal implementation classes.

### DIFF
--- a/maven-resolver-util/src/main/java/org/eclipse/aether/util/PathUtils.java
+++ b/maven-resolver-util/src/main/java/org/eclipse/aether/util/PathUtils.java
@@ -87,9 +87,31 @@ public final class PathUtils {
         if (value != null && !value.isEmpty()) {
             // Important: "equals .." and not "contains ..", as if escape attempted, it will contain path separators
             // OTOH: version "1.." is valid version string!
-            if (value.equals("..") || value.contains("/") || value.contains("\\")) {
+            // Colon is not a valid character in a coordinate component.
+            if (value.equals("..") || value.contains("/") || value.contains("\\") || value.contains(":")) {
                 throw new IllegalArgumentException(
-                        "Invalid " + label + ": must not contain '..', '/' or '\\': " + value);
+                        "Invalid " + label + ": must not contain '..', '/', '\\' or ':': " + value);
+            }
+        }
+    }
+
+    /**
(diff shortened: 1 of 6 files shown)


maven-resolver-util/src/main/java/org/eclipse/aether/util/PathUtils.java
```diff
@@ -87,9 +87,31 @@ public static void validatePathComponent(String value, String label) {
...
+    }
+
+    /**
+     * Validates a coordinate component that is expanded into multiple path segments by replacing each dot with a
+     * path separator, like the group ID is. Beside the checks done by
+     * {@link #validatePathComponent(String, String)}, it rejects values containing empty dot-separated segments
+     * (leading, trailing or consecutive dots), as the expansion of such values does not compose a valid
+     * relative path.
+     *
+     * @since 2.0.22
```
gnodet: Minor: `@since 2.0.22` is incorrect — tag `maven-resolver-2.0.22` has already been released without this method. Should be `@since 2.0.23`.

```suggestion
     * @since 2.0.23
```

gnodet (approved):
Solid security hardening that closes a real path traversal gap via groupIds with empty dot-segments. The layered validation approach (component-level + composed-path defense-in-depth) is correct and comprehensive. All four `getGroupId().replace('.', '/')` call sites are now validated.

One minor fix needed: the `@since` tag should be updated.

### 📋 PR Metadata

| Labels | _(none)_ | `fix` |

### 🔀 Backport Status

✅ Backport to 1.9.x covered by PR #2092.

cstamas (approved):

PR #2088, opened 2026-08-30
Resolve file transport locations with path operations instead of substring matching

slachiewicz:
`FileTransporter.getPath` tested the raw location string with `path.contains("../")` before resolving
it against the repository base directory. A substring test on an unparsed string is a fragile way to
decide whether a resolved path stays where it should, and it does not consider an absolute location
at all.

This replaces it with file system path operations:

- reject a null location,
- reject a location that parses as an absolute path,
- reject any location with a `..` segment,
- resolve against the base directory, normalize, and require the result to remain under it.

The tests run against both a real file system and an in-memory one (Jimfs), including on Windows,
where an absolute location has a different shape. `absoluteLocation` in the test builds that URI in a
platform-neutral way rather than hard-coding a separator.

### DIFF
--- a/maven-resolver-transport-file/src/main/java/org/eclipse/aether/transport/file/FileTransporter.java
+++ b/maven-resolver-transport-file/src/main/java/org/eclipse/aether/transport/file/FileTransporter.java
@@ -163,10 +163,22 @@ final class FileTransporter extends AbstractTransporter {
 
     private Path getPath(TransportTask task, boolean required) throws Exception {
         String path = task.getLocation().getPath();
-        if (path.contains("../")) {
+        if (path == null) {
+            throw new IllegalArgumentException("illegal resource path: null");
+        }
+        Path relative = fileSystem.getPath(path);
+        if (relative.isAbsolute()) {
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Clean, well-tested security hardening that fixes two confirmed bypass vectors in the old `path.contains("../")` check:

1. **Bare `..` segment** — `".."` does not contain `"../"`, so the old check passed it through.
2. **Absolute paths** — `basePath.resolve("/etc/passwd")` ignores the base directory entirely.

The new implementation uses proper path operations: reject null, reject absolute paths, reject `..` segments, resolve+normalize against base dir, and verify containment. Defense-in-depth approach (explicit segment rejection AND `startsWith` containment check) is good security practice.

- Using `fileSystem.getPath(path)` rather than `Paths.get(path)` is correct for non-default filesystems.
- `normalize()` does not resolve symlinks (same limitation as old code, not a regression).
- Test coverage is thorough with four GET rejection tests, though PUT/PEEK test counts differ slightly.

cstamas (approved):

PR #2089, opened 2026-08-30
Validate the repository key used as a local repository path segment

slachiewicz:
When the local repository is split, the repository key is spliced directly into the local repository
path prefix by `LocalPathPrefixComposerFactorySupport`, and into the file name used by
`SparseDirectoryTrustedChecksumsSource`. The key comes from a configurable
`RepositoryKeyFunction`, so its shape is not guaranteed to be a usable single path segment.

This validates the key with the existing `PathUtils.validatePathComponent` before using it, so the
prefix composes a valid relative path.

No public API changes.


gnodet (approved):
Well-targeted security hardening that closes a real gap in `stringToPathSegment` (the `.` and `..` cases pass through unchanged since they contain none of the characters in the replacement map) and adds defense-in-depth validation at the two critical call sites where repository keys are used as standalone directory segments.

The layered approach is sound: `stringToPathSegment` sanitizes at the source (inside `RepositoryIdHelper.idToPathSegment`), and `validatePathComponent` acts as a safety net at the point of use.

- `SummaryFileTrustedChecksumsSource.summaryFile()` also uses the repository key in filename construction — could benefit from the same validation in a follow-up (pre-existing, not introduced by this PR).
- Backward compatibility risk is negligible: only affects repository IDs that are exactly `.` or `..`.

cstamas (approved):

PR #2090, opened 2026-08-30
Carry repository provenance into aggregateRepositories

slachiewicz:
`RemoteRepositoryManager.aggregateRepositories` cannot currently tell where the recessive repository
definitions came from. Repositories declared by an artifact descriptor downloaded during dependency
collection arrive by the same path as repositories the build itself supplied, so session mirror,
proxy and authentication settings are applied identically to both.

This adds an overload carrying that provenance:

```java
aggregateRepositories(session, dominant, recessive, recessiveIsRaw, recessiveIsFromDescriptor)
```

An implementation can then apply session authentication only to repositories the operator
configured, and withhold it from descriptor-declared ones unless an operator-defined mirror has been
selected for them. Repositories supplied by the build keep receiving mirror, proxy and authentication
settings exactly as documented today.

The default implementation ignores the new argument and delegates to the existing four-argument
method, so behaviour is unchanged for any implementation that does not override it. Marked
`@since 2.0.23`.

### DIFF
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/impl/RemoteRepositoryManager.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/impl/RemoteRepositoryManager.java
@@ -78,7 +78,7 @@ public interface RemoteRepositoryManager {
      * @param recessiveIsFromDescriptor {@code true} if the recessive repository definitions were declared by a remote
      *            artifact descriptor (POM) rather than by the build itself, {@code false} otherwise.
      * @return The aggregated list of remote repositories, never {@code null}.
-     * @since 2.0.22
+     * @since 2.0.23
      * @see RepositorySystemSession#getMirrorSelector()
      * @see RepositorySystemSession#getProxySelector()
      * @see RepositorySystemSession#getAuthenticationSelector()
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/impl/RemoteRepositoryManager.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/impl/RemoteRepositoryManager.java
@@ -55,6 +55,44 @@ public interface RemoteRepositoryManager {
             boolean recessiveIsRaw);
 
     /**
+     * Aggregates repository definitions by merging duplicate repositories and optionally applies mirror, proxy and
+     * authentication settings from the supplied session, additionally distinguishing the provenance of the recessive
+     * repository definitions. Repository definitions that originate from a remote artifact descriptor (i.e. a POM
+     * downloaded during dependency collection) are remotely supplied input: implementations may withhold session
+     * authentication from them unless an operator-defined mirror has been selected for them, so that session
+     * authentication is applied only to repositories the operator configured. Repository definitions supplied
+     * by the build itself (e.g. via
+     * {@code RepositorySystem#newResolutionRepositories}) must keep receiving mirror, proxy and authentication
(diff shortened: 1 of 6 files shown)


maven-resolver-impl/src/main/java/org/eclipse/aether/impl/RemoteRepositoryManager.java
```diff
@@ -54,6 +54,44 @@ List<RemoteRepository> aggregateRepositories(
...
+     * @param session The repository session during which the repositories will be accessed, must not be {@code null}.
+     * @param dominantRepositories The current list of remote repositories to merge the new definitions into, must not
+     *            be {@code null}.
+     * @param recessiveRepositories The remote repositories to merge into the existing list, must not be {@code null}.
+     * @param recessiveIsRaw {@code true} if the recessive repository definitions have not yet been subjected to mirror,
+     *            proxy and authentication settings, {@code false} otherwise.
+     * @param recessiveIsFromDescriptor {@code true} if the recessive repository definitions were declared by a remote
+     *            artifact descriptor (POM) rather than by the build itself, {@code false} otherwise.
+     * @return The aggregated list of remote repositories, never {@code null}.
+     * @since 2.0.22
```
gnodet: `@since 2.0.22` is incorrect — version 2.0.22 was released on 2026-08-20 without this method. Should be `@since 2.0.23`.

```suggestion
     * @since 2.0.23
```

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultRemoteRepositoryManager.java
```diff
@@ -53,6 +55,30 @@
...
+     * are merged into the effective repository list during dependency collection. When disabled (the default),
+     * session authentication is only applied to such a repository when an operator-defined mirror has been selected
+     * for it; if credentials would have matched a descriptor-declared repository, a warning naming the repository ID
+     * and URL is logged instead. Repositories supplied by the build itself (e.g. aggregated via
+     * {@code RepositorySystem#newResolutionRepositories}) are unaffected and keep receiving matching credentials.
+     * Enabling this restores the legacy behavior of applying matching session authentication to descriptor
+     * declared repositories regardless of their provenance.
+     *
+     * @since 2.0.22
+     * @configurationSource {@link RepositorySystemSession#getConfigProperties()}
```
gnodet, line 73: Same `@since` issue — should be `2.0.23`.

```suggestion
     * @since 2.0.23
```

gnodet (commented):
Sound security hardening that prevents credential leakage from `settings.xml` servers to repositories declared by transitive dependency POMs. The design is correct — all three dependency collector implementations properly pass `recessiveIsFromDescriptor=true`, and the backward-compatible default interface method is the right API evolution strategy.

Two `@since` version tags need updating (2.0.22 has already been released without this feature).

### 📋 PR Metadata

| Labels | _(none)_ | `enhancement`, `fix` |

### 🔀 Backport Status

✅ Backport to 1.9.x covered by PR #2092.

gnodet (approved):
Re-review after new commits: both `@since` tag issues from the first review have been corrected to `2.0.23` (commit 6114824). Auth-scoping logic, API design, and test coverage all look sound.

Minor nit: the PR description body still references `@since 2.0.22` — consider updating for consistency, though this doesn't affect the code.

cstamas (approved):

PR #2091, opened 2026-08-30
Compose named lock keys one segment per coordinate field

slachiewicz:
The named lock mappers build a lock name by concatenating artifact coordinate fields. The fields are
inserted as-is, so a field containing a separator produces a name with more segments than intended.
`BasedirNameMapper` then resolves that name against the locks base directory, and the file lock
factory creates and deletes a lock file at the resolved path.

Two changes:

- `GAVNameMapper` and `GAECVNameMapper` compose each coordinate field through `fieldToSegment`, so a
  field always contributes exactly one segment.
- `BasedirNameMapper` resolves the delegate-provided name against the base directory and requires the
  result to stay under it, since that is where lock files are created and removed.

Marked `@since 2.0.23`. No public API changes.

### DIFF
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/synccontext/named/BasedirNameMapper.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/synccontext/named/BasedirNameMapper.java
@@ -105,7 +105,7 @@ public class BasedirNameMapper implements NameMapper {
      * not contained under it, as the file lock factory creates (and deletes on close) lock files at the
      * resolved path.
      *
-     * @since 2.0.22
+     * @since 2.0.23
      */
     private static String resolveContained(Path basedir, String name) {
         Path resolved = basedir.resolve(name).toAbsolutePath();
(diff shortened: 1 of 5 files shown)


gnodet (approved):
Well-crafted security hardening with a sound two-layer defense:

1. **`fieldToSegment` sanitization** — `PathUtils.stringToPathSegment` replaces all illegal path characters, preventing coordinate fields from introducing path separators that could escape the locks directory.
2. **`BasedirNameMapper.resolveContained` containment check** — standard `normalize()+startsWith()` pattern, correctly marked `private static` to prevent subclass override. Provides defense-in-depth against any delegate that emits traversal sequences.

The `HashingNameMapper` path is inherently safe since it hashes the entire name into a fixed hex string before resolution, eliminating attacker-controlled content.

- Lock names will change for coordinates that previously contained path separator characters (rare/malicious) — worth a release notes mention.
- No `GAECVNameMapperTest` exists, though the mechanism is inherited from `GAVNameMapper` and tested there.

cstamas (approved):

PR #2100, opened 2026-08-31
Fix: log only once the warning in DefRemoteRepoMan

cstamas:
As current code on master logs overwhelming amount of WARNs onto console


gnodet (approved):
Clean, well-scoped fix that deduplicates two WARN-level log messages using session-scoped tracking via `SessionData` and `ConcurrentHashMap`. The approach is consistent with the existing `logMirror` pattern and correctly uses `SessionData` (always available) rather than `RepositoryCache` (can be null).

Thread safety is ensured through `SessionData.computeIfAbsent` returning a shared `ConcurrentHashMap`, and `ConcurrentHashMap.computeIfAbsent` guaranteeing at-most-once execution per key. Both WARN call sites on master are converted — no sites were missed.

Minor observations (non-blocking):
- `logWarnOnce` does not guard with `isWarnEnabled()` before computing the key (the existing `logMirror` guards with `isDebugEnabled()`). In practice WARN is almost always enabled, so the overhead is negligible.

PR #2101, opened 2026-08-31
Compose path segments from repository ids on the 1.9.x line

slachiewicz:
The 1.9.x line predates the repository key function layer, so the split local repository prefix (`LocalPathPrefixComposerFactorySupport`, four sites) and the trusted checksums sources (`SummaryFileTrustedChecksumsSource`, `SparseDirectoryTrustedChecksumsSource`) splice the raw repository id into paths. This routes those ids through `PathUtils.stringToPathSegment` and validates the result, matching the 2.x behaviour after #2089 and #2099. Also ports `GAECVNameMapperTest` from master; `GAECVNameMapper` itself needed no change since it inherits the earlier `fieldToSegment` handling from `GAVNameMapper`.

Repository ids that are exactly `.`/`..` or contain path separators now compose to a neutralized segment (e.g. `-DOTDOT-`) instead of being spliced verbatim; well-formed ids are unaffected.

Verified: `mvn -pl maven-resolver-impl -am test` on JDK 17 → 379 tests, 0 failures.

### DIFF
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/LocalPathPrefixComposerFactorySupport.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/LocalPathPrefixComposerFactorySupport.java
@@ -23,6 +23,7 @@ import org.eclipse.aether.artifact.Artifact;
 import org.eclipse.aether.metadata.Metadata;
 import org.eclipse.aether.repository.RemoteRepository;
 import org.eclipse.aether.util.ConfigUtils;
+import org.eclipse.aether.util.PathUtils;
 
 /**
  * Support class for {@link LocalPathPrefixComposerFactory} implementations: it predefines and makes re-usable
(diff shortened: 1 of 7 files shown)


gnodet (approved):
Clean, well-adapted backport of repository-id path-segment sanitization from master (#2089, #2099) to the 1.9.x line. All six injection points where raw repository ids are spliced into filesystem paths are correctly guarded, tests are comprehensive, and the code follows existing conventions.

Key observations:
- Defense-in-depth pattern (`stringToPathSegment` then `validatePathComponent`) is consistent with the 2.x approach.
- The new `GAECVNameMapperTest` is a clean port from master following 1.9.x test conventions (JUnit 4, Hamcrest).
- The three `dotDotRepositoryIdIsNeutralized` tests each verify that `..` is neutralized to `-DOTDOT-`, including a good negative assertion in the SparseDirectory test.
- The 1.6.x branch does not require further backport — the affected code paths don't exist there.

### 📋 PR Metadata

| Milestone | `2.0.23` | `1.9.28` |

This PR targets `maven-resolver-1.9.x`, so the milestone should be `1.9.28` rather than `2.0.23`.

Issue #2096, opened 2026-08-31
Flaky test or an issue lurking around


cstamas:
### Affected version

master

### Bug description

```
[ERROR] Tests run: 18, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 33.43 s <<< FAILURE! -- in org.eclipse.aether.internal.impl.NamedLocksTrackingFileManagerTest
[ERROR] org.eclipse.aether.internal.impl.NamedLocksTrackingFileManagerTest.testLockingOnCanonicalPath(FS)[1] -- Time elapsed: 7.951 s <<< FAILURE!
org.opentest4j.AssertionFailedError: expected: <[]> but was: <[java.lang.IllegalStateException: Failed to lock for update the tracking file /tmp/NamedLocksTrackingFileManagerTest17636213089660401633.tmp]>
	at org.eclipse.aether.internal.impl.TrackingFileManagerTestSupport.testLockingOnCanonicalPath(TrackingFileManagerTestSupport.java:219)
	at java.base/java.lang.reflect.Method.invoke(Method.java:565)
	at java.base/java.util.Optional.ifPresent(Optional.java:182)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:186)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:214)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:186)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:186)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:214)
	at java.base/java.util.Iterator.forEachRemaining(Iterator.java:133)
	at java.base/java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1939)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:570)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:560)
	at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:153)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:176)
[...]

PR #2102, opened 2026-08-31
Optimize TrackingFileManager unit tests and fix potential lock leaks (#2094)

### LINKED ISSUE #2094: NamedLocksTrackingFileManagerTest is flaky on ubuntu-latest
### Affected version

HEAD (observed on PR builds, 2026-08-30)

### Bug description

`NamedLocksTrackingFileManagerTest`, via `TrackingFileManagerTestSupport`, fails intermittently on Linux and passes on rerun with no code change.

```
[ERROR] Tests run: 18, Failures: 1, Errors: 0, Skipped: 0
[ERROR] NamedLocksTrackingFileManagerTest>TrackingFileManagerTestSupport.testLock...
```

Observed on `Verify / ubuntu-latest jdk-21-zulu 4.0.0-rc-6` and, in a separate run, on `Verify / ubuntu-latest jdk-25-zulu 4.0.0-rc-6`. The excerpt is truncated because the runs were re-run afterwards, which replaced the logs.

### Why this looks like flakiness rather than a real failure

The clearest evidence: one of the two branches failed this test, then passed it after a push whose **only** change was a Javadoc `@since` tag. A comment-only edit cannot affect file locking, so the same code produced both outcomes.

Beyond that:

- Neither pull request touches the named-locks or tracking-file-manager code.
- Two different JDKs on the same OS, and green on every other matrix cell.
- Both went green on rerun.

### Possibly related

#2030 (closed) reported `IpcNamedLockFactoryIT` as flaky with a timeout, and wondered about thread safety. Different class, but the same area, so it may be worth checking whether a shared cause remains.

I have not reproduced this locally; it did not occur in any local run I made.

*This issue was created with AI assistance.*

slachiewicz:
### Summary
1. **Prevent File Descriptor Exhaustion in Tests**: Reduces iteration count in `TrackingFileManagerTestSupport` from 1,000 to 100 in leak checks and canonical path locking tests.
2. **Lock Leak Hardening**:
   - `NamedLockFactoryAdapter`: Ensures `namedLock.close()` is invoked if lock acquisition fails/times out or is interrupted, preventing a permanent +1 reference count leak.
   - `CompositeNamedLock`: Automatically unlocks previously acquired sub-locks if a thread is interrupted or encounters an exception mid-way during multi-lock acquisition.

### Root Cause Analysis (macOS / CI FD Exhaustion)
* In `TrackingFileManagerTestSupport`, each loop iteration creates a new temporary file with a distinct path, creating a distinct `NamedLockKey` and lock file.
* `FileLockNamedLockFactory` maintains an LRU pool of up to `MAX_CACHED_CHANNELS = 200` open file channels for closed locks.
* Running 1,000 iterations per test method saturates the 200-channel cache. On macOS (where default `ulimit -n` is 256), combining 200 cached channels with JVM/Surefire baseline descriptors (~50 FDs) directly exhausts the 256 FD limit and throws `Too many open files`.
* Lowering iterations to 100 ensures channel cache growth peaks at ~100 FDs per test method (staying safely below 256), while still thoroughly verifying leak-free behavior and canonical path deduplication.
* Reduces test runtime from 18.8s to 5.0s (-73% speedup).

Resolves #2094; relates to #2096.

### DIFF
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/synccontext/named/NamedLockFactoryAdapter.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/synccontext/named/NamedLockFactoryAdapter.java
@@ -258,15 +258,21 @@ public final class NamedLockFactoryAdapter {
                     }
                 } catch (InterruptedException e) {
                     // if we are here, means we were interrupted: fail
-                    namedLock.close();
-                    close();
+                    try {
+                        namedLock.close();
+                    } finally {
+                        close();
+                    }
                     Thread.currentThread().interrupt();
                     throw new RuntimeException(e);
                 }
             }
             // if we are here, means all attempts were unsuccessful: fail
-            namedLock.close();
-            close();
+            try {
+                namedLock.close();
+            } finally {
+                close();
+            }
             String message = "Could not acquire " + lockKind + " lock for "
                     + lockSubjects(artifacts, metadatas) + " in " + timeStr
                     + "; consider using '" + CONFIG_PROP_TIME
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/synccontext/named/NamedLockFactoryAdapter.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/synccontext/named/NamedLockFactoryAdapter.java
@@ -258,12 +258,14 @@ public final class NamedLockFactoryAdapter {
                     }
                 } catch (InterruptedException e) {
                     // if we are here, means we were interrupted: fail
+                    namedLock.close();
                     close();
                     Thread.currentThread().interrupt();
                     throw new RuntimeException(e);
                 }
             }
             // if we are here, means all attempts were unsuccessful: fail
+            namedLock.close();
             close();
             String message = "Could not acquire " + lockKind + " lock for "
                     + lockSubjects(artifacts, metadatas) + " in " + timeStr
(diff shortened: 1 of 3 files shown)


gnodet (approved):
Correct and well-motivated fix for a lock reference-count leak introduced during the 2.0.0 composite-lock refactoring, plus sound defensive handling for `CompositeNamedLock` interruption, and a justified test iteration reduction.

Key observations:
- The `NamedLockFactoryAdapter` leak is a real regression from the 2.0.0 refactoring — the 1.9.x branch already called `namedLock.close()` on acquisition failure. This PR correctly restores that behavior.
- The `CompositeNamedLock` fix properly handles the case where a thread is interrupted while acquiring the Nth sub-lock: the already-acquired N-1 sub-locks in the `step` deque are now unlocked. The try-catch-Throwable + rethrow pattern handles both `InterruptedException` and unexpected unchecked exceptions cleanly.
- Reducing iterations from 1000 to 100 is well-justified given the `FileLockNamedLockFactory` LRU pool of 200 channels vs macOS default 256 FD limit. 100 iterations still provides meaningful leak detection coverage.
- No backport needed — `CompositeNamedLock` is `@since 2.0.0` (doesn't exist in 1.9.x/1.6.x), and the 1.9.x branch already has the correct close behavior in `NamedLockFactoryAdapter`.

Minor observation (non-blocking): the `namedLock.close()` + `close()` sequence on acquisition failure is not wrapped in try-finally, so if `namedLock.close()` threw, the subsequent `close()` would be skipped. This is extremely unlikely (it only decrements a ref count) and the fix is already a strict improvement over the status quo.

slachiewicz:
Addressed in commit 8c73e073: wrapped `namedLock.close()` and `close()` in `try-finally` in both the interrupted and timeout catch paths in `NamedLockFactoryAdapter` to guarantee both cleanup steps execute even if one throws.

gnodet (approved):
Re-review after new commit `8c73e073`. The single post-approval change directly addresses the previous review's observation by wrapping `namedLock.close()` / `close()` in try-finally in both error paths of `NamedLockFactoryAdapter`. This correctly ensures the reference-count decrement always executes even if `namedLock.close()` throws.

No other changes were introduced — all pre-approval fixes (lock leak, `CompositeNamedLock` defensive interruption handling, test iteration reduction) remain unchanged. LGTM.

cstamas (approved):

PR #2085, opened 2026-08-30
fix: bind credentials to origin and reject TLS downgrade redirects

gnodet:
## Summary

Fixes 5 findings from the maven-resolver security audit (scan-maven-resolver-20260811):

| Finding | Severity | Description |
|---------|----------|-------------|
| f002 | MEDIUM | Settings credentials attached by bare ID to POM-declared repositories |
| f008 | MEDIUM | Apache transport silently follows https-to-http downgrade redirects |
| f009 | MEDIUM | JDK transport Authenticator hands credentials to any challenging host |
| f016 | MEDIUM | Jetty transport follows https-to-http redirects without a downgrade guard |
| f027 | LOW | Operator-configured static auth headers follow redirects cross-origin |

**Root cause:** Credential release decisions ignore where the request is actually going: selectors match on bare repository id with no host binding, authenticators answer any challenger, redirect handling follows cross-origin and https-to-http hops.

**Fix:** Enforce one rule across all transports: credentials go only to the configured origin, and https never silently becomes http.

## Test plan
- [ ] Existing tests pass
- [ ] Credential scoping validated per transport
- [ ] TLS downgrade rejection tested

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2081, opened 2026-08-30
fix: bound response reads and sanitize remote-derived strings

gnodet:
## Summary

Fixes 4 findings from the maven-resolver security audit (scan-maven-resolver-20260811):

| Finding | Severity | Description |
|---------|----------|-------------|
| f004 | MEDIUM | HTTP error-response bodies read unbounded into memory (decompression-amplified OOM) |
| f024 | LOW | Hostile remote checksum file causes unbounded allocation |
| f026 | LOW | Server Last-Modified future date permanently pins artifact against refresh |
| f028 | LOW | Remote-derived strings reach console logs without control-character sanitization |

**Root cause:** Response bodies, checksum files, timestamps, and strings from the remote side are consumed with no size cap, no plausibility clamp, and no control-character sanitization.

**Fix:** Bound before allocating: cap error-body and checksum reads, clamp timestamps, sanitize control characters in log messages.

## Test plan
- [ ] Existing tests pass
- [ ] Error body read capped at 64KB
- [ ] Checksum file read capped
- [ ] Last-Modified clamping tested
- [ ] Control character sanitization tested

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1954, opened 2026-07-11
[MRESOLVER-1954] Provide Apache HttpClient 5.x transporter module (maven-resolver-transport-apache5)


neilscallywag:
### New feature, improvement proposal

Hi maintainers,

Could the project please consider migrating maven-resolver-transport-apache away from the older Apache HttpComponents coordinates:

HttpComponents coordinates: - [org.apache.httpcomponents:httpcore](https://mvnrepository.com/artifact/org.apache.httpcomponents/httpcore) and [org.apache.httpcomponents:httpclient](https://mvnrepository.com/artifact/org.apache.httpcomponents/httpclient)

The newer coordinates are: - [org.apache.httpcomponents.core5:httpcore5](https://mvnrepository.com/artifact/org.apache.httpcomponents.core5/httpcore5) and  [org.apache.httpcomponents.client5:httpclient5](https://mvnrepository.com/artifact/org.apache.httpcomponents.client5/httpclient5)

[The developers also recommend moving to the newer versions](https://lists.apache.org/thread/qpnhw3ybsx2jq2oknzroq90jh5xts8lo) and the older coordinates appear to no longer be actively maintained. Our build scanners have also highlighted a CVE (similar to the OP of the thread) affecting HttpCore 5 which, according to the attached Apache discussion, may also affect version 4. Apache maintainers indicated that a public patch for version 4 may not be released. For info, we are not affiliated with the poster of the attached apache thread.

I understand that HttpComponents 5 may not be a drop-in compatible with 4.x and that this would require implementation and testing changes rather than only updating the coordinates. 

Some of our services depend transitively on httpcomponents library, so we are approaching the developers down the chain to ask them to upgrade, [example](https://github.com/smallrye/smallrye-beanbag/issues/280). We have a workaround specific for our use case in the time being so it is not an issue that requires urgency.

Thanks :) !!

kwin:
There is some draft in #742 for Apache HTTP Client 5.x.

neilscallywag:
Thanks! Will KIV.

slachiewicz:
### Summary & Migration Plan

As noted, upgrading from Apache HttpComponents 4.x (`httpclient` / `httpcore`) to HttpComponents 5.x (`org.apache.httpcomponents.client5:httpclient5` / `org.apache.httpcomponents.core5:httpcore5`) is not a drop-in dependency bump due to breaking API changes across connection management, SSL/TLS configuration, authentication, and execution models.

[...]

PR #2103, opened 2026-08-31
[MRESOLVER-1954] Add maven-resolver-transport-apache5 module backed by Apache HttpComponents 5.x

slachiewicz:
Draft PR for https://github.com/apache/maven-resolver/issues/1954

### Overview

Following maintainer feedback (coexistence with Wagon HTTP which shares HttpClient 4.x), this PR introduces a new dedicated transporter module: **`maven-resolver-transport-apache5`**, leaving the existing `maven-resolver-transport-apache` (HttpClient 4.5.x) untouched.

This allows Maven 3.x / 3.10.x to continue using `apache` (HC 4.5.x) while modern integrations (e.g. Maven 4.1+) can adopt Apache HttpClient 5.x. Selection follows the usual transporter factory rules: the module's factory is `Apache5TransporterFactory` (component name `apache5`, default priority 5.0, the same as the HttpClient 4 factory), so it is selected when it is the only Apache transport on the class path, or by raising its priority with `-Daether.priority.Apache5TransporterFactory=<value>` when both are present. The distinct class name keeps that priority key from also matching the HttpClient 4 factory.

### Summary of Changes

1. **New Module:** `maven-resolver-transport-apache5`
   - Artifact: `org.apache.maven.resolver:maven-resolver-transport-apache5`
   - Component: `@Named("apache5")`, class `Apache5TransporterFactory`
   - Config prefix: `aether.transport.apache5.*`
   - Dependencies: Apache HttpClient `5.4.2` / HttpCore `5.3.3`
2. **Behavior Parity & Hardening** (ported from the HttpClient 4 transport as of #2085 and #2081):
   - `ResolverRedirectStrategy` (no https→http redirects unless `aether.transport.apache5.followInsecureRedirects=true`), `OriginScopedHeadersInterceptor` (configured headers and `Authorization` not re-sent cross-origin), credentials scoped to protocol/host/port via HC5 `AuthScope`, bounded RFC 9457 error-body read and `Last-Modified` clamp through the shared helpers.
   - Adapted `SharingHttpContext`, session pooling, TLS `SSLSession` context adaptation, and proxy routing to HC5 APIs. When a local bind address is configured the route planner is rebuilt by hand with the same precedence as `HttpClientBuilder`: an explicit repository proxy wins over system properties.
   - Thread-safe auth scheme caching using HC5 `StateHolder` and `BasicAuthCache`.
   - Subclassed `DefaultHttpRequestRetryStrategy` to maintain backoff & retry parity.
   - Public suffix hostname verification via `HttpsSupport.getDefaultHostnameVerifier()`.
   - Guarded `PutTaskEntity` against premature transmission on expect-continue rejection.
3. **Shared utility change (affects all HTTP transports):** `HttpTransporterUtils.getHttpExpectContinue` now accepts a `Boolean` configuration value as well as a `String`. Previously a `Boolean` set programmatically on the session was ignored and the transport fell back to its default. This also changes the HttpClient 4 and JDK transports, which use the same helper; a unit test covers both value types.

### Verification
- Full reactor build (32 modules) passing cleanly with JDK 21.
- All unit & integration tests passing in `maven-resolver-transport-apache5` (shared `HttpTransporterTest` suite included).
- `maven-resolver-transport-apache` (v4) passes all tests untouched.
- Spotless and Apache RAT checks passing.

### DIFF
--- a/maven-resolver-transport-apache5/pom.xml
+++ b/maven-resolver-transport-apache5/pom.xml
@@ -1,0 +1,122 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
+  https://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing,
+  software distributed under the License is distributed on an
+  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  KIND, either express or implied.  See the License for the
+  specific language governing permissions and limitations
+  under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.apache.maven.resolver</groupId>
+    <artifactId>maven-resolver</artifactId>
+    <version>2.0.23-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>maven-resolver-transport-apache5</artifactId>
+
+  <name>Maven Artifact Resolver Transport Apache 5</name>
+  <description>A transport implementation for repositories using http:// and https:// URLs based on Apache HTTP Client 4.5.</description>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.maven.resolver</groupId>
+      <artifactId>maven-resolver-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.maven.resolver</groupId>
+      <artifactId>maven-resolver-spi</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.maven.resolver</groupId>
+      <artifactId>maven-resolver-util</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpclient</artifactId>
+      <version>4.5.14</version>
+      <exclusions>
+        <exclusion>
+          <!-- using jcl-over-slf4j instead -->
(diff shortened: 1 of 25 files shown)


cstamas:
Supercool, but few remarks:
* we need apache-httpclient 4.x still (ie. it shares same client with Wagon HTTP)
* could you just "move" this into _new_ module, named like `maven-resolver-transport-apache5`? That would allow them to coexist
* so integrating like (ie Maven) have a free choice which one to use?

slachiewicz:
I'll check how wagon beehive and I this it should be possible to move it also quickly.
So far this change is more for 2.1 than 2.0 but if it works I'm open to integrate it faster

cstamas:
I'd not touch Wagon, is legacy, everything is moving away from it (maybe site only is not?).
My point is that it would be better to have the two apache transports (4 and 5) coexist, as it would allow for example to keep 3.10 with apache4 but in Maven 4 (ie 4.1) move to apache5?

slachiewicz:
@cstamas Updated as suggested! 

### Summary of updates:
- **Dedicated Module:** Created `maven-resolver-transport-apache5` (backed by Apache HttpClient `5.4.2` / HttpCore `5.3.3`) with `@Named("apache5")` and package `org.eclipse.aether.transport.apache5`.
- **Preserved Existing Transporter:** `maven-resolver-transport-apache` (HttpClient 4.5.x) is left completely untouched for full compatibility with Wagon HTTP and existing Maven 3.x setups.
- **Reviewable Git History:** Structured the branch with a 2-commit history:
  1. Module creation derived from `maven-resolver-transport-apache`.
  2. The exact HC5 migration + behavior parity hardening.
- **Verification:** Full reactor build (32 modules) and all 118 unit/integration tests in `apache5` (as well as `apache`) pass cleanly.

cstamas (approved):

kwin:
I would rather opt for reducing the number of HTTP transporters in resolver. This module is the fifth one:
1. Apache HTTP Client 4.5
2. JRE HTTP Client (Java 11+)
3. Jetty HTTP Client (Java 17+)
4. UrlConnection
5. Apache HTTP Client 5.x

Why is 2. not sufficient with 1. and 4. as fallbacks for Java 8?

slachiewicz:
> I would rather opt for reducing the number of HTTP transporters in resolver. This module is the fifth one:
> 
> 1. Apache HTTP Client 4.5
> Why is 2. not sufficient with 1. and 4. as fallbacks for Java 8?

goal is to move away from  not supported http 4.5

kwin:
For me option 1 (based on legacy Apache HTTP Client 4.5) is good enough as fallback for Java 8. Maven 4 will use JRE HTTP Client. I don't yet see a need to ship another HTTP impl.

cstamas:
Why was this closed?

cstamas:
So, why was this closed?

PR #2104, opened 2026-09-01
Fix: tidy up f013

cstamas:
Flip system key instead.


gnodet (approved):
Looks good — clean improvement over the lazy `volatile` init. Two minor observations, neither blocking:

1. **Stale Javadoc on `realBasePath`:** the comment still says "Lazily computed" but the field is now `final` and eagerly initialized. Worth a quick update (e.g. "Resolved at construction time").

2. **Unrelated `RemoteRepository.java` change:** the `@see` fully-qualified fix is harmless but unrelated to the `realBasePath` improvement — could be a separate commit or dropped to keep the diff focused.

Otherwise: `final` field, fail-fast at construction, `IOException` surfacing cleanly through the package-private constructor to the factory — strictly better than the previous pattern. 👍

PR #2105, opened 2026-09-01
docs: remove outdated assembly plugin consequences

jiteshkhatri11:
## Description

Remove the outdated **Important Consequences** section from `common-misconceptions.md`.

The section references `MASSEMBLY-1008` and `MRESOLVER-391`, which are no longer appropriate for the current documentation.

## Changes

* Remove the outdated Assembly plugin discussion.
* Keep the current explanation of runtime and test dependency graphs unchanged.
* Remove the extra blank line before the code block.

## Verification

* `git diff --check` passes.
* Maven build passes successfully.

### DIFF
--- a/src/site/markdown/common-misconceptions.md
+++ b/src/site/markdown/common-misconceptions.md
@@ -1,4 +1,5 @@
-# Common Misconceptions`r`n<!--
+# Common Misconceptions
+<!--
 Licensed to the Apache Software Foundation (ASF) under one
 or more contributor license agreements.  See the NOTICE file
 distributed with this work for additional information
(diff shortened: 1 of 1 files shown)


src/site/markdown/common-misconceptions.md
```diff
@@ -1,5 +1,4 @@
-# Common Misconceptions
-<!--
+# Common Misconceptions`r`n<!--
```
slachiewicz: this is wrong, revert it

cstamas:
The documented behaviour is still true, also depends on what Maven and plugin versions are used.

jiteshkhatri11:
Thanks for the clarification. I misunderstood the applicability of the documented Assembly Plugin behavior and treated the referenced issues as making the section outdated.

I'll restore the section and update the PR accordingly.

PR #2108, opened 2026-09-01
Fix listener error swallowing (#1990), SyncContext double-close (#1992), and test flakiness (#1350)

### LINKED ISSUE #1350: [MRESOLVER-674] Derandomize GenericVersionTest.testCompareUuidRandom
**[Elliotte Rusty Harold](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=elharo)** opened **[MRESOLVER-674](https://issues.apache.org/jira/browse/MRESOLVER-674?redirect=false)** and commented

Using random in tests makes test non-reproducible

---
No further details from [MRESOLVER-674](https://issues.apache.org/jira/browse/MRESOLVER-674?redirect=false)

### LINKED ISSUE #1990: Listener RuntimeException silently swallowed by default in ChainedRepositoryListener/ChainedTransferListener
**Affected version:** HEAD

**Files:**
- `maven-resolver-util/src/main/java/org/eclipse/aether/util/listener/ChainedRepositoryListener.java:117`
- `maven-resolver-util/src/main/java/org/eclipse/aether/util/listener/ChainedTransferListener.java:118`

```java
@SuppressWarnings("EmptyMethod")
protected void handleError(...) {
    // default just swallows errors
}
```

All `RuntimeException`s thrown by any chained listener (NPEs, `IndexOutOfBoundsException`, etc.) are silently swallowed unless the user subclasses and overrides `handleError`. This makes listener bugs invisible.

Originally reported in #1944.

### LINKED ISSUE #1992: SyncContext double-close risk in DefaultArtifactResolver and DefaultMetadataResolver
**Affected version:** HEAD

**Files:**
- `maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultArtifactResolver.java:189-190, 393-397, 430-432`
- `maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultMetadataResolver.java:142-143, 302-373`

Two `SyncContext` instances (`shared` and `exclusive`) are created in try-with-resources. Inside `resolve()`, `current` starts as `shared`, and when switching to `exclusive` (line 393-396), `current.close()` is called, then `current = exclusive`. The outer try-with-resources also closes both on exit (line 202 for try-with-resources, and line 431 for the inner `current.close()`). While `SyncContext.close()` is documented as idempotent, any implementation that violates this contract would cause issues.

Originally reported in #1944.

slachiewicz:
### Summary of Fixes

1. **[#1990](https://github.com/apache/maven-resolver/issues/1990)**: `ChainedRepositoryListener` and `ChainedTransferListener` now log exceptions from listeners at WARNING level in `handleError()` instead of silently swallowing them. Added unit tests verifying that an exception in one listener does not disrupt delivery to subsequent listeners in the multicast chain.
2. **[#1992](https://github.com/apache/maven-resolver/issues/1992)**: Refactored `DefaultArtifactResolver` and `DefaultMetadataResolver` to eliminate redundant internal/finally `SyncContext.close()` calls, ensuring `SyncContext` instances are closed strictly once. Added unit test `testSyncContextIsClosedExactlyOnce` verifying single-close lifecycle.
3. **[#1350](https://github.com/apache/maven-resolver/issues/1350) ([MRESOLVER-674])**: Seeded `Random` in `GenericVersionTest.testCompareUuidRandom` with a fixed seed to ensure deterministic, reproducible test runs across CI environments.

### DIFF
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultArtifactResolver.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultArtifactResolver.java
@@ -203,20 +203,18 @@ public class DefaultArtifactResolver implements ArtifactResolver {
             throws ArtifactResolutionException {
         requireNonNull(session, "session cannot be null");
         requireNonNull(requests, "requests cannot be null");
-        try (SyncContext shared = syncContextFactory.newInstance(session, true);
-                SyncContext exclusive = syncContextFactory.newInstance(session, false)) {
-            Collection<Artifact> artifacts = new ArrayList<>(requests.size());
-            SystemDependencyScope systemDependencyScope = session.getSystemDependencyScope();
-            for (ArtifactRequest request : requests) {
-                if (systemDependencyScope != null
-                        && systemDependencyScope.getSystemPath(request.getArtifact()) != null) {
-                    continue;
-                }
-                artifacts.add(request.getArtifact());
+        SyncContext shared = syncContextFactory.newInstance(session, true);
+        SyncContext exclusive = syncContextFactory.newInstance(session, false);
+        Collection<Artifact> artifacts = new ArrayList<>(requests.size());
+        SystemDependencyScope systemDependencyScope = session.getSystemDependencyScope();
+        for (ArtifactRequest request : requests) {
+            if (systemDependencyScope != null && systemDependencyScope.getSystemPath(request.getArtifact()) != null) {
+                continue;
             }
-
-            return resolve(shared, exclusive, artifacts, session, requests);
+            artifacts.add(request.getArtifact());
         }
+
+        return resolve(shared, exclusive, artifacts, session, requests);
     }
 
     @SuppressWarnings("checkstyle:methodlength")
@@ -470,7 +468,13 @@ public class DefaultArtifactResolver implements ArtifactResolver {
                 return results;
             }
         } finally {
-            current.close();
+            try {
+                current.close();
+            } finally {
+                if (current == shared) {
+                    exclusive.close();
+                }
+            }
         }
     }
 
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultArtifactResolver.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultArtifactResolver.java
@@ -212,15 +212,10 @@ public class DefaultArtifactResolver implements ArtifactResolver {
             artifacts.add(request.getArtifact());
         }
 
-        SyncContext shared = syncContextFactory.newInstance(session, true);
-        SyncContext exclusive;
-        try {
-            exclusive = syncContextFactory.newInstance(session, false);
-        } catch (RuntimeException e) {
-            shared.close();
-            throw e;
+        try (SyncContext shared = new CloseOnceSyncContext(syncContextFactory.newInstance(session, true));
+                SyncContext exclusive = new CloseOnceSyncContext(syncContextFactory.newInstance(session, false))) {
+            return resolve(shared, exclusive, artifacts, session, requests);
         }
-        return resolve(shared, exclusive, artifacts, session, requests);
     }
 
     @SuppressWarnings("checkstyle:methodlength")
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultArtifactResolver.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultArtifactResolver.java
@@ -203,8 +203,6 @@ public class DefaultArtifactResolver implements ArtifactResolver {
             throws ArtifactResolutionException {
         requireNonNull(session, "session cannot be null");
         requireNonNull(requests, "requests cannot be null");
-        SyncContext shared = syncContextFactory.newInstance(session, true);
-        SyncContext exclusive = syncContextFactory.newInstance(session, false);
         Collection<Artifact> artifacts = new ArrayList<>(requests.size());
         SystemDependencyScope systemDependencyScope = session.getSystemDependencyScope();
         for (ArtifactRequest request : requests) {
@@ -214,6 +212,14 @@ public class DefaultArtifactResolver implements ArtifactResolver {
             artifacts.add(request.getArtifact());
         }
 
+        SyncContext shared = syncContextFactory.newInstance(session, true);
+        SyncContext exclusive;
+        try {
+            exclusive = syncContextFactory.newInstance(session, false);
+        } catch (RuntimeException e) {
+            shared.close();
+            throw e;
+        }
         return resolve(shared, exclusive, artifacts, session, requests);
     }
 
@@ -424,8 +430,9 @@ public class DefaultArtifactResolver implements ArtifactResolver {
                 }
 
                 if (!groups.isEmpty() && current == shared) {
-                    current.close();
+                    SyncContext sharedContext = current;
                     current = exclusive;
+                    sharedContext.close();
                     continue;
                 }
 
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultMetadataResolver.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultMetadataResolver.java
@@ -144,15 +144,10 @@ public class DefaultMetadataResolver implements MetadataResolver {
             metadata.add(request.getMetadata());
         }
 
-        SyncContext shared = syncContextFactory.newInstance(session, true);
-        SyncContext exclusive;
-        try {
-            exclusive = syncContextFactory.newInstance(session, false);
-        } catch (RuntimeException e) {
-            shared.close();
-            throw e;
+        try (SyncContext shared = new CloseOnceSyncContext(syncContextFactory.newInstance(session, true));
+                SyncContext exclusive = new CloseOnceSyncContext(syncContextFactory.newInstance(session, false))) {
+            return resolve(shared, exclusive, metadata, session, requests);
         }
-        return resolve(shared, exclusive, metadata, session, requests);
     }
 
     @SuppressWarnings("checkstyle:methodlength")
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultMetadataResolver.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultMetadataResolver.java
@@ -139,15 +139,14 @@ public class DefaultMetadataResolver implements MetadataResolver {
             RepositorySystemSession session, Collection<? extends MetadataRequest> requests) {
         requireNonNull(session, "session cannot be null");
         requireNonNull(requests, "requests cannot be null");
-        try (SyncContext shared = syncContextFactory.newInstance(session, true);
-                SyncContext exclusive = syncContextFactory.newInstance(session, false)) {
-            Collection<Metadata> metadata = new ArrayList<>(requests.size());
-            for (MetadataRequest request : requests) {
-                metadata.add(request.getMetadata());
-            }
-
-            return resolve(shared, exclusive, metadata, session, requests);
+        SyncContext shared = syncContextFactory.newInstance(session, true);
+        SyncContext exclusive = syncContextFactory.newInstance(session, false);
+        Collection<Metadata> metadata = new ArrayList<>(requests.size());
+        for (MetadataRequest request : requests) {
+            metadata.add(request.getMetadata());
         }
+
+        return resolve(shared, exclusive, metadata, session, requests);
     }
 
     @SuppressWarnings("checkstyle:methodlength")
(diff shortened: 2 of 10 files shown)


maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultArtifactResolver.java
```diff
@@ -203,20 +203,18 @@ public List<ArtifactResult> resolveArtifacts(
...
-            Collection<Artifact> artifacts = new ArrayList<>(requests.size());
-            SystemDependencyScope systemDependencyScope = session.getSystemDependencyScope();
-            for (ArtifactRequest request : requests) {
-                if (systemDependencyScope != null
-                        && systemDependencyScope.getSystemPath(request.getArtifact()) != null) {
-                    continue;
-                }
-                artifacts.add(request.getArtifact());
+        SyncContext shared = syncContextFactory.newInstance(session, true);
+        SyncContext exclusive = syncContextFactory.newInstance(session, false);
```
gnodet: **Resource leak (HIGH):** Removing try-with-resources introduces a resource leak. If `syncContextFactory.newInstance(session, false)` on the next line throws, the `shared` SyncContext created here is never closed — the `resolve()` method's finally block only runs if `resolve()` is actually entered.

The original try-with-resources guaranteed both contexts were closed regardless of where an exception occurred. Since `SyncContext` wraps named locks, a leaked context could hold a lock open.

One approach:
```java
SyncContext shared = syncContextFactory.newInstance(session, true);
SyncContext exclusive;
try {
    exclusive = syncContextFactory.newInstance(session, false);
} catch (RuntimeException | Error e) {
    shared.close();
    throw e;
}
```
Or alternatively, keep try-with-resources for the creation but move the close logic into the `resolve()` method's finally block using a pattern that handles both the try-with-resources close and the explicit close.

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultMetadataResolver.java
```diff
@@ -139,15 +139,14 @@ public List<MetadataResult> resolveMetadata(
...
-        try (SyncContext shared = syncContextFactory.newInstance(session, true);
-                SyncContext exclusive = syncContextFactory.newInstance(session, false)) {
-            Collection<Metadata> metadata = new ArrayList<>(requests.size());
-            for (MetadataRequest request : requests) {
-                metadata.add(request.getMetadata());
-            }
-
-            return resolve(shared, exclusive, metadata, session, requests);
+        SyncContext shared = syncContextFactory.newInstance(session, true);
+        SyncContext exclusive = syncContextFactory.newInstance(session, false);
```
gnodet: **Resource leak (HIGH):** Same pattern as `DefaultArtifactResolver` — if `newInstance(session, false)` throws, the `shared` context from the previous line is never closed. The `resolve()` method's finally block only runs if `resolve()` is entered.

maven-resolver-impl/src/test/java/org/eclipse/aether/internal/impl/DefaultArtifactResolverTest.java
```diff
@@ -1083,4 +1083,39 @@ public void add(RepositorySystemSession session, LocalMetadataRegistration reque
...
+
+        org.eclipse.aether.SyncContext countingSyncContext = new org.eclipse.aether.SyncContext() {
+            @Override
+            public void acquire(
+                    Collection<? extends Artifact> artifacts,
+                    Collection<? extends org.eclipse.aether.metadata.Metadata> metadatas) {}
+
+            @Override
+            public void close() {
+                closeCount.incrementAndGet();
```
gnodet, line 1098: **Misleading test (MEDIUM):** The factory `(s, shared) -> countingSyncContext` ignores its parameters and always returns the same instance. So `shared` and `exclusive` inside the resolver both point to the same object.

The comment on line 1119 says "2 instances were created" but only 1 exists — it is closed twice. If the code had a bug where it closed one context twice and the other never, this test would not catch it.

Consider using two separate instances with separate counters:
```java
AtomicInteger sharedCloseCount = new AtomicInteger();
AtomicInteger exclusiveCloseCount = new AtomicInteger();
// ... factory returns distinct instances based on the `shared` parameter
```

maven-resolver-util/src/main/java/org/eclipse/aether/util/listener/ChainedRepositoryListener.java
```diff
@@ -110,12 +110,14 @@ public void remove(RepositoryListener listener) {
         }
     }
```
gnodet, line 114: **JUL vs SLF4J (LOW):** This uses `java.util.logging.Logger` while the project's `maven-resolver-impl` module uses SLF4J exclusively. The choice is understandable since `maven-resolver-util` doesn't depend on SLF4J, but these WARNING messages will not appear in SLF4J-configured logging setups unless a JUL-to-SLF4J bridge is installed.

Also, several fully-qualified class names are used inline (`java.util.logging.Logger`, `java.util.logging.Level.WARNING`) instead of proper import statements — same applies in `ChainedTransferListener.java` and the test files. Using imports would be more consistent with the rest of the codebase.

gnodet (changes requested):
Two confirmed resource leaks (HIGH) in `DefaultArtifactResolver` and `DefaultMetadataResolver` where removing try-with-resources creates a gap: if the second `syncContextFactory.newInstance()` call throws, the first `shared` SyncContext is never closed. The original code handled this correctly via try-with-resources semantics. The test for the SyncContext fix is technically valid but misleading — it uses a single SyncContext instance for both shared and exclusive roles, making the assertion comment inaccurate.

The listener error handling improvement and test derandomization are sound.

---

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultArtifactResolver.java
gnodet: **Resource leak (HIGH):** Removing try-with-resources introduces a resource leak. If `syncContextFactory.newInstance(session, false)` on the next line throws, the `shared` SyncContext created here is never closed — the `resolve()` method's finally block only runs if `resolve()` is actually entered.

The original try-with-resources guaranteed both contexts were closed regardless of where an exception occurred. Since `SyncContext` wraps named locks, a leaked context could hold a lock open.

One approach:
```java
SyncContext shared = syncContextFactory.newInstance(session, true);
SyncContext exclusive;
try {
    exclusive = syncContextFactory.newInstance(session, false);
} catch (RuntimeException | Error e) {
    shared.close();
    throw e;
}
```
Or alternatively, keep try-with-resources for the creation but move the close logic into the `resolve()` method's finally block using a pattern that handles both the try-with-resources close and the explicit close.

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultMetadataResolver.java
gnodet: **Resource leak (HIGH):** Same pattern as `DefaultArtifactResolver` — if `newInstance(session, false)` throws, the `shared` context from the previous line is never closed. The `resolve()` method's finally block only runs if `resolve()` is entered.

maven-resolver-impl/src/test/java/org/eclipse/aether/internal/impl/DefaultArtifactResolverTest.java
gnodet, line 1098: **Misleading test (MEDIUM):** The factory `(s, shared) -> countingSyncContext` ignores its parameters and always returns the same instance. So `shared` and `exclusive` inside the resolver both point to the same object.

The comment on line 1119 says "2 instances were created" but only 1 exists — it is closed twice. If the code had a bug where it closed one context twice and the other never, this test would not catch it.

Consider using two separate instances with separate counters:
```java
AtomicInteger sharedCloseCount = new AtomicInteger();
AtomicInteger exclusiveCloseCount = new AtomicInteger();
// ... factory returns distinct instances based on the `shared` parameter
```

maven-resolver-util/src/main/java/org/eclipse/aether/util/listener/ChainedRepositoryListener.java
gnodet, line 114: **JUL vs SLF4J (LOW):** This uses `java.util.logging.Logger` while the project's `maven-resolver-impl` module uses SLF4J exclusively. The choice is understandable since `maven-resolver-util` doesn't depend on SLF4J, but these WARNING messages will not appear in SLF4J-configured logging setups unless a JUL-to-SLF4J bridge is installed.

Also, several fully-qualified class names are used inline (`java.util.logging.Logger`, `java.util.logging.Level.WARNING`) instead of proper import statements — same applies in `ChainedTransferListener.java` and the test files. Using imports would be more consistent with the rest of the codebase.

gnodet (changes requested):
Two confirmed resource leaks (HIGH) in `DefaultArtifactResolver` and `DefaultMetadataResolver` where removing try-with-resources creates a gap: if the second `syncContextFactory.newInstance()` call throws, the first `shared` SyncContext is never closed. The original code handled this correctly via try-with-resources semantics. The test for the SyncContext fix is technically valid but misleading — it uses a single SyncContext instance for both shared and exclusive roles, making the assertion comment inaccurate.

The listener error handling improvement and test derandomization are sound.

---

elharo (commented):
The claude reviews should be addressed

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultArtifactResolver.java
slachiewicz: Addressed in a9d1d5c7 using your alternative suggestion: restored try-with-resources by wrapping both contexts in a new package-private `CloseOnceSyncContext` that delegates `close()` at most once. This covers the creation window for both `RuntimeException` and `Error`, while `resolve()` can still close the shared context early (before switching to exclusive) without risking a double-close of the underlying context.

maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultMetadataResolver.java
slachiewicz: Same fix applied here in a9d1d5c7 — try-with-resources with the `CloseOnceSyncContext` wrapper, so a failure while creating the exclusive context closes the shared one, and the early close during the shared→exclusive switch stays safe.

maven-resolver-impl/src/test/java/org/eclipse/aether/internal/impl/DefaultArtifactResolverTest.java
slachiewicz: Good catch — fixed in 0d76d597: the test now uses two distinct counting instances, with the factory dispatching on the `shared` flag, and asserts each instance is closed exactly once.

maven-resolver-util/src/main/java/org/eclipse/aether/util/listener/ChainedRepositoryListener.java
slachiewicz: Imports cleaned up in 0d76d597 (also in `ChainedTransferListener` and the tests). JUL is kept deliberately since `maven-resolver-util` has no SLF4J dependency; happy to switch if adding the dependency is preferred.

elharo (approved):

gnodet (approved):
**Re-review after follow-up commits (0d76d597, a9d1d5c7).** All previous findings are addressed:

1. ✅ **Resource leak (HIGH) — DefaultArtifactResolver:** Restored try-with-resources using `CloseOnceSyncContext` wrapper. The creation window is now safe: if `newInstance(session, false)` throws, try-with-resources closes the shared context. The `CloseOnceSyncContext` correctly delegates `close()` at most once, and the `resolve()` finally block properly handles the case where `current == shared` (closes exclusive too via nested try-finally).

2. ✅ **Resource leak (HIGH) — DefaultMetadataResolver:** Same fix applied symmetrically.

3. ✅ **Misleading test (MEDIUM):** Now uses two distinct counting instances dispatched by the `shared` flag, each asserted closed exactly once. New `testSharedSyncContextIsClosedWhenExclusiveCannotBeCreated` test directly covers the resource leak scenario. Test also added for `DefaultMetadataResolver`.

4. ✅ **JUL imports (LOW):** Fully-qualified class names replaced with proper imports. JUL deliberately kept since `maven-resolver-util` has no SLF4J dependency — reasonable tradeoff.

The `CloseOnceSyncContext` wrapper is well designed: package-private, clear Javadoc, `IllegalStateException` on acquire-after-close. The `closed` field correctly omits `volatile` since `SyncContext` is single-threaded by contract.

**Note:** No milestone is set — this should target 2.0.23.

cstamas (approved):

cstamas:
Applied two changes:
* while SyncContext is ST, the common pattern in Resolver is use of AtomicBoolean to implement idempotent close method
* applied same pattern to only one existing SyncContext implementation, thus making close method idempotent

PR #2111, opened 2026-09-01
Create the local repository directory before resolving its real path

slachiewicz:
Since #2104 `EnhancedLocalRepositoryManager` resolves the real path of its base directory in the constructor, and `Path.toRealPath()` requires the path to exist.

A local repository that has not been created yet (first build on a machine, or a new `-Dmaven.repo.local`) therefore made the enhanced factory throw `NoLocalRepositoryManagerException`. `DefaultLocalRepositoryProvider` treats that as "try the next factory" and quietly selected `SimpleLocalRepositoryManager`, so the whole first session ran without `_remote.repositories` tracking. Before #2104 the real path was computed lazily, only for files that already existed, so the directory was always present.

The fix creates the directory before resolving it. Maven creates it on first write anyway and the enhanced manager needs it for its tracking files, so nothing observable changes for a usable local repository; an unwritable location still fails the same way it did before.

Two tests cover the factory directly and the provider's fallback selection with a not-yet-existing base directory; both fail on master with the `NoSuchFileException` cause.

### DIFF
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/EnhancedLocalRepositoryManager.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/EnhancedLocalRepositoryManager.java
@@ -149,6 +149,8 @@ class EnhancedLocalRepositoryManager extends SimpleLocalRepositoryManager {
         this.trackingFilename = requireNonNull(trackingFilename);
         this.trackingFileManager = requireNonNull(trackingFileManager);
         this.localPathPrefixComposer = requireNonNull(localPathPrefixComposer);
+        // a fresh local repository does not exist yet; toRealPath() requires it to
+        Files.createDirectories(getRepository().getBasePath());
         this.realBasePath = getRepository().getBasePath().toRealPath();
     }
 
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Clean, minimal regression fix. PR #2104 moved `toRealPath()` from lazy to eager evaluation in the `EnhancedLocalRepositoryManager` constructor, but `toRealPath()` requires the directory to exist. When the local repository has not been created yet (first build, or a new `-Dmaven.repo.local`), the constructor threw `IOException`, causing silent fallback to `SimpleLocalRepositoryManager`. The fix — a single `Files.createDirectories()` call before `toRealPath()` — is correct, idempotent, and safe for concurrent use.

The two tests are well-designed: `newInstanceForNotYetExistingBasedir` verifies the factory layer, while `providerSelectsEnhancedManagerForNotYetExistingBasedir` tests the complete fallback chain through `DefaultLocalRepositoryProvider`. Both fail on origin/master without the fix.

### 📋 PR Metadata

| Category | _(unlabeled)_ | `bug` |
| Labels | _(none)_ | + `bug` |

### 🔀 Backport Status

✅ Not needed — the regression source (PR #2104) was only merged to master and is not present on either maintenance branch (`maven-resolver-1.9.x`, `maven-resolver-1.6.x`).

---

cstamas (approved):

Issue #1991, opened 2026-07-21
Volatile fields with inconsistent synchronization in IpcClient


elharo:
**Affected version:** HEAD

**File:** `maven-resolver-named-locks-ipc/src/main/java/org/eclipse/aether/named/ipc/IpcClient.java:78-87`

Fields `initialized`, `socket`, `output`, `input`, `receiver` are `volatile` but their compound use (read + method call) lacks synchronization. `send()` reads `output` into a local variable at line 316 without holding the monitor, then synchronizes on the local reference at line 323, but `close()` nulls `output` under `synchronized(this)` at line 361. `receive()` has the same pattern with `input`.

Originally reported in #1944.

Issue #1989, opened 2026-07-21
Production error logging to System.out in IpcServer


elharo:
**Affected version:** HEAD

**File:** `maven-resolver-named-locks-ipc/src/main/java/org/eclipse/aether/named/ipc/IpcServer.java:196`

```java
private static void error(String msg, Throwable t) {
    System.out.println("[ipc] [error] " + msg);
    t.printStackTrace(System.out);
}
```

All errors in the IPC server go to stdout instead of a proper logging framework (SLF4J). In a production Maven build, these messages are invisible or interleaved with build output. Same issue for `debug()` (line 184) and `info()` (line 190) in the same file.

Originally reported in #1944.

Issue #1988, opened 2026-07-21
Catching Throwable in non-framework code in PutTaskRequestContent


elharo:
**Affected version:** HEAD

**File:** `maven-resolver-transport-jetty/src/main/java/org/eclipse/aether/transport/jetty/PutTaskRequestContent.java:246, 298`

```java
} catch (Throwable t) {
    lockedSetTerminal(Content.Chunk.from(t, true));
}
```

Catches `OutOfMemoryError`, `StackOverflowError`, `InternalError`, etc. These should propagate. Wrapping them as terminal chunks converts fatal JVM errors into regular failure conditions, making them invisible to diagnostics.

Originally reported in #1944.

PR #2116, opened 2026-09-04
Fix concurrency, logging, and error handling issues (#1988, #1989, #1991, #1995)

slachiewicz:
### Summary of Changes

This PR addresses four open defects identified during the code audit:

1. **[#1988](https://github.com/apache/maven-resolver/issues/1988)**: In `PutTaskRequestContent.java`, change `catch (Throwable t)` to `catch (Exception t)` at lines 246 and 298. Catching `Throwable` trapped JVM fatal errors (`OutOfMemoryError`, `StackOverflowError`, `LinkageError`) and wrapped them into failed content chunks rather than letting them propagate for proper diagnostics.
2. **[#1989](https://github.com/apache/maven-resolver/issues/1989)**: In `IpcServer.java`, replace direct `System.out` / `System.err` console writes with standard SLF4J `Logger` (`LOGGER.debug`, `LOGGER.info`, `LOGGER.error`), ensuring server output integrates with standard logging configurations.
3. **[#1991](https://github.com/apache/maven-resolver/issues/1991)**: In `IpcClient.java`, synchronize retrieval and operations on volatile fields (`output`, `input`, `socket`) to prevent race conditions and NPEs when `close()` is invoked concurrently with active `send()` or `receive()`.
4. **[#1995](https://github.com/apache/maven-resolver/issues/1995)**: In `CollectResult.java`, back `exceptions` and `cycles` collections with `CopyOnWriteArrayList` and add thread-safe accessors (`setExceptions`, `setCycles`), ensuring thread-safe access during multi-threaded dependency collection (such as in `BfDependencyCollector`).

### Verification
- `mvn test -pl maven-resolver-transport-jetty,maven-resolver-named-locks-ipc,maven-resolver-api,maven-resolver-impl` passed with 781 tests, 0 failures, 0 errors.
- Checkstyle and Spotless verified clean.

### DIFF
--- a/maven-resolver-api/src/main/java/org/eclipse/aether/collection/CollectResult.java
+++ b/maven-resolver-api/src/main/java/org/eclipse/aether/collection/CollectResult.java
@@ -18,9 +18,8 @@
  */
 package org.eclipse.aether.collection;
 
-import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
 
 import org.eclipse.aether.RepositorySystem;
 import org.eclipse.aether.graph.DependencyCycle;
@@ -37,9 +36,9 @@ public final class CollectResult {
 
     private final CollectRequest request;
 
-    private List<Exception> exceptions;
+    private List<Exception> exceptions = new CopyOnWriteArrayList<>();
 
-    private List<DependencyCycle> cycles;
+    private List<DependencyCycle> cycles = new CopyOnWriteArrayList<>();
 
     private DependencyNode root;
 
@@ -50,8 +49,6 @@ public final class CollectResult {
      */
     public CollectResult(CollectRequest request) {
         this.request = requireNonNull(request, "dependency collection request cannot be null");
-        exceptions = Collections.emptyList();
-        cycles = Collections.emptyList();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


cstamas:
Can we get the build green?

slachiewicz:
Dropped the `[#1989]` commit; the other three are unchanged. `IpcAdapterIT` and `IpcAdapterNoForkIT` were failing on every OS and JDK because of it.

`IpcClient` forks the daemon with a two-jar classpath built from `getJarPath(getClass()) + getJarPath(IpcServer.class)`, both of which resolve to `maven-resolver-named-locks-ipc` itself. SLF4J is not on it, so the static `LOGGER` field failed during class initialisation in the forked JVM:

```
$ java -cp maven-resolver-named-locks-ipc/target/classes org.eclipse.aether.named.ipc.IpcServer
Exception in thread "main" java.lang.NoClassDefFoundError: org/slf4j/LoggerFactory
	at org.eclipse.aether.named.ipc.IpcServer.<clinit>(IpcServer.java:52)
```

The daemon died before it could connect back, the client's `future.get(5, TimeUnit.SECONDS)` expired, and every lock operation surfaced as `RuntimeException: Unable to create new sync context`.

#1989 is not reachable by adding the dependency either: the fork would also need an SLF4J binding, and without one the daemon logs nothing. `System.out` is the working channel here, since `IpcClient` already redirects the daemon's output to `resolver-ipcsync-<id>.log` via `ProcessBuilder.Redirect`. Reopening #1989 would need the classpath question settled first.

Verified: `IpcAdapterIT` 9 tests, 0 failures, 0 errors, where it was 5 failures and 4 errors before; full build 2055 tests, 0 failures, 0 errors.

*This comment was created with AI assistance.*

PR #2119, opened 2026-09-04
[1.9.x] Bring five changes over from master

slachiewicz:
Brings the 1.9.x line level with `master` on five changes it never received.

| Change | What `master` has that 1.9.x does not |
| --- | --- |
| Checksum reads and logged values | `LogSanitizer`, absent on 1.9.x, and an upper bound on checksum file reads in `ChecksumUtils` |
| Header scoping across redirects | `OriginScopedHeadersInterceptor` and `ResolverRedirectStrategy`, neither present on 1.9.x |
| Checksum policy merge | The policy strength comparator and mirror clamping in `DefaultRemoteRepositoryManager` |
| Descriptor cache key | `DataPool.toKey` keying on `(artifact, repositories)`; 1.9.x keys on the artifact alone, so a descriptor read from one repository can be handed back for a request naming another |
| Not-found markers | Expiry of stale markers once an artifact is downloaded |

The transport module differs between the lines: `master` renamed it to `maven-resolver-transport-apache`, so the header-scoping change is applied against 1.9.x's `maven-resolver-transport-http` rather than lifted.

Two further changes from the same batch are deliberately not here: bounding RFC 9457 error-body reads, and requiring verification before relabeling a resolved artifact. Both are on `master` but need a hand-port on this line (no SPI reporter, `File` instead of `Path`) and are left for a separate PR.

Verified: `mvn clean install` -> BUILD SUCCESS, 1078 tests, 0 failures, 0 errors; `spotless:check` clean across all 18 modules.

### DIFF
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/AbstractChecksumPolicy.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/AbstractChecksumPolicy.java
@@ -55,7 +55,11 @@ abstract class AbstractChecksumPolicy implements ChecksumPolicy {
             throws ChecksumFailureException {
         requireNonNull(algorithm, "algorithm cannot be null");
         requireNonNull(exception, "exception cannot be null");
-        logger.debug("Could not validate {} checksum for {}", algorithm, resource.getResourceName(), exception);
+        logger.debug(
+                "Could not validate {} checksum for {}",
+                algorithm,
+                LogSanitizer.sanitize(resource.getResourceName()),
+                exception);
     }
 
     @Override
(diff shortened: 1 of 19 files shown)


gnodet (approved):
Well-structured backport of five security and correctness changes from master to the 1.9.x line, correctly adapted to the 1.9.x API surface (`java.io.File`, 3-param `RepositoryPolicy`, HttpClient 4.x). All changes are well-tested with thorough coverage.

**Observations (non-blocking):**

- **Session-aware not-found logging** (medium): The `SESSION_NOT_FOUNDS` tracking mechanism (session-vs-previous-session log-level distinction) is not present on master yet — it originates from open PR #2112 which hasn't been merged. This means 1.9.x leads master on this feature, which the PR description's own principle ("a maintenance line should not lead") discourages. The behavioral impact is minor (log level only: DEBUG for same-session fall-through vs WARN for stale cross-session markers). Consider merging #2112 to master first, or dropping the session-awareness from this PR.

- `DescriptorKey.hashCode()` uses only the artifact while `equals()` also checks repositories — valid but increases hash collisions. Including repositories in the hash would be a cheap improvement.

- `effectivePort` duplicates `OriginScopedHeadersInterceptor.schemeDefaultPort` — acceptable for a maintenance backport.

- `captureStdErr` in tests depends on slf4j-simple binding to System.err — fragile but functional.

**Positive notes:**
- Master-to-1.9.x adaptation is carefully done throughout (File vs Path, HttpClient 4.x vs 5.x, `getKey()` vs `repositoryKeyFunction`).
- `@since 1.9.28` tags consistently applied to all new public API elements.
- `isAlgorithmHeaderFormat` replaces regex with a non-backtracking linear scan — good security practice.

### 📋 PR Metadata

PR #2120, opened 2026-09-04
Present the IPC token when attaching to a running daemon, and validate Redisson YAML addresses

slachiewicz:
Two changes to `maven-resolver-named-locks-ipc` and `maven-resolver-named-locks-redisson`.

**IPC token on the reconnect path.** `master` already writes a bootstrap token when a client spawns the daemon itself (#2083). The path that attaches to an *already running* daemon does not: it reads the address out of the lock file and connects without presenting anything. This extracts `connectToExistingServer`, teaches it the `address\ttoken` layout, and sends the token on that path too, so both ways in are treated the same.

**Redisson address validation.** `master` has the `aether.syncContext.named.redisson.allowInsecureAddress` property but nothing that reads the addresses out of the YAML, so the property governs nothing for a `Config.fromYAML` configuration. This adds `validateConfigAddresses`, covering the single-server, cluster, replicated, sentinel and master-slave shapes. A plaintext non-loopback address is refused with an `IllegalStateException` unless the property opts in, in which case it is logged as a warning.

Replaces #2117, whose description claimed the IPC token itself was new. It is not — only the reconnect path is.

Verified: `mvn clean install` -> BUILD SUCCESS, 2057 tests, 0 failures, 0 errors; `spotless:check` clean.

### DIFF
--- a/maven-resolver-named-locks-ipc/src/main/java/org/eclipse/aether/named/ipc/IpcClient.java
+++ b/maven-resolver-named-locks-ipc/src/main/java/org/eclipse/aether/named/ipc/IpcClient.java
@@ -118,6 +118,8 @@ public class IpcClient {
                     ByteChannel wrapper = new ByteChannelWrapper(socket);
                     input = new DataInputStream(Channels.newInputStream(wrapper));
                     output = new DataOutputStream(Channels.newOutputStream(wrapper));
+                    output.writeUTF(bootstrapToken != null ? bootstrapToken : "");
+                    output.flush();
                     receiver = new Thread(this::receive);
                     receiver.setDaemon(true);
                     receiver.start();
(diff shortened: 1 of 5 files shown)


gnodet (approved):
Well-structured security-hardening PR that closes two real gaps: (1) IPC clients reconnecting to an existing daemon now present the bootstrap token read from the lock file, and (2) Redisson addresses loaded from YAML configuration are validated against the insecure-address guard. Both changes are correctly implemented and properly tested.

**Observations (non-blocking):**

- The `bootstrapToken` field Javadoc (IpcClient.java line 97) is stale — it states the field "remains null when this client attached to an already running server," but after this PR `connectToExistingServer` parses and sets it. Consider updating the doc.

- The rejection log message (IpcServer.java line 273) includes the raw attacker-supplied token value verbatim. Consider logging only the fact of rejection without the value.

- `getServersConfig` uses `setAccessible(true)` for protected `Config` methods (`getMasterSlaveServersConfig`, `getReplicatedServersConfig`). If Redisson renames these, the reflection silently returns null and those topologies go unvalidated. Acceptable since the three most common topologies use the public `isXxxConfig()` path.

**Positive notes:**
- Wire-level protocol change (mandatory token handshake) is safe since client and daemon always ship together.
- Redisson YAML test correctly uses `lazyInitialization: true` to avoid requiring a real Redis server.
- Both IPC and Redisson test additions adequately cover the new behavior.

### 📋 PR Metadata

PR #2121, opened 2026-09-04
[1.9.x] Reject locally cached artifacts whose on-disk path spells different coordinates

slachiewicz:
Ports master's real-path check for locally cached artifacts to the 1.9.x line, which does not have it.

On case-insensitive or normalization-preserving filesystems (the macOS and Windows defaults) a file cached for one set of coordinates also answers lookups for coordinates that differ only in case or Unicode normalization. The tracking data is compared exactly, so the aliased file is treated as present-but-untracked and accepted with no download and no checksum verification. `checkFind` now compares the requested relative path against the symlink-resolved real path and treats a mismatch as not present.

Two deliberate differences from master, both forced by the line:

- **The base path is resolved lazily.** Master resolves it in the constructor, which had to gain `throws IOException` and a `Files.createDirectories` call because a fresh local repository does not exist yet. On 1.9.x that would push the checked exception through `EnhancedLocalRepositoryManagerFactory.newInstance`, so the value is resolved on first use instead. The `&&` short-circuit in `checkFind` means it is only reached once a file is known to exist, so the fresh-repository case never gets there.
- **The property is `aether.enhancedLocalRepository.verifyRealPath`**, matching this line's existing `aether.enhancedLocalRepository.trackingFilename`. Master uses `aether.lrm.enhanced.verifyRealPath` via its newer `PREFIX_LRM` namespace, which 1.9.x does not have.

Master is on `Path` throughout (`checkFind(Path…)`, `Files.isRegularFile`, `result.setPath`); 1.9.x is on `File`, so this is a port rather than a cherry-pick.

On the tests: `testFindCaseAliasedArtifactIsNotAvailable` only exercises the new branch on a case-insensitive filesystem — elsewhere the aliased file simply does not exist and the assertion holds for the ordinary reason. I confirmed the guard is what makes it pass, by re-running it with `verifyRealPath=false` and observing the alias resolve (`available = true`) on a case-insensitive volume. The other two tests are portable: verification enabled must not disturb an ordinary lookup, and the opt-out must not either.

Verified: `mvn clean install` -> BUILD SUCCESS, 1044 tests, 0 failures, 0 errors.

### DIFF
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/EnhancedLocalRepositoryManager.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/EnhancedLocalRepositoryManager.java
@@ -19,6 +19,8 @@
 package org.eclipse.aether.internal.impl;
 
 import java.io.File;
+import java.io.IOException;
+import java.nio.file.Path;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
@@ -34,6 +36,9 @@ import org.eclipse.aether.repository.LocalArtifactRegistration;
 import org.eclipse.aether.repository.LocalArtifactRequest;
 import org.eclipse.aether.repository.LocalArtifactResult;
 import org.eclipse.aether.repository.RemoteRepository;
+import org.eclipse.aether.util.ConfigUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import static java.util.Objects.requireNonNull;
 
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Correct, well-justified backport of master's real-path verification to 1.9.x. The lazy initialization with double-checked locking is properly thread-safe, fail-closed error handling is correct, and the property name follows the 1.9.x namespace convention.

**Positive notes:**
- Lazy `realBasePath` initialization is the right adaptation: avoids adding `throws IOException` to the factory's `newInstance` signature, which would be API-breaking.
- Double-checked locking (`volatile` field + `synchronized` + inner null check) is correctly implemented.
- `&&` short-circuit in `checkFind` guarantees `toRealPath()` is only called when the file exists, avoiding the fresh-repository failure that bit master in f2c15deb.
- Test is well-designed: passes on both case-sensitive (Linux) and case-insensitive (macOS/Windows) filesystems for different reasons.

**Low-severity observations (non-blocking):**
- `hasFaithfulRealPath` silently swallows `IOException | IllegalArgumentException` — fail-closed is correct, but a `LOGGER.debug()` would help diagnose unexpected re-downloads. Same omission exists on master.
- Tests cover only local artifact path; no test for case-aliased remote artifact (the loop branch in `find()`). Risk is minimal since both call the same `checkFind` method.

PR #2122, opened 2026-09-04
[1.9.x] Bound the error response body read by RFC9457Reporter

slachiewicz:
Ports the bounded error-body read from `master` (#2081) to the 1.9.x line. On this line `RFC9457Reporter` is a concrete class inside `maven-resolver-transport-http` rather than the SPI abstraction `master` has, so this is a port rather than a cherry-pick.

`getBody` previously handed the whole entity to `EntityUtils.toString`, so the size of an error response body was decided by the server. It now reads at most 64 KiB (`MAX_BODY_BYTES`) and leaves the rest unread. A body that does not parse as problem details, including one truncated at the limit, no longer surfaces as a `JsonSyntaxException` from `RFC9457Parser`: `generateException` returns and `HttpTransporter.handleStatus` reports the plain `HttpResponseException` for the status, the same fallback `master` uses.

The test server gained two `rfc9457/` paths, a non-JSON body and a payload larger than the limit, and both are expected to arrive as a plain 403 rather than as problem details.

Verified: `mvn -pl maven-resolver-transport-http -am verify` -> BUILD SUCCESS, 91 tests in the module, 0 failures; `spotless:apply` produced no further changes.

### DIFF
--- a/maven-resolver-transport-http/src/main/java/org/eclipse/aether/transport/http/RFC9457/RFC9457Reporter.java
+++ b/maven-resolver-transport-http/src/main/java/org/eclipse/aether/transport/http/RFC9457/RFC9457Reporter.java
@@ -18,16 +18,25 @@
  */
 package org.eclipse.aether.transport.http.RFC9457;
 
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
 
 import org.apache.http.Header;
+import org.apache.http.HttpEntity;
 import org.apache.http.HttpHeaders;
 import org.apache.http.client.methods.CloseableHttpResponse;
-import org.apache.http.util.EntityUtils;
 
 public class RFC9457Reporter {
     public static final RFC9457Reporter INSTANCE = new RFC9457Reporter();
+
+    /**
+     * Maximum number of bytes read from an error response body. Problem details payloads are small, while the body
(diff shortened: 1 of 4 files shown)


gnodet (approved):
Clean, correct backport of the bounded error-body read from master (#2081) to the 1.9.x concrete `RFC9457Reporter`. The `readBody` implementation faithfully reproduces master, the fallback from malformed/truncated bodies to the plain HTTP status is properly wired, and test coverage is adequate.

**Positive notes:**
- `readBody()` correctly caps at `MAX_BODY_BYTES` (64 KiB), handles EOF, and uses the same 8 KiB chunk size as master.
- The `ByteArrayOutputStream` approach avoids pre-allocating the full 64 KiB for smaller bodies.
- Null check on `RFC9457Parser.parse()` return value guards against `Gson.fromJson` returning Java null for JSON literal `null`.
- Truncated stream is closed via try-with-resources; `EntityUtils.consumeQuietly()` in the caller's finally block handles the already-closed stream gracefully.
- Test framework correctly uses JUnit 4, matching 1.9.x convention.

**Low-severity observation (non-blocking):**
- Master's `RFC9457ReporterTest` has additional unit tests for `generateException()` directly. The 1.9.x port covers these scenarios through integration tests in `HttpTransporterTest`, which is acceptable but leaves unit-level behavior less directly tested.

### 📋 PR Metadata

PR #2123, opened 2026-09-04
[1.9.x] Verify cached artifacts before registering them for another repository

slachiewicz:
Ports two `DefaultArtifactResolver` changes from `master` (#2080) to the 1.9.x line. `master` is on `Path` and a `ConfigurationKeys` class; 1.9.x is on `File` with literal keys, so this is a port.

**Existence check for artifacts cached from another repository.** When an artifact is present in the local repository but tracked for a repository that is not in the current build context, 1.9.x issues a bare existence check against the queried repository and, on success, registers the cached file for it. No content is transferred, so the cached bytes are never compared against that repository's checksums. The artifact is now downloaded again through the regular transfer path, which applies the repository's checksum policy before the registration is written. The previous behavior is available under `aether.artifactResolver.existenceCheckRelabel=true`, documented in `configuration.md`. The cost is one GET instead of one HEAD, in a situation that already logs at INFO ("present in the local repository, but cached from a remote repository ID that is unavailable").

**Snapshot normalization copy.** The base-version copy that snapshot normalization writes next to a timestamped snapshot was not registered in the tracking file, on either the download path or the local-hit path, so `EnhancedLocalRepositoryManager` later treated it as a locally installed file. Both paths now register the copy for the same repository as its source.

Tests cover the default download, the opt-in existence check, and the double registration on a timestamped snapshot.

Verified: `mvn -pl maven-resolver-impl -am verify` -> BUILD SUCCESS, 386 tests in the module, 0 failures, 1 skipped (pre-existing); `spotless:apply` produced no further changes.


gnodet (approved):
Clean, faithful port of master #2080 changes to 1.9.x. The `File`-vs-`Path`, JUnit 4-vs-5, and literal-key-vs-constant adaptations are all correct. Both behavioral changes (existence-check-to-download default, snapshot normalization registration) are logically sound and match master exactly.

**Positive notes:**
- Port correctly adapts all master patterns to 1.9.x conventions: `File`/`setFile`/`getFile`, inline string literals, single-param `result.addException(e)`, JUnit 4 `@Test public void`.
- `resolveCachedFromForeignRepository` test helper cleverly captures `isExistenceCheck()` flags, clearly proving default-download vs. legacy-existence-check behavior.
- Thread safety unaffected: new registration calls are inside the same exclusive `SyncContext` lock.

**Low-severity observations (non-blocking):**
- Snapshot normalization test covers only the download path. The local-hit path registration (lines 397-410) is not directly testable because `TestLocalRepositoryManager.find()` never sets a repository on the result. Same gap exists in master.
- `existenceCheckRelabel` description in `configuration.md` reads as though describing default behavior when it actually describes the `true` (non-default) case. Consistent with master.

### 📋 PR Metadata

PR #2125, opened 2026-09-04
[1.9.x] Qualify local repository tracking entries with the repository URL

slachiewicz:
Ports the tracking-key part of #2080 and #2104 to the 1.9.x line. `master` selects the function through `RepositoryKeyFunctionFactory` and `aether.lrm.enhanced.trackingRepositoryKeyFunction`; 1.9.x has no key-function layer, so the manager gains a small `TrackingKeyFunction` enum and the property lives in this line's namespace as `aether.enhancedLocalRepository.trackingRepositoryKeyFunction`, next to `trackingFilename`.

The repository component of a `_remote.repositories` entry was the bare repository id, so any two repositories that share an id were one origin as far as the enhanced local repository manager is concerned. The entry is now `<id>-<sha1(url)>`, with the context folded into the hash for repository managers. That is byte for byte the `nid_hurl` shape `master` writes, so a local repository shared between Maven 3.9 and Maven 4 reads and writes the same entries. Path composition and split prefixes keep using the id alone; only the tracking entries change.

**Upgrade impact, please weigh before merging.** An entry written under the old shape does not match a lookup under the new one. The artifact stays tracked, so the untracked-file branch does not accept it, and it is fetched again once through the checksum-verified path. On first run after upgrading, that is a re-download of every remotely cached artifact a build touches. `master` accepted the same trade for 2.0.23; on 1.9.x the population is every Maven 3.9 user. Setting the property to `nid` restores the old shape.

Tests: a repository with the same id and a different URL no longer matches; the written key has the qualified shape; the `nid` opt-out matches by id. All three also run under the split layout through `EnhancedSplitLocalRepositoryManagerTest`.

Verified: `mvn -pl maven-resolver-impl -am verify` -> BUILD SUCCESS, 389 tests in the module, 0 failures, 1 skipped (pre-existing); `spotless:apply` produced no further changes.


gnodet (approved):
Clean, faithful backport of the tracking-key URL-qualification from master (#2080/#2104) to 1.9.x. The `urlQualifiedRepositoryKey` method produces byte-identical output to master's `nidAndHurlRepositoryKey`, ensuring cross-line tracking-file interop.

**Positive notes:**
- Read and write paths are symmetric: `checkFind` and `getRepositoryKeys` both use `getTrackingRepositoryKey`.
- `TrackingKeyFunction` enum is the right simplification for 1.9.x (vs master's `RepositoryKeyFunctionFactory` layer).
- Config property name correctly follows 1.9.x namespace convention.
- Factory parser uses `Locale.ENGLISH` for case-insensitive matching and wraps `IllegalArgumentException` with descriptive message.
- Upgrade impact (one-time re-download) is well-documented with `nid` opt-out.
- Tests cover qualified key shape, same-id-different-URL rejection, and NID opt-out — all inherited by the split layout test.

**Low-severity observation (non-blocking):**
- Minor Javadoc grammar nit in the class doc: "does not accept it, but unavailable" → "but is unavailable".

### 📋 PR Metadata

PR #2132, opened 2026-09-08
[MRESOLVER-2076] Fix auto conflict resolver selection

jiteshkhatri11:
## Summary

Fix the `auto` conflict resolver configuration so that it selects the appropriate conflict resolver implementation based on the dependency graph size and available memory.

Previously, `auto` unconditionally delegated to `ClassicConflictResolver`.

The implementation now:
- Selects `PathConflictResolver` when the dependency graph fits within the estimated memory threshold.
- Falls back to `ClassicConflictResolver` for larger graphs.
- Ensures conflict IDs are available before making the selection.
- Keeps explicit `classic` and `path` configurations unchanged.

## Testing

- `mvn -pl maven-resolver-util -Dtest=ConflictResolverTest test`
- `mvn -pl maven-resolver-util -am test`
- `mvn -pl maven-resolver-util -DskipTests spotless:check`
- `git diff --check`

All tests passed.

### DIFF
--- a/maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/transformer/ConflictResolver.java
+++ b/maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/transformer/ConflictResolver.java
@@ -18,6 +18,7 @@
  */
 package org.eclipse.aether.util.graph.transformer;
 
+import java.util.ArrayDeque;
 import java.util.Arrays;
 import java.util.Collection;
 
@@ -119,8 +120,8 @@ public class ConflictResolver implements DependencyGraphTransformer {
     /**
      * The name of the conflict resolver implementation to use: "auto" (default), "path", or "classic" (same as Maven 3).
      * <p>
-     * When set to "auto", the resolver will currently just use "classic". The idea here, is that this value will
-     * always select the best (most robust, most performant) one, which currently is "classic".
+     * When set to "auto", the resolver automatically selects the most appropriate implementation based on the
+     * dependency graph size and available memory.
      *
      * @since 2.0.11
      * @configurationSource {@link RepositorySystemSession#getConfigProperties()}
@@ -254,8 +255,12 @@ public class ConflictResolver implements DependencyGraphTransformer {
             throws RepositoryException {
         String cf = ConfigUtils.getString(
                 context.getSession(), DEFAULT_CONFLICT_RESOLVER_IMPL, CONFIG_PROP_CONFLICT_RESOLVER_IMPL);
+
         ConflictResolver delegate;
-        if (AUTO_CONFLICT_RESOLVER.equals(cf) || CLASSIC_CONFLICT_RESOLVER.equals(cf)) {
+
+        if (AUTO_CONFLICT_RESOLVER.equals(cf)) {
+            delegate = selectConflictResolver(node, context);
+        } else if (CLASSIC_CONFLICT_RESOLVER.equals(cf)) {
             delegate = new ClassicConflictResolver(versionSelector, scopeSelector, optionalitySelector, scopeDeriver);
         } else if (PATH_CONFLICT_RESOLVER.equals(cf)) {
             delegate = new PathConflictResolver(versionSelector, scopeSelector, optionalitySelector, scopeDeriver);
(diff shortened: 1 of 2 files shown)


maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/transformer/ConflictResolver.java
```diff
@@ -254,18 +255,76 @@ public DependencyNode transformGraph(DependencyNode node, DependencyGraphTransfo
...
+     * Automatically selects the conflict resolver based on the estimated memory requirements.
+     * PathConflictResolver is used for dependency trees that fit within the memory threshold,
+     * while ClassicConflictResolver is used for larger trees.
+     */
+    private ConflictResolver selectConflictResolver(DependencyNode node, DependencyGraphTransformationContext context)
+            throws RepositoryException {
+
+        if (context.get(TransformationContextKeys.CONFLICT_IDS) == null) {
+            new ConflictIdSorter().transformGraph(node, context);
+        }
```
gnodet: 🔴 **Bug: wrong context key checked.** This checks `CONFLICT_IDS` (set by `ConflictMarker`) but should check `SORTED_CONFLICT_IDS` (set by `ConflictIdSorter`). The original code in PR #1938 correctly checked `SORTED_CONFLICT_IDS`.

As written, if `ConflictMarker` has already run (populating `CONFLICT_IDS`) but `ConflictIdSorter` hasn't, this guard passes and the sorter never runs. In practice both delegate resolvers (`ClassicConflictResolver` line 119-124 and `PathConflictResolver` line 141-146) independently check for `SORTED_CONFLICT_IDS` and call `ConflictIdSorter` themselves, so this doesn't cause a runtime failure — but it makes this block dead code that never triggers when it should.

```suggestion
        if (context.get(TransformationContextKeys.SORTED_CONFLICT_IDS) == null) {
            new ConflictIdSorter().transformGraph(node, context);
        }
```

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/transformer/ConflictResolver.java
```diff
@@ -254,18 +255,76 @@ public DependencyNode transformGraph(DependencyNode node, DependencyGraphTransfo
...
+
+        Runtime rt = Runtime.getRuntime();
+        long available = rt.maxMemory() - (rt.totalMemory() - rt.freeMemory());
```
gnodet, line 288: ⚠️ **Unreliable heap measurement.** `Runtime.freeMemory()` reflects the free space in the *currently allocated* heap, not total available memory. Its value fluctuates wildly depending on whether GC has recently run, allocation pressure from other threads, and JVM ergonomics.

This means the same dependency graph can get `PathConflictResolver` on one build and `ClassicConflictResolver` on the next, producing non-deterministic behavior — a property Maven users don't expect from dependency resolution. The original PR #1938 had this same issue but was accepted as a pragmatic trade-off alongside extensive benchmarking and the PathConflictResolver memory optimizations. Without those optimizations, the risk is higher.

Consider requesting a GC before measurement (`System.gc()` — advisory but common for memory heuristics), or using a fixed configurable threshold instead of runtime measurement.

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/transformer/ConflictResolver.java
```diff
@@ -254,18 +255,76 @@ public DependencyNode transformGraph(DependencyNode node, DependencyGraphTransfo
...
+
+        // Walk the dependency tree to count total nodes (including diamond-expanded duplicates).
+        // The Path tree mirrors this structure, so the count directly reflects Path tree size.
+        // Use early-exit: stop counting once we exceed the threshold.
+        if (treeExceedsThreshold(node, maxPathNodes)) {
+            return new ClassicConflictResolver(versionSelector, scopeSelector, optionalitySelector, scopeDeriver);
+        } else {
+            return new PathConflictResolver(versionSelector, scopeSelector, optionalitySelector, scopeDeriver);
+        }
+    }
```
gnodet, line 302: ⚠️ **Missing tests.** The original PR #1938 that introduced this logic included 5 test cases covering: auto mode selection, explicit path/classic config dispatch, unknown config rejection, and default config behavior. This PR re-introduces the production code without any tests.

Given this code was removed for stability reasons, re-introducing it without tests — especially for the memory threshold heuristic and the fallback behavior — is risky. At minimum, tests should verify:
1. `auto` selects `PathConflictResolver` for small graphs
2. `auto` selects `ClassicConflictResolver` for large graphs
3. Explicit `path` and `classic` configs are unaffected
4. The tree-walking threshold check terminates early

maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/transformer/ConflictResolver.java
```diff
@@ -119,8 +120,8 @@ public class ConflictResolver implements DependencyGraphTransformer {
     /**
      * The name of the conflict resolver implementation to use: "auto" (default), "path", or "classic" (same as Maven 3).
      * <p>
```
gnodet, line 122: 💡 **Javadoc update is premature.** The class-level Javadoc (lines 46-53) still says PathConflictResolver is "Not yet recommended for production" and advises "All projects: Use ClassicConflictResolver for optimal correctness." This Javadoc update claims auto mode intelligently selects the right one, but the class-level docs discourage using PCR at all. Either the class-level docs should be updated to match, or this Javadoc should stay conservative.

gnodet (changes requested):
## Review Summary

This PR re-introduces the auto-selection heuristic for `PathConflictResolver` vs `ClassicConflictResolver` that was **intentionally removed** in [cf2eaeff](https://github.com/apache/maven-resolver/commit/cf2eaeff) (PR #1942) by @cstamas ~2.5 months ago. That commit's message says:

> Put PCR aside for now (#1942) — It's proven a bit problematic, and memory hungry, while the promise of O(N) over O(N^2) is cool.

The original auto-selection logic was introduced in [d331083b](https://github.com/apache/maven-resolver/commit/d331083b) (PR #1938) with comprehensive tests (5 new test cases), memory optimizations to `PathConflictResolver`, and extensive benchmarking. This PR re-introduces a subset of that logic — just the `selectConflictResolver` and `treeExceedsThreshold` methods — without the accompanying `PathConflictResolver` memory optimizations from #1938, and without any tests.

Before re-enabling auto-selection, the underlying `PathConflictResolver` issues that motivated the revert need to be resolved first. The PR description doesn't reference the revert or explain what has changed to make PCR safe to auto-select again.

See inline comments for specific code issues.

jiteshkhatri11:
@gnodet Thanks for the detailed review. I went through the concerns, especially the relationship between #1938 and #1942.

I agree that the current implementation needs to address the previous PathConflictResolver memory/safety concerns before re-enabling automatic selection.

I'll first fix the `SORTED_CONFLICT_IDS` check and add the missing tests for auto selection, explicit resolver configurations, and threshold/fallback behavior.

I'll also review #1938 and #1942 to make sure the auto-selection logic is reintroduced together with the necessary safeguards.

Thanks again for pointing this out.

PR #2134, opened 2026-09-08
fix: extend tracking fallback to accept same-id-different-URL entries

gnodet:
## Problem

PR #2133 added a legacy tracking fallback for entries written by older resolvers using ID-only keys (`nid` format). However, it does **not** fix the actual Maven IT failures.

The real problem (identified by @cstamas): when ITs run, support artifacts are installed from real Central, producing `nid_hurl` tracking entries:
```
sisu-maven-plugin-1.1.0.jar>central-8fac3ebd6edbaca3c794783fa38088af0aa128e7=
```
where `8fac3ebd...` = `sha1(https://repo.maven.apache.org/maven2)`.

Then each IT overrides `central` to `file:target/null` to prevent remote access. The resolver looks for:
```
sisu-maven-plugin-1.1.0.jar>central-<sha1(file:target/null)>=
```

The sha1 hashes differ → exact lookup misses → artifact treated as "present but unavailable" → tries to download from `file:target/null` → 💥. All 9 IT jobs fail (4229 failures across 507 test classes).

## Fix

Add a second fallback stage inside the existing `legacyTrackingFallback` guard: after the legacy ID-only check, scan tracking entries for any key matching the repo-ID prefix (`repoId-`). This accepts artifacts tracked under the same logical repository identity regardless of the URL hash suffix.

The prefix uses `repoId-` (with the dash separator from the `nid_hurl` format) to avoid false positives with repos whose ID shares a prefix (e.g. `central` must not match `centralbackup-<sha1>`).

## Tests

- `testUrlQualifiedTrackingSameIdDifferentUrlAcceptedViaFallback`: same-id-different-URL now accepted (was: `assertFalse`)
- `testUrlQualifiedTrackingAcceptsSameIdDifferentUrlEntries`: **new** — simulates the IT scenario (Central → `file:target/null`)
- Updated factory tests to reflect relaxed same-id behavior
- All 587 resolver tests pass

## Related

- Follow-up to #2133
- Fixes [apache/maven#13078 (comment)](https://github.com/apache/maven/pull/13078#issuecomment-5583775310)

### DIFF
--- a/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/EnhancedLocalRepositoryManager.java
+++ b/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/EnhancedLocalRepositoryManager.java
@@ -326,6 +326,27 @@ class EnhancedLocalRepositoryManager extends SimpleLocalRepositoryManager {
                     result.setRepository(repository);
                     return true;
                 }
+                // Same-ID-different-URL fallback: if the tracking file contains a URL-qualified entry for the
+                // same repository ID but with a different URL hash (e.g. real Central tracked as
+                // "central-<sha1(realUrl)>=" but the current build overrides central to "file:target/null"),
+                // the exact lookup misses because sha1(realUrl) != sha1(file:target/null). Match by repo-ID
+                // prefix: any entry starting with "filename>repoId-" is accepted as originating from the same
+                // logical repository.
+                String repoIdPrefix = getKey(path, legacyKey + "-");
+                for (Object key : props.keySet()) {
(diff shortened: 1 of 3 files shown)


gnodet (commented):
Solid fix. The prefix-based same-ID fallback correctly resolves the IT failure scenario (Central URL override producing mismatched SHA1 hashes in tracking entries). Well-guarded behind `legacyTrackingFallback`, thorough inline comments, and adequate test coverage.

cstamas:
Does not this fully defies f013 fix?

cstamas:
I'd recommend to close this, if you agree that #2135 supersede this

PR #2138, opened 2026-09-11
[MRESOLVER-1995] Synchronize CollectResult access

### LINKED ISSUE #1995: Non-thread-safe access in synchronized Results class in DependencyCollectorDelegate

jiteshkhatri11:
## Description

Fixes #1995.

Synchronize access to the exception and dependency cycle collections in `CollectResult` to make concurrent updates thread-safe.

### Changes

* Synchronize `getExceptions()`
* Synchronize `addException()`
* Synchronize `getCycles()`
* Synchronize `addCycle()`
* Add a regression test covering concurrent exception additions

### Testing

* `mvn -pl maven-resolver-api test`
* Tests run: 91
* Failures: 0
* Errors: 0
* Skipped: 0

### DIFF
--- a/maven-resolver-api/src/main/java/org/eclipse/aether/collection/CollectResult.java
+++ b/maven-resolver-api/src/main/java/org/eclipse/aether/collection/CollectResult.java
@@ -68,7 +68,7 @@ public final class CollectResult {
      *
      * @return The exceptions that occurred, never {@code null}.
      */
-    public List<Exception> getExceptions() {
+    public synchronized List<Exception> getExceptions() {
         return exceptions;
     }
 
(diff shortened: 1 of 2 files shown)


elharo (changes requested):
If this is needed, it would likely be safer, faster, and more robust to use thread safe collections instead of synchronizing. Also we should get rid of the lazy initialization. And fix the exposure of mutable internal state. Both the existing code and the proposed fix look naive.

